PR #1, opened 2021-02-09
rework pkg/compare

jaypipes:
There are three primary issues with the `google/go-cmp` package that
currently makes up the implementation of `pkg/compare` in ACK:

First, the google/go-cmp package contains the following warning in its
documentation:

> It is intended to only be used in tests, as performance is not a goal
> and it may panic if it cannot compare the values. Its propensity
> towards panicking means that its unsuitable for production environments
> where a spurious panic may be fatal.

Secondly, `go-cmp` has an awkward (IMHO) interface for setting up custom
comparison logic, relying on a Reporter interface that is neither
intuitive or extensible.

Finally, because `go-cmp` must deal with unknown data types in its
comparison implementation, it necessarily needs to use the `reflect` Go
package. This package suffers from a similarly non-intuitive and
unreadable interface (pun intended) for type inspection. However,
because ACK constructs its service controllers from API model files that
include field type information, there is no need to have runtime type
inspection like `reflect` and `go-cmp`. Instead, we can read the API
model files for the aws-sdk-go module for the service API in question
and generate Go code that uses concrete type comparison logic.

This patch lays the groundwork for swapping out our mechanisms in the
reconciler for comparing resources.

We replace the use of `google/go-cmp` in `pkg/compare` with a more
intuitive interface for processing and identifying differences between
two CRs under comparison. There is an `ackcompare.Delta` struct that has
two methods `Add` and `DifferentAt`. `ackcompare.Delta.Add` is used to
append a difference between two resources at a particular field path to
the `Delta` struct. `DifferentAt` returns whether the resources under
comparison at a specified fieldPath are different.

### DIFF
--- a/pkg/compare/delta_test.go
+++ b/pkg/compare/delta_test.go
@@ -1,0 +1,63 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package compare_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/aws-controllers-k8s/runtime/pkg/compare"
+)
+
+type Foo struct {
+	Bar string
+	Baz Baz
+}
+
+type Baz struct {
+	Y string
+}
+
+func TestDifferentAt(t *testing.T) {
+	require := require.New(t)
+
+	a := Foo{
+		Bar: "a_bar",
+		Baz: Baz{
+			Y: "a_baz_y",
+		},
+	}
+	b := Foo{
+		Bar: "b_bar",
+		Baz: Baz{
+			Y: "b_baz_y",
+		},
+	}
+
+	d := compare.NewDelta()
+	d.Add("", a, nil)
+	require.True(d.DifferentAt(""))
+
+	d = compare.NewDelta()
+	d.Add("Bar", a.Bar, b.Bar)
+	require.True(d.DifferentAt("Bar"))
+	require.False(d.DifferentAt("Baz"))
+
+	d = compare.NewDelta()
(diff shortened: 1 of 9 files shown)


pkg/compare/path.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Push adds a new part to the Path.
+func (p Path) Push(part string) {
+	p.parts = append(p.parts, part)
+}
+
+// Pop removes the last part from the Path
+func (p Path) Pop() {
+	p.parts = p.parts[:len(p.parts)-1]
```
vijtrip2, line 31: Should there be an empty check? is this prone to OutOfBoundsException currently?

A-Hilaly (approved):
LGTM 👍 i left few nit comments

pkg/compare/slice.go
```diff
@@ -0,0 +1,32 @@
...
+	if len(a) != len(b) {
+		return false
+	}
+	for _, aVal := range a {
+		if !util.InStringPs(*aVal, b) {
+			return false
+		}
+	}
+	return true
+}
```
A-Hilaly, line 32: nit: i'm not sure if this is a big deal but this function can return true for `["a", "a", "b"]` and `["a", "b", "b"]` . Maybe worth to concider in the future

jaypipes: ah, good point. I will update and add a test case for that.

pkg/compare/path.go
jaypipes: Nice catch. Adding test...

pkg/compare/delta_test.go
```diff
@@ -0,0 +1,59 @@
...
+	d := compare.NewDelta()
+	d.Add("", a, nil)
+
+	require.True(d.DifferentAt(""))
+
+	d = compare.NewDelta()
+	d.Add("Bar", a.Bar, b.Bar)
+
+	d = compare.NewDelta()
+	d.Add("Baz.Y", a.Baz.Y, b.Baz.Y)
```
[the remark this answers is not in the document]
jaypipes: lol, not sure why I forgot that... fixed.

jaypipes:
@A-Hilaly @vijtrip2 thanks for the review guys. Updated with fixes. Please review.

pkg/compare/slice.go
```diff
@@ -0,0 +1,39 @@
...
+		sb[x] = *b[x]
+	}
+	sort.Strings(sa)
+	sort.Strings(sb)
+	for x, aVal := range sa {
+		bVal := sb[x]
+		if aVal != bVal {
+			return false
+		}
+	}
```
A-Hilaly, line 37: Nice!

A-Hilaly (approved):
/lgtm let's ship it 🚢

PR #2, opened 2021-02-10
Create adoption reconciler

RedbackThomson:
Issue #, if available: 

Description of changes:
- Rename reconciler struct to `ACKReconciler` to allow for new reconciler types other than `AWSResourceReconciler`
- Create an adoption reconciler which supports adoption through string names and IDs (not yet ARNs)


pkg/runtime/reconciler.go
```diff
@@ -35,27 +35,31 @@ import (
...
+// It implements the upstream controller-runtime `Reconciler` interface.
+//
+// The upstream controller-runtime.Manager object ends up managing MULTIPLE
+// controller-runtime.Controller objects (each containing a single resourceReconciler
+// object)s and sharing watch and informer queues across those controllers.
+type resourceReconciler struct {
+	reconciler
+	rmf acktypes.AWSResourceManagerFactory
+	rd  acktypes.AWSResourceDescriptor
+}
```
jaypipes, line 58: Nice. This is exactly the kind of polymorphism I was thinking of when we chatted about the implementation. ++

pkg/runtime/service_controller_test.go
```diff
@@ -99,13 +99,7 @@ func TestServiceController(t *testing.T) {
 	reg := ackrt.NewRegistry()
 	reg.RegisterResourceManagerFactory(rmf)
 
-	vi := ackrt.VersionInfo{
-		GitCommit:  "test-commit",
-		GitVersion: "test-version",
-		BuildDate:  "now",
-	}
-
-	sc := ackrt.NewServiceController("bookstore", "bookstore.services.k8s.aws", vi)
```
jaypipes, line 108: you just need to rebase with the latest runtime repo code. The above is necessary due to the breakout of the runtime and service controller code and not having a single pkg/version to use.

pkg/types/ack_adoption_reconciler.go
```diff
@@ -0,0 +1,19 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package types
+
+// ACKAdoptionReconciler
+type ACKAdoptionReconciler interface {
+	ACKReconciler
+}
```
jaypipes, line 19: Is there a need for this interface?

pkg/types/ack_reconciler.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// ACKReconciler is responsible for reconciling the state of any single custom
+// resource within the cluster.
+//
+// The upstream controller-runtime.Manager object ends up managing MULTIPLE
+// controller-runtime.Controller objects (each containing a single
+// ACKReconciler object)s and sharing watch and informer queues across
+// those controllers.
+type ACKReconciler interface {
```
jaypipes, line 29: I'd prefer to rename this to just `Reconciler`. `ACKReconciler` is stutterish and redundant. :) The reason AWSResourceReconciler was chosen was that it described the thing that was being reconciled. An `ACKReconciler` wouldn't be reconciling "ACK"s, obviously.

pkg/types/aws_resource_descriptor.go
```diff
@@ -65,4 +68,7 @@ type AWSResourceDescriptor interface {
 	// the resource. This will allow the Kubernetes API server to delete the
 	// underlying CR.
 	MarkUnmanaged(AWSResource)
+	// MarkAdopted places descriptors on the custom resource that indicate the
+	// resource was not created from within ACK.
+	MarkAdopted(AWSResource)
```
jaypipes, line 73: This should be fine in the resource descriptor interface. Not sure whether in the future the process of marking adoption will mean setting AWS tags on the resource. If it does, then this may need to move to the AWSResourceManager interface.... we'll just have to wait and see.

scripts/lib/k8s.sh
```diff
@@ -0,0 +1,26 @@
...
+# controller_gen_version_equals accepts a string version and returns 0 if the
+# installed version of controller-gen matches the supplied version, otherwise
+# returns 1
+#
+# Usage:
+#
+#   if controller_gen_version_equals "v0.4.0"; then
+#       echo "controller-gen is at version 0.4.0"
+#   fi
+k8s_controller_gen_version_equals() {
```
jaypipes, line 18: This is unused I think?

scripts/build-runtime.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
jaypipes, line 1: how about we call this `scripts/build-apis.sh` since we're not actually building the runtime (since it's not a binary and doesn't have any code generation bits like the APIs do.

templates/boilerplate.txt
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
```
jaypipes, line 12: unused file.

scripts/build-runtime.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# A script that builds a single ACK service controller for an AWS service API
```
jaypipes, line 3: copy/pasta oops. :)

jaypipes (changes requested):
This is a great start, @RedbackThomson! Thanks!

Some suggestions and comments inline for you to ponder.

pkg/types/ack_adoption_reconciler.go
RedbackThomson: Just futureproofing. But not at this stage, no

scripts/lib/k8s.sh
RedbackThomson: Yeah I was thinking of checking within the `build-runtime.sh` script, again. Since there is a chance someone is calling the script directly, rather than through any of the scripts in `community`. But I don't like the code duplication.

templates/boilerplate.txt
RedbackThomson: Used as part of `controller-gen` in the build script.

pkg/types/ack_adoption_reconciler.go
jaypipes: OK, let's leave it out for now, then.

pkg/types/ack_reconciler.go
```diff
@@ -0,0 +1,37 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
jaypipes, line 1: mind changing the filename to `reconciler.go`?

pkg/types/aws_resource.go
```diff
@@ -51,4 +53,8 @@ type AWSResource interface {
 	// apimachinery/runtime.Object and the Kubernetes
 	// apimachinery/apis/meta/v1.Object interfaces
 	RuntimeMetaObject() RuntimeMetaObject
+	// SetObjectMeta sets the ObjectMeta field for the resource
+	SetObjectMeta(meta metav1.ObjectMeta)
+	// SetARN will set the ARN field in the resource's metadata status field
+	SetARN(*ackv1alpha1.AWSResourceName)
```
jaypipes, line 59: What we need is an interface that will help the `AWSResourceManager` determine which field in the `acktypes.AWSResource`'s underlyihng `Spec` or `Status` struct maps to each field in a `ackcorev1alpha1.AWSIdentifiers` struct. We already know that the `ackcorev1alpha1.AWSIdentifiers.ARN` field maps to the `Status.ACKMetadata.ARN` field because we standardize on that. What we don't really know is which field in the `Spec` corresponds to the `ackcorev1alpha1.AWSIdentifiers.Name` and `ackcorev1alpha1.AWSIdentifiers.ID` field.

Something like this might work?

```go
// AWSIdentifierFieldMapper helps a resource manager determine which fields
// in a resource represent name or generated IDs.
type AWSIdentifierFieldMapper interface {
    // NameField returns the path to the field in the underlying resource's `Spec`
    // or `Status` struct that represents the resource's name. This will be empty
    // if there is no name field for the resource.
    NameField() string
    // IDField returns the path to the field in the underlying resource's `Spec`
    // or `Status` struct that represents the resource's AWS-generated ID field.
    // This will be empty if there is no such field for the resource.
    IDField() string
}
```

We could then have the `AWSResourceDescriptor` interface have a method that returns a `AWSIdentifierFieldMapper`-implementing struct:

```go
type AWSResourceDescriptor interface {
// ...
    // IdentifierFieldMapper returns an `AWSIdentifierFieldMapper` that instructs callers
    // which fields in the `AWSResource` represent name or generated IDs.
    IdentifierFieldMapper() AWSIdentifierFieldMapper
}
```

Furthemore, instead of a `SetNameField` and `SetARN` method (and I presume you would eventually want a `SetID` method?) on the `AWSResource` interface above, how about just adding a `SetIdentifiers` method that accepts a single parameter of type `ackcorev1alpha1.AWSIdentifiers`? That way, the `AdoptedResourceReconciler` can construct a new `AWSResource` and set its identifier fields using a single method.

So, instead of this code that you have above in the `AdoptedResourceReconciler`:

```go
	// Set spec fields prior to reading
	if desired.Spec.AWS.Name != nil {
		readableResource.SetNameField(*desired.Spec.AWS.Name)
	} else if desired.Spec.AWS.ID != nil {
		readableResource.SetNameField(*desired.Spec.AWS.ID)
	} else if desired.Spec.AWS.ARN != nil {
		readableResource.SetARN(desired.Spec.AWS.ARN)
	} else {
		return fmt.Errorf("must provide at least one value for identifier")
	}
```

you could just do this instead:

```go
    readableResource.SetIdentifiers(*desired.Spec.AWS)
```

and let the plumbing inside the controller's `pkg/resource/$RESOURCE/descriptor.go` and `pkg/resource/$RESOURCE/resource.go` files use the `AWSIdentiferFieldMapper` to determine the right fields in the underlying resource to pull out.

jaypipes (changes requested):

apis/core/v1alpha1/annotations.go
```diff
@@ -16,15 +16,14 @@ package v1alpha1
...
 	// indicating to the ACK service controller that it should expect a backend
 	// AWS service API resource to already exist (and that ACK should "adopt"
-	// the resource into its management). If this annotation is NOT SET on a
-	// CR, that means the user expects the ACK service controller to create the
-	// backend AWS service API resource.
-	AnnotationARN = AnnotationPrefix + "arn"
+	// the resource into its management). If this annotation is set to false on
+	// a CR, that means the user expects the ACK service controller to create
+	// the backend AWS service API resource.
+	AnnotationAdopted = AnnotationPrefix + "adopted"
```
jaypipes, line 26: Been thinking more about this annotation and whether we really need it. I'm not so sure we do, at least in a boolean form. A couple questions that highlight my concerns:

1) If this annotation is not present on a CR, what does that mean?

2) Above, it states "If this annotation is set to false on a CR, that means the user expects the ACK service controller to  create the backend AWS service API resource." But this isn't really true, right? It's not the annotation's presence or lack of presence that dictates whether the service controller takes some action. Instead, it's the creation of an `AdoptedResource` CR explicitly by the Kubernetes user that is the signal for a service controller to grab the latest observed state of the referred-to resource instead of calling the API's Create operation.

3) Given 2), would it make more sense to add an annotation called `services.k8s.aws/adopted-on` that is a timestamp of when the `AdoptedResourceReconciler` successfully created the resource CR after pulling the latest observed state from the AWS API?

apis/core/v1alpha1/conditions.go
```diff
@@ -23,6 +23,9 @@ import (
 type ConditionType string
 
 const (
+	// ConditionTypeAdopted indicates that the adopted resource custom resource
+	// has been successfully reconciled and the target has been created
+	ConditionTypeAdopted ConditionType = "ACK.Adopted"
```
jaypipes, line 28: Is this condition to be applied to the `AdoptedResource` CR or to the target resource CR? If the latter, then perhaps we don't need a `services.k8s.aws/adopted-on` annotation at all, since the Condition has a timestamp assocated with its creation.

RedbackThomson: This condition is currently being applied to the `AdoptedResource` CR to describe to the user that it has essentially reached the terminal condition.

apis/core/v1alpha1/annotations.go
RedbackThomson: Sure, I could see more value in this becoming a timestamp. It was initially a boolean so that you could easily run a filter of `services.k8s.aws/adopted=='true'` on all resource annotations, but I understand that timestamp probably provide that plus more.
This annotation isn't being used by any system, it really is just informational to the user.

akartsky (approved):

pkg/runtime/service_controller.go
```diff
@@ -70,6 +74,15 @@ func (c *serviceController) GetReconcilers() []acktypes.AWSResourceReconciler {
 	return c.reconcilers
 }
 
+// GetResourceManagerFactories returns the map of resource manager
+// factories, keyed by the GroupKind of the resource managed by the resource
+// manager produced by that factory
+func (c *serviceController) GetResourceManagerFactories() map[string]acktypes.AWSResourceManagerFactory {
```
jaypipes, line 80: :+1:

jaypipes (approved):
I'm eager to see e2e tests for this functionality. :)

pkg/runtime/adoption_reconciler.go
```diff
@@ -0,0 +1,468 @@
...
+
+// onError will patch the adopted resource with the given error and return the
+// same error back
+func (r *adoptionReconciler) onError(
+	ctx context.Context,
+	res *ackv1alpha1.AdoptedResource,
+	err error,
+) error {
+	r.patchAdoptedCondition(ctx, res, err)
+	return err
```
A-Hilaly, line 256: Error return value is not check here, did you mean `return r.patchAdoptedCondition(ctx, res, err)` ?

pkg/runtime/service_controller.go
```diff
@@ -124,6 +137,12 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 		}
 		c.reconcilers = append(c.reconcilers, rec)
 	}
+
+	rec := NewAdoptionReconciler(c, c.log, cfg, c.metrics)
+	if err := rec.BindControllerManager(mgr); err != nil {
+		return err
+	}
+	c.adoptionReconciler = rec
```
A-Hilaly, line 145: nit/QQ: Would it be valuable to add a flag to enable/disable adoption reconciler? i'm thinking about `--enable-adoption-reconciler` and default it to `true`

pkg/runtime/adoption_reconciler.go
```diff
@@ -0,0 +1,468 @@
...
+	)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// handleReconcileError will handle errors from reconcile handlers, which
+// respects runtime errors.
+func (r *adoptionReconciler) handleReconcileError(err error) (ctrlrt.Result, error) {
```
A-Hilaly, line 371: nit: similar to `resourceReconciler.handleReconcileError` - in the future we can merge them

apis/core/v1alpha1/types.go
```diff
@@ -0,0 +1,99 @@
...
+	// +patchStrategy=merge
+	OwnerReferences []metav1.OwnerReference `json:"ownerReferences,omitempty" patchStrategy:"merge" patchMergeKey:"uid" protobuf:"bytes,6,rep,name=ownerReferences"`
+}
+
+type AdoptionStatus string
+
+const (
+	AdoptionStatus_Adopted AdoptionStatus = "Adopted"
+	AdoptionStatus_Error   AdoptionStatus = "Error"
+)
```
A-Hilaly, line 99: QQ: i can't find where these constants are used, are they?

pkg/runtime/adoption_reconciler.go
```diff
@@ -0,0 +1,468 @@
...
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	ackcfg "github.com/aws-controllers-k8s/runtime/pkg/config"
+	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
+	ackmetrics "github.com/aws-controllers-k8s/runtime/pkg/metrics"
+	"github.com/aws-controllers-k8s/runtime/pkg/requeue"
+	ackrtcache "github.com/aws-controllers-k8s/runtime/pkg/runtime/cache"
+	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	k8sctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
```
A-Hilaly, line 36: tiny nit: running `goimports` can help organizing imports

pkg/runtime/adoption_reconciler.go
```diff
@@ -0,0 +1,468 @@
...
+	roleARN, _ := r.cache.Accounts.GetAccountRoleARN(string(acctID))
+	return ackv1alpha1.AWSResourceName(roleARN)
+}
+
+// getRegion returns the AWS region that the given resource is in or should be
+// created in. If the CR have a region associated with it, it is used. Otherwise
+// we look for the namespace associated region, if that is set we use it. Finally
+// if none of these annotations are set we use the use the region specified in the
+// configuration is used
+func (r *adoptionReconciler) getRegion(
```
A-Hilaly, line 432: same thing as for `handleReconcileError`

A-Hilaly (commented):
Neat !! Very nice to see the adoption reconciler in action! :+1: 
I left few non-blocking comments/questions

apis/core/v1alpha1/types.go
jaypipes: Good point. I don't believe this is used anywhere and probably should be removed.

pkg/runtime/service_controller.go
jaypipes: Since resources can be opted out of adoption (via the generator.yaml), I'm OK with adding a CLI flag after this first patch is merged. We can even consider a CRD that allows the service controller to be configured without restarting the service controller -- and that CRD could allow configuring things like the resources that should be adoptable.

jaypipes (changes requested):
@RedbackThomson just remove the unused string constants for adopted status and I'll merge this PR.

pkg/runtime/adoption_reconciler.go
RedbackThomson: Correct. This is meant to wrap an error, ensuring that it is implanted into the adopted resource before we raise it. It's a wrapper around `return err` in other methods.

jaypipes (approved):
muy bueno :)

PR #5, opened 2021-02-23
SecretKeyRef - Runtime Implementation

nmvk:
This contains the Secret type, k8 Secrets API call
to fetch the Secret.

Other PRs -
Code gen - https://github.com/aws-controllers-k8s/code-generator/pull/21
ElastiCache generated code -


pkg/runtime/reconciler.go
```diff
@@ -76,6 +77,7 @@ func (r *reconciler) BindControllerManager(mgr ctrlrt.Manager) error {
 	r.kc = mgr.GetClient()
 	r.cache = ackrtcache.New(clientset, r.log)
 	r.cache.Run()
+	r.clientSet = *clientset
```
jaypipes, line 80: You don't need to use the clientset. and yes, I know the cache stuff uses it, but I don't actually like the fact that the cache stuff uses the clientset, which is just syntactic sugar. See below for how to use the controller-runtime's Client interface for secrets.

pkg/runtime/reconciler.go
```diff
@@ -85,12 +87,38 @@ func (r *reconciler) BindControllerManager(mgr ctrlrt.Manager) error {
...
+	namespace := ref.Namespace
+	if namespace == "" {
+		namespace = "default"
+	}
+
+	secret, err := r.clientSet.CoreV1().Secrets(namespace).Get(ctx, ref.Name, metav1.GetOptions{})
+
+	if err != nil {
+		return nil, err
+	}
```
jaypipes, line 109: Instead of the above use of clientset, you can just use the controller-runtime's client, which I think is actually more elegant and shorter:

```go
	nsn := client.ObjectKey{
		Namespace: ref.Namespace,
		Name: ref.Name,
	}
	var secret corev1.Secret
	if err := r.kc.Get(ctx, nsn, &secret); err != nil {
		return nil, err
	}
```

Note that `client.ObjectKey` is simply an alias for apimachinery's `types.NamespacedName`:

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#ObjectKey

pkg/errors/error.go
```diff
@@ -42,6 +42,12 @@ var (
 	// Terminal is returned with resource is in Terminal Condition
 	Terminal = fmt.Errorf(
 		"resource is in terminal condition")
+	// SecretTypeNotSupported is returned if non opaque secret is used.
+	SecretTypeNotSupported = fmt.Errorf(
+		"only opaque secrets can be used")
+	// SecretTypeNotFound is returned if secret does not exist.
```
jaypipes, line 48: s/SecretTypeNotFound/SecretNotFound/

However, I don't think we really need a type-specific NotFound error in this package, do we? Can we just return the `NotFound` error?

pkg/errors/error.go
```diff
@@ -42,6 +42,12 @@ var (
...
+	SecretNotFound = fmt.Errorf(
+		"only opaque secrets can be used")
```
jaypipes, line 50: wrong error message (copy-pasta)

pkg/types/secret.go
```diff
@@ -0,0 +1,27 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
jaypipes, line 1: This file should go in `pkg/apis/core/v1alpha1/secret.go`

jaypipes (changes requested):
Thanks @nmvk! Please see inline for some requested changes.

nmvk:
@jaypipes This does not take care of updates, https://github.com/aws-controllers-k8s/community/issues/700 for further enhancing runtime to handle those cases.

jaypipes (approved):
:+1:

kumargauravsharma (approved):

PR #6, opened 2021-02-24
requeue based on conditions even if resource is in sync

surajkota:
Currently, the reconciler successfully exits (i.e. returns nil) if the latest and desired CR match even if the resource is TempOutofSync condition is True. The controller should:

1. continue to reconcile periodically if the TempOutofSync condition is False
1. update the conditions including its message, state, reason and transition time if needed

This PR fixes point 1.

Related PR: [aws-controllers-k8s/sagemaker-controller#3](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/3)

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -204,29 +204,29 @@ func (r *reconciler) sync(
 		)
 	} else {
 		// Check to see if the latest observed state already matches the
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
:+1: approving this for now. would be good to get a unit test of all this in a later patch (and I'll probably add those tests in my delta-branch stuff, so you're all good.

PR #7, opened 2021-03-03
Allow optional endpoint argument

nmvk:
Most AWS including ElastiCache allows customers to invoke APIs using FIPS
endpoint.

https://docs.aws.amazon.com/general/latest/gr/rande.html#FIPS-endpoints

This commit introduces ability to specify endpoint URL like AWS CLI.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -15,6 +15,7 @@ package config
 
 import (
 	"errors"
+	"net/url"
 
 	flag "github.com/spf13/pflag"
 	"go.uber.org/zap/zapcore"
@@ -29,6 +30,7 @@ const (
 	flagEnableDevLogging     = "enable-development-logging"
 	flagAWSAccountID         = "aws-account-id"
 	flagAWSRegion            = "aws-region"
+	flagAWSEndpointURL       = "aws-endpoint-url"
 	flagLogLevel             = "log-level"
 	flagResourceTags         = "resource-tags"
 )
@@ -41,6 +43,7 @@ type Config struct {
 	EnableDevelopmentLogging bool
 	AccountID                string
 	Region                   string
+	EndpointURL              string
 	LogLevel                 string
 	ResourceTags             []string
 }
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -29,6 +30,7 @@ const (
 	flagEnableDevLogging     = "enable-development-logging"
 	flagAWSAccountID         = "aws-account-id"
 	flagAWSRegion            = "aws-region"
+	flagAWSServiceEndpoint   = "aws-service-endpoint"
```
jaypipes, line 33: If this flag should be a URL (as it seems it needs to be below), this should be called `aws-endpoint-url`.

pkg/config/config.go
```diff
@@ -41,6 +43,7 @@ type Config struct {
 	EnableDevelopmentLogging bool
 	AccountID                string
 	Region                   string
+	ServiceEndpoint          string
```
jaypipes, line 46: s/ServiceEndpoint/EndpointURL/g

jaypipes (changes requested):
I think we should take advantage of the https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/ module, including its `ResolverFunc` interface instead of manually parsing URLs in the startup of the controller.

In addition, we should be using the https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/ module's `Partition` support instead of manually setting endpoint URLs if at all possible...

pkg/runtime/session.go
```diff
@@ -42,6 +43,9 @@ func (c *ServiceController) newSession(
 		STSRegionalEndpoint: endpoints.RegionalSTSEndpoint,
 	}
```
jaypipes, line 44: Add the conditional here instead.

```go
	if endpointURL != nil {
		awsCfg.Endpoint = endpointURL
	}
```

pkg/config/config.go
```diff
@@ -79,6 +82,11 @@ func (cfg *Config) BindFlags() {
 		"",
 		"The AWS Region in which the service controller will create its resources",
 	)
+	flag.StringVar(
+		&cfg.EndpointURL, flagAWSEndpointURL,
+		"",
+		"The AWS endpoint URL the service controller will use to create its resources",
```
jaypipes, line 88: Please add some text that says this is only used to override the default behaviour of aws-sdk-go that constructs endpoint URLs automatically based on service and region.

jaypipes (changes requested):

kumargauravsharma:
As discussed the @jaypipes offline, merging this PR as the requested changes have been addressed by @nmvk

kumargauravsharma (approved):

PR #9, opened 2021-04-05
Adding Recoverable Condition

jkuruba:
Introducing RecoverableCondition for errors that are transient or can be fixed and the sync continues

### DIFF
--- a/apis/core/v1alpha1/conditions.go
+++ b/apis/core/v1alpha1/conditions.go
@@ -32,13 +32,12 @@ const (
 	//		- As a result of InvalidArgument in input yaml
 	//		- Resource server state is "create-failed"
 	ConditionTypeTerminal ConditionType = "ACK.Terminal"
-	// ConditionTypeNonTerminal indicates that the error may be resolved
+	// ConditionTypeRecoverable indicates that the error may be resolved
 	// without needing to update the custom resource spec and sync will continue.
-	// Any error not marked as Terminal will fall in this category.
 	// Examples include:
 	//		- ServiceUnavailable errors that are transient
 	//		- AccessDeniedException that needs correct credentials
-	ConditionTypeNonTerminal ConditionType = "ACK.NonTerminal"
+	ConditionTypeRecoverable ConditionType = "ACK.Recoverable"
 )
 
 // Condition is the common struct used by all CRDs managed by ACK service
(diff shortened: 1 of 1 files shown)


jaypipes (changes requested):
I'm good with this, except after further considering, I think I'd prefer to call this `ConditionTypeRecoverable` instead of `ConditionTypeNonTerminal`. Sorry for going back and forth on this, @jkuruba!

apis/core/v1alpha1/conditions.go
```diff
@@ -32,6 +32,13 @@ const (
 	//		- As a result of InvalidArgument in input yaml
 	//		- Resource server state is "create-failed"
 	ConditionTypeTerminal ConditionType = "ACK.Terminal"
+	// ConditionTypeNonTerminal indicates that the error may be resolved
+	// without needing to update the custom resource spec and sync will continue.
+	// Any error not marked as Terminal will fall in this category.
+	// Examples include:
+	//		- ServiceUnavailable errors that are transient
+	//		- AccessDeniedException that needs correct credentials
+	ConditionTypeNonTerminal ConditionType = "ACK.NonTerminal"
```
surajkota, line 41: Can `ACK.NonTerminal` be confusing to users? should it say `ACK.NonTerminalError`/ `ACK.NonTerminalRateLimiting` or something else ?
because anything apart from `ACK.Terminal` is expected to be NonTerminal e.g. `ACK.ResourceSynced`

jaypipes: My preference is `ACK.Recoverable`. These aren't necessarily errors, per-se, and what we want to convey to Kubernetes users is that the Condition may be removed through no action of their own.

surajkota: +1 `ACK.Recoverable` 

> we want to convey to Kubernetes users is that the Condition may be removed through no action of their own

not necessarily, action might be required depending on the error but user need not have to resubmit the resource. e.g. 
- `accessDenied` user needs to add some permissions to the role.
- `ThrottlingError/ServiceUnavailable` no action required

jaypipes (approved):

apis/core/v1alpha1/conditions.go
jaypipes: > Would you still call it `Recoverable` ?

Yes, since the user does not have to change the Spec in order for the Condition to be recovered from.

PR #10, opened 2021-04-27
Ensure adopted resources are managed

RedbackThomson:
Attempting to solve https://github.com/aws-controllers-k8s/community/issues/746

Sets the adopted resource as managed on first cycle of updates - rather than in the create loop like a typical resource.


pkg/runtime/reconciler.go
```diff
@@ -237,6 +237,14 @@ func (r *resourceReconciler) Sync(
 			"arn", latest.Identifiers().ARN(),
 		)
 	} else {
+		// Ensure that adopted resources are marked as managed during the first
+		// update cycle
+		if isAdopted && !r.rd.IsManaged(desired) {
+			if err = r.setResourceManaged(ctx, desired); err != nil {
```
surajkota, line 243: lgtm, adoption reconciler created the CR and an instance of resource manager marks it as managed when it reconciles aganist it as opposed to marking managed [in adoption reconciler itself](https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/adoption_reconciler.go#L203)

don't need to check for `!r.rd.IsManaged(desired)` as its already checked in `setResourceManged()` and cant think of a reason to check isAdopted as well. Will let @jaypipes comment

jaypipes: Yeah, agree with @surajkota. Should be fine to just do `setResourceManaged` with no conditionals.

jaypipes (changes requested):

surajkota (approved):

surajkota:
some unit tests are failing

surajkota (approved):

jaypipes (approved):

PR #11, opened 2021-05-07
adopt only if APIGroup match

surajkota:
**Return early if the API group for target resource does not match the controller where adopted reconciler is running in**. This will help solve two issues:
  1. Mutiple controller scenario(different services), resource status always set to `ACK.Adopted = False` because multiple controller try to update the Status(backoff and retry) and only one can succeed
  2. If user tries to adopt a resourceKind which is not supported in the controller or they has a typo in Kind, they will see an error

PR does not solve the following issue:
- If user makes an error in APIGroup(does not match any controller), all controllers will go to no-op. None will status populate status/error

-----

**Enchance logging in adoption resource reconciler**

Previously
- Log did not have any information about which resource it was reconciling for
```
2021-05-07T17:23:04.146Z	INFO	ackrt	starting adoption reconciliation
```
With this PR (intentionally added a new line below, not in code)
```
2021-05-07T22:33:45.037Z	INFO	ackrt
starting adoption reconciliation	{"kind": "AdoptedResource", "namespace": "default", "name": "m-ep-2", "generation": 1}
```

https://github.com/aws-controllers-k8s/community/issues/755

### DIFF
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -30,6 +30,7 @@ import (
 	ackmetrics "github.com/aws-controllers-k8s/runtime/pkg/metrics"
 	"github.com/aws-controllers-k8s/runtime/pkg/requeue"
 	ackrtcache "github.com/aws-controllers-k8s/runtime/pkg/runtime/cache"
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -85,6 +86,18 @@ func (r *adoptionReconciler) reconcile(req ctrlrt.Request) error {
 	}
 
 	gk := r.getTargetResourceGroupKind(res)
+
+	// Check if the target API group matches with the controller
+	var controllerRMF acktypes.AWSResourceManagerFactory
+	for _, v := range r.sc.GetResourceManagerFactories() {
+		controllerRMF = v
+		break
+	}
+	if gk.Group != controllerRMF.ResourceDescriptor().GroupKind().Group {
+		ackrtlog.DebugAdoptedResource(r.log, res, "target resource API group is not of this service. no-op")
+		return nil
+	}
+
 	// Look up the rmf for the given target resource GVK
 	rmf, ok := (r.sc.GetResourceManagerFactories())[gk.String()]
 	if !ok {
@@ -111,8 +124,7 @@ func (r *adoptionReconciler) reconcile(req ctrlrt.Request) error {
 		return err
 	}
 
-	// TODO(RedbackThomson): Better logging for adopted resources
-	r.log.Info("starting adoption reconciliation")
+	ackrtlog.InfoAdoptedResource(r.log, res, "starting adoption reconciliation")
 
 	rm, err := rmf.ManagerFor(
 		r.cfg, r.log, r.metrics, r, sess, acctID, region,
@@ -460,7 +472,7 @@ func NewAdoptionReconciler(
 	return &adoptionReconciler{
 		reconciler: reconciler{
 			sc:      sc,
-			log:     log.WithName("ackrt"),
+			log:     log.WithName("adopted-reconciler"),
 			cfg:     cfg,
 			metrics: metrics,
 		},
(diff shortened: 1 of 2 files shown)


pkg/runtime/adoption_reconciler.go
```diff
@@ -85,6 +85,18 @@ func (r *adoptionReconciler) reconcile(req ctrlrt.Request) error {
...
 	gk := r.getTargetResourceGroupKind(res)
+
+	// Check if the target API group matches with the controller
+	var controllerRMF acktypes.AWSResourceManagerFactory
+	for _, v := range r.sc.GetResourceManagerFactories() {
+		controllerRMF = v
+		break
+	}
+	if gk.Group != controllerRMF.ResourceDescriptor().GroupKind().Group {
+		LogAdoptedResource(r.log, res).V(0).Info("target resource API group is not of this service. no-op")
```
RedbackThomson, line 96: I'd make this into a `.Debug`. I'm really trying to avoid spamming logs with information, especially if this is an intentional no-op.

pkg/runtime/adoption_reconciler.go
```diff
@@ -466,3 +478,26 @@ func NewAdoptionReconciler(
...
+// namespace, name, etc
+func LogAdoptedResource(
+	log logr.Logger,
+	res *ackv1alpha1.AdoptedResource,
+	additionalValues ...interface{},
+) logr.Logger {
+	ns := res.Namespace
+	resName := res.Name
+	generation := res.Generation
+	kind := res.Kind
```
jaypipes, line 492: This is always going to be AdoptedResource. I think a more useful included value might be `res.Kubernetes.Kind` and `res.Kubernetes.Group`, which represent the target CR that the adopted resource reconciler will end up creating.

https://github.com/aws-controllers-k8s/runtime/blob/10d1b5339b6030fcaedf65d84e637d12a8779ab1/apis/core/v1alpha1/identifiers.go#L29-L32

pkg/runtime/adoption_reconciler.go
```diff
@@ -466,3 +478,26 @@ func NewAdoptionReconciler(
 		},
 	}
 }
+
+// LogAdoptedResource returns a logger with log values set for the resource's kind,
+// namespace, name, etc
+func LogAdoptedResource(
```
jaypipes, line 484: Can you place this into `pkg/runtime/log` and call it `AdaptAdoptedResource` and follow the same pattern used in there for a `DebugAdoptedResource` and `InfoAdoptedResource` function please?

https://github.com/aws-controllers-k8s/runtime/blob/10d1b5339b6030fcaedf65d84e637d12a8779ab1/pkg/runtime/log/resource.go#L22-L69

jaypipes (changes requested):

pkg/runtime/log/resource.go
```diff
@@ -67,3 +68,48 @@ func InfoResource(
...
+// resource's kind, namespace, name, etc
+func AdaptAdoptedResource(
+	log logr.Logger,
+	res *v1alpha1.AdoptedResource,
+	additionalValues ...interface{},
+) logr.Logger {
+	ns := res.Namespace
+	resName := res.Name
+	generation := res.Generation
+	kind := res.Kind
```
jaypipes, line 82: should be `res.Kubernetes.Kind`, right?

jaypipes (changes requested):

pkg/runtime/adoption_reconciler.go
surajkota: Can we add this in future?

this will add value when we have more detailed error messages

pkg/runtime/adoption_reconciler.go
```diff
@@ -460,7 +472,7 @@ func NewAdoptionReconciler(
 	return &adoptionReconciler{
 		reconciler: reconciler{
 			sc:      sc,
-			log:     log.WithName("ackrt"),
+			log:     log.WithName("adopted-reconciler"),
```
jaypipes, line 475: :+1:

pkg/runtime/log/resource.go
```diff
@@ -67,3 +68,50 @@ func InfoResource(
...
+	additionalValues ...interface{},
+) logr.Logger {
+	ns := res.Namespace
+	resName := res.Name
+	generation := res.Generation
+	group := res.Spec.Kubernetes.Group
+	kind := res.Spec.Kubernetes.Kind
+	vals := []interface{}{
+		"target_group", group,
+		"target_kind", kind,
```
jaypipes, line 86: :+1: perfect.

jaypipes (approved):
Awesome, thank you @surajkota!

surajkota:
@jaypipes addressed comments

PR #12, opened 2021-05-11
Namespace scope service controllers

akartsky:
Add ability to listen to specific namespace for service controllers 

https://github.com/aws-controllers-k8s/community/issues/686

Related PR:
https://github.com/aws-controllers-k8s/code-generator/pull/60

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -33,7 +33,7 @@ const (
 	flagAWSEndpointURL       = "aws-endpoint-url"
 	flagLogLevel             = "log-level"
 	flagResourceTags         = "resource-tags"
-	flagListenNamespace      = "listen-namespace"
+	flagWatchNamespace       = "watch-namespace"
 )
 
 // Config contains configuration otpions for ACK service controllers
@@ -47,7 +47,7 @@ type Config struct {
 	EndpointURL              string
 	LogLevel                 string
 	ResourceTags             []string
-	ListenNamespace          string
+	WatchNamespace           string
 }
 
 // BindFlags defines CLI/runtime configuration options
@@ -102,9 +102,9 @@ func (cfg *Config) BindFlags() {
 		"Configures the ACK service controller to always set key/value pairs tags on resources that it manages.",
 	)
 	flag.StringVar(
-		&cfg.ListenNamespace, flagListenNamespace,
+		&cfg.WatchNamespace , flagWatchNamespace,
 		"",
-		"Specific namespace the service controller will listen to for object creation from CRD. "+ 
+		"Specific namespace the service controller will watch for object creation from CRD. "+ 
 			" By default it will listen to all namespaces",
 	)
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -33,6 +33,7 @@ const (
 	flagAWSEndpointURL       = "aws-endpoint-url"
 	flagLogLevel             = "log-level"
 	flagResourceTags         = "resource-tags"
+	flagListenNamespace      = "listen-namespace"
 )
 
 // Config contains configuration otpions for ACK service controllers
@@ -46,6 +47,7 @@ type Config struct {
 	EndpointURL              string
 	LogLevel                 string
 	ResourceTags             []string
+	ListenNamespace          string
 }
 
 // BindFlags defines CLI/runtime configuration options
@@ -98,6 +100,12 @@ func (cfg *Config) BindFlags() {
 		&cfg.ResourceTags, flagResourceTags,
 		[]string{},
 		"Configures the ACK service controller to always set key/value pairs tags on resources that it manages.",
+	)
+	flag.StringVar(
+		&cfg.ListenNamespace, flagListenNamespace,
+		"",
+		"Specific namespace the service controller will listen to for object creation from CRD. "+ 
+			" By default it will listen to all namespaces",
 	)
 }
 
--- a/pkg/runtime/cache/cache.go
+++ b/pkg/runtime/cache/cache.go
@@ -58,10 +58,10 @@ type Caches struct {
 
 // New creates a new Caches object from a kubernetes.Interface and
 // a logr.Logger
-func New(clientset kubernetes.Interface, log logr.Logger) Caches {
+func New(clientset kubernetes.Interface, log logr.Logger, listenNamespace string) Caches {
 	return Caches{
 		Accounts:   NewAccountCache(clientset, log),
-		Namespaces: NewNamespaceCache(clientset, log),
+		Namespaces: NewNamespaceCache(clientset, log, listenNamespace),
 	}
 }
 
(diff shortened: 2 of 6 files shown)


pkg/runtime/cache/namespace.go
```diff
@@ -72,40 +76,45 @@ func NewNamespaceCache(clientset kubernetes.Interface, log logr.Logger) *Namespa
...
-	return ok &&
-		(object.ObjectMeta.Name == "ack-system" ||
-			object.ObjectMeta.Name == "kube-system" ||
-			object.ObjectMeta.Name == "kube-public")
+	if !ok {
+		return false
+	} 
+
+	if listenNamespace == "" {
+		return true
```
RedbackThomson, line 92: I think it's still valid to check that the namespace is not in `kube-system` or `kube-public` (`ack-system` is probably fine?).

pkg/runtime/cache/namespace.go
```diff
@@ -72,40 +76,45 @@ func NewNamespaceCache(clientset kubernetes.Interface, log logr.Logger) *Namespa
...
+	} else {
```
RedbackThomson, line 93: nit: unnecessary `else`

pkg/config/config.go
```diff
@@ -99,6 +101,11 @@ func (cfg *Config) BindFlags() {
 		[]string{},
 		"Configures the ACK service controller to always set key/value pairs tags on resources that it manages.",
 	)
+	flag.StringVar(
+		&cfg.ListenNamespace, flagListenNamespace,
+		"",
+		"The namespace the controller will listen to.",
```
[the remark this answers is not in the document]
surajkota: +1 rename it to watchNamespace? sounds more k8s related

pkg/runtime/cache/cache.go
```diff
@@ -58,10 +58,10 @@ type Caches struct {
 
 // New creates a new Caches object from a kubernetes.Interface and
 // a logr.Logger
-func New(clientset kubernetes.Interface, log logr.Logger) Caches {
+func New(clientset kubernetes.Interface, log logr.Logger, ListenNamespace string) Caches {
```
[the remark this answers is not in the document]
akartsky: done

pkg/runtime/cache/namespace.go
akartsky: done

pkg/config/config.go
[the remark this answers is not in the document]
akartsky: Added more info. 

ya i can change it to watchNamespace

pkg/config/config.go
```diff
@@ -99,6 +101,12 @@ func (cfg *Config) BindFlags() {
 		[]string{},
 		"Configures the ACK service controller to always set key/value pairs tags on resources that it manages.",
 	)
+	flag.StringVar(
+		&cfg.ListenNamespace, flagListenNamespace,
+		"",
+		"Specific namespace the service controller will listen to for object creation from CRD. "+ 
+			" By default it will listen to all namespaces",
```
A-Hilaly, line 108: you can also add that in the default case, the controller will ignore the `kube-system`, `kube-public` and `ack-system` namespaces

pkg/runtime/cache/namespace.go
```diff
@@ -55,6 +55,10 @@ type NamespaceCache struct {
 	sync.RWMutex
 
 	log logr.Logger
+	// Provide a namespace specifically to listen to. 
+	// Provide empty string to listen to all namespaces.
```
A-Hilaly, line 59: ditto

pkg/runtime/cache/namespace.go
A-Hilaly: IMO we should keep ignoring `ack-system`... But i'm still hesitant about this - what if a user installs a controller in a different namespace than `ack-system`?

pkg/runtime/cache/namespace.go
```diff
@@ -72,40 +76,49 @@ func NewNamespaceCache(clientset kubernetes.Interface, log logr.Logger) *Namespa
...
+		return false
+	} 
+
+	if listenNamespace == object.ObjectMeta.Name {
+		return true
+	} else if object.ObjectMeta.Name == "kube-system" || object.ObjectMeta.Name == "kube-public" {
+		return false
+	} else if listenNamespace == "" {
+		return true
+	}
```
A-Hilaly, line 97: nit: we can simplify this if/else block to something like:
```go
	if listenNamespace != "" {
		return listenNamespace == object.ObjectMeta.Name
	}
	return object.ObjectMeta.Name != "kube-system" && object.ObjectMeta.Name != "kube-public"
```

pkg/config/config.go
akartsky: Nice catch but actually the controller will listen to all the namespaces by default 
only the namespace cache will ignore kube-system and kube-public 

added comment to `namespace cache`

pkg/runtime/cache/namespace.go
akartsky: done

pkg/runtime/cache/namespace.go
akartsky: add the comment

A-Hilaly (approved):
:+1:

pkg/config/config.go
```diff
@@ -33,6 +33,7 @@ const (
 	flagAWSEndpointURL       = "aws-endpoint-url"
 	flagLogLevel             = "log-level"
 	flagResourceTags         = "resource-tags"
+	flagWatchNamespace       = "watch-namespace"
```
jaypipes, line 36: Can you make this `watch-namespaces` and allow for multiple string args please? (see how `flagResourceTags` is created...)

pkg/config/config.go
```diff
@@ -99,6 +101,12 @@ func (cfg *Config) BindFlags() {
 		[]string{},
 		"Configures the ACK service controller to always set key/value pairs tags on resources that it manages.",
 	)
+	flag.StringVar(
+		&cfg.WatchNamespace , flagWatchNamespace,
+		"",
+		"Specific namespace the service controller will watch for object creation from CRD. "+
```
jaypipes, line 107: Change the above to read:

"Zero or more Kubernetes namespaces the controller will watch for changes in."

The reason is because this change does *not* make the controller only watch for *objects* being created in specific namespaces. What this change does is make the controller only watch for specific *namespaces* having their annotations changed. There's a big difference between the two things.

jaypipes (changes requested):

pkg/config/config.go
jaypipes: Ignore this comment. I forgot upstream `ctrlrt.NewManager` call doesn't support >1 namespace.

pkg/config/config.go
jaypipes: Ignore the above.

jaypipes (approved):

PR #13, opened 2021-05-12
Add additional keys to adopted resource spec

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/766

Some resources require more than an ARN/Name/ID to describe a single resource. This PR adds an `additionalKeys` field to the `AWSIdentifiers` to allow for arbitrary identifiers as needed by any resources.

Generator code to support this change: https://github.com/aws-controllers-k8s/code-generator/pull/62


A-Hilaly:
@RedbackThomson is gonna rework this PR, sometime next week

/hold

apis/core/v1alpha1/identifiers.go
```diff
@@ -21,6 +21,9 @@ type AWSIdentifiers struct {
 	// NameOrId is a user-supplied string identifier for the resource. It may
 	// or may not be globally unique, depending on the type of resource.
 	NameOrID *string `json:"nameOrID,omitempty"`
+	// AdditionalKeys represents any additional arbitrary identifiers used when
+	// describing the target resource.
+	AdditionalKeys map[string]*string `json:"additionalKeys,omitempty"`
```
jaypipes, line 26: Any reason this is `map[string]*string` instead of `map[string]string`?

RedbackThomson: I was using the other `*string` fields as reference. Is there a hard and fast rule to when we should use `*string` over `string`?

jaypipes: The use of `map[string]*string` is really an aws-sdk-go thing. Since this isn't an aws-sdk-go shape field, I just thought it would make more sense to use a standard `map[string]string`. Pointer types generally signal optional fields (except for aws-sdk-go, which makes everything a pointer type, unfortunately).

RedbackThomson: Fair enough. I've updated all unnecessary `*string` to `string` types for the adopted resource APIs.

jaypipes (approved):
++

PR #17, opened 2021-05-20
Add new condition ACK.Advisory

nmvk:
New condition ACK.Advisory to set the condition
when any immutable field is modified.

This is needed for https://github.com/aws-controllers-k8s/code-generator/pull/70

### DIFF
--- a/apis/core/v1alpha1/conditions.go
+++ b/apis/core/v1alpha1/conditions.go
@@ -41,6 +41,10 @@ const (
 	//		- ServiceUnavailable errors that are transient
 	//		- AccessDeniedException that needs correct credentials
 	ConditionTypeRecoverable ConditionType = "ACK.Recoverable"
+	// ConditionTypeAdvisory indicates any advisory info that may be present in the resource.
+	// Examples include
+	//      - Modifying an immutable field after it was created
+	ConditionTypeAdvisory ConditionType = "ACK.Advisory"
 )
 
 // Condition is the common struct used by all CRDs managed by ACK service


apis/core/v1alpha1/conditions.go
```diff
@@ -41,6 +41,10 @@ const (
 	//		- ServiceUnavailable errors that are transient
 	//		- AccessDeniedException that needs correct credentials
 	ConditionTypeRecoverable ConditionType = "ACK.Recoverable"
+	// ConditionTypeWarning indicates any warning that may be present in the resource.
+	// Examples include
+	//      - Modifying an immutable field after it was created
+	ConditionTypeWarning ConditionType = "ACK.Warning"
```
jaypipes, line 47: I actually would prefer to call this `ConditionTypeInfo` or `ConditionTypeAdvisory` instead of `ConditionTypeWarning`. The word "warning" is generally reserved for serious things that should be addressed ASAP by an operator. In the case of an immutable field being modified in the `Spec` of a resource, that's definitely not something that needs to be addressed ASAP by an operator (since it's not dangerous or anything)

jaypipes (changes requested):

apis/core/v1alpha1/conditions.go
nmvk: Changed this to `ConditionTypeAdvisory` as someone may ignore `ConditionTypeInfo` without noticing

RedbackThomson (approved):

jaypipes (approved):
:+1:

PR #18, opened 2021-05-26
Add `pkg/compare/Path.MarshalJSON` function to implement `json.Marshaler` interface

A-Hilaly:
Issue https://github.com/aws-controllers-k8s/community/issues/772

Description of changes:
- Add `MarshalJSON` function method to help encoding `pkg/comapare.Path` unexported fields

### DIFF
--- a/pkg/compare/path.go
+++ b/pkg/compare/path.go
@@ -13,12 +13,30 @@
 
 package compare
 
-import "strings"
+import (
+	"encoding/json"
+	"strings"
+)
 
 // Path provides a JSONPath-like struct and field-member "route" to a
-// particular field within a compared struct
+// particular field within a compared struct. Path implements json.Marshaler
+// interface.
 type Path struct {
 	parts []string
+}
+
+// MarshalJSON returns the JSON encoding of a Path object.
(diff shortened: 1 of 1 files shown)


A-Hilaly:
If exporting the `parts` field is a somehow a problem, another solution would be to implement the `json.Marshaller` interface. Something similar to:
```go
func (p *Path) MarshalJSON() ([]byte, error) {
	return json.Marshal(
		struct{
                         Parts []string
                }{
                         p.parts,
                },
	)
}
```

RedbackThomson:
> If exporting the `parts` field is a somehow a problem, another solution would be to implement the `json.Marshaller` interface. Something similar to:
> 
> ```go
> func (p *Path) MarshalJSON() ([]byte, error) {
> 	return json.Marshal(
> 		struct{
>                          Parts []string
>                 }{
>                          p.parts,
>                 },
> 	)
> }
> ```

I almost like this idea more, because it's very explicit about how we are using parts with JSON. Relying on casing seems like a great way for someone to accidentally break logging in the future when they determine it "doesn't need to be exported".

PR #20, opened 2021-06-04
fix Delta.DifferentAt behavior via changes to Path.Contains

echen-98:
Fixes issue https://github.com/aws-controllers-k8s/community/issues/808

### DIFF
--- a/pkg/compare/delta_test.go
+++ b/pkg/compare/delta_test.go
@@ -53,11 +53,14 @@ func TestDifferentAt(t *testing.T) {
 	d = compare.NewDelta()
 	d.Add("Bar", a.Bar, b.Bar)
 	require.True(d.DifferentAt("Bar"))
-	require.False(d.DifferentAt("Baz"))
+	require.False(d.DifferentAt("Baz")) // diff exists but was not added to Delta
 
 	d = compare.NewDelta()
 	d.Add("Baz.Y", a.Baz.Y, b.Baz.Y)
 	require.True(d.DifferentAt("Baz"))
-	require.True(d.DifferentAt("Y"))
-	require.False(d.DifferentAt("Bar"))
+	require.True(d.DifferentAt("Baz.Y"))
+	require.False(d.DifferentAt("Y")) // there is no diff for top-level field "Y"
+	require.False(d.DifferentAt("Bar")) // diff exists but it was not added to Delta
+	require.False(d.DifferentAt("Baz.Y.Z")) // subject length exceeds length of diff Path
+	require.False(d.DifferentAt("Baz.Z"))  // matches Path top-level field but not sub-field
 }
--- a/pkg/compare/path.go
+++ b/pkg/compare/path.go
@@ -51,14 +51,28 @@ func (p Path) Pop() {
 	}
 }
 
-// Contains returns true if the supplied string appears within the Path
+//	Contains returns true if the supplied string, delimited on ".", matches
+//	p.parts up to the length of the supplied string.
+//		e.g. if the Path p represents "A.B":
+//			subject "A" -> true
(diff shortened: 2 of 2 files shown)


pkg/compare/path.go
```diff
@@ -51,14 +51,30 @@ func (p Path) Pop() {
 	}
 }
 
-// Contains returns true if the supplied string appears within the Path
+// TODO: case sensitivity?
```
echen-98, line 54: should we keep case-sensitive comparison?

pkg/compare/delta_test.go
```diff
@@ -53,11 +53,14 @@ func TestDifferentAt(t *testing.T) {
...
 	d = compare.NewDelta()
 	d.Add("Baz.Y", a.Baz.Y, b.Baz.Y)
 	require.True(d.DifferentAt("Baz"))
-	require.True(d.DifferentAt("Y"))
-	require.False(d.DifferentAt("Bar"))
+	require.True(d.DifferentAt("Baz.Y"))
+	require.False(d.DifferentAt("Y")) // there is no diff for top-level field "Y"
+	require.False(d.DifferentAt("Bar")) // diff exists but it was not added to Delta
+	require.False(d.DifferentAt("Baz.Y.Z")) // subject length exceeds length of diff Path
+	require.False(d.DifferentAt("Baz.Z"))  // matches Path top-level field but not sub-field
```
RedbackThomson, line 65: ++ Super clear test cases! Love it!

pkg/compare/path.go
RedbackThomson: I think so, yes. We expect case-sensitivity in all Go fields and I'd expect that these strings are reflections of Go fields.

RedbackThomson (approved):

pkg/compare/path.go
echen-98: Ok, removed TODO comment

pkg/compare/path.go
```diff
@@ -51,14 +51,29 @@ func (p Path) Pop() {
...
-// Contains returns true if the supplied string appears within the Path
+/* Contains returns true if the supplied string, delimited on ".", matches
+	p.parts up to the length of the supplied string.
+	e.g. if the Path p represents "A.B":
+		subject "A" -> true
+		subject "A.B" -> true
+		subject "A.B.C" -> false
+		subject "B" -> false
+		subject "A.C" -> false
+ */
```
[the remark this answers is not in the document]
A-Hilaly: You can also instruct `godoc` to generate testable examples in the package documentation: https://blog.golang.org/examples / https://blog.golang.org/godoc

A-Hilaly (commented):
Very nice, thanks Eric!
I left one comment below

pkg/compare/delta_test.go
jaypipes: agreed, excellent clarity @echen-98

pkg/compare/path.go
```diff
@@ -51,14 +51,28 @@ func (p Path) Pop() {
 	}
 }
 
-// Contains returns true if the supplied string appears within the Path
+//	Contains returns true if the supplied string, delimited on ".", matches
+//	p.parts up to the length of the supplied string.
```
jaypipes, line 55: femto-nit: not sure if it's your IDE or editor changing the above, but comments should have a single space (not `\t`) after the `//` and before the first letter of the comment...

jaypipes (approved):
Great stuff @echen-98 :) Thank you!

pkg/compare/path.go
echen-98: Got it, will keep in mind for next time

PR #21, opened 2021-06-16
Returning the latest resource from the Delete() call

kumargauravsharma:
Related issue: https://github.com/aws-controllers-k8s/community/issues/836

This PR contains code changes to patch custom resource with latest details during long running delete.
It does so if resource manager's Delete function return requeue related errors. 
The expectation is that resource manager's ReadOne method would set appropriate conditions (example: `ACK.ResourceSynced` to false with message) and provide latest details about the resource which is being deleted.
Conditions can also be set at some common place where common conditions logic should reside.

Testing:
`make test` passed for runtime.

**Alternate approach** that was considered, but was not taken up as it involved non-intrusive changes to resource manager's APIs. Following are the details:
Currently, [resource manager's `Delete` API](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/pkg/resource/manager.go.tpl#L136) returns only one data field of type `error`, it does not return `resource`.
Also, the [`sdk.go::sdkDelete()` logic](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/pkg/resource/sdk.go.tpl#L141) does not parse the return result output of service delete API.

It can be updated such that `sdk.go::sdkDelete()` logic parses the output response, and allows a `sdk_delete_post_set_output` hook for service controllers to have custom logic (as needed). The resultant resource reference would then be returned from this method.
In case, requeue is required then the return values from this method will have both resource, error (requeue type) as not nil.
And the [reconciler:cleanup() logic here](https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/reconciler.go#L327), would patch the current resource with the returned resource from `rm.Delete()` call and return the error (if any).

Though this change is similar to the approach taken for [resource manager ReadOne here](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/pkg/resource/manager.go.tpl#L76-L78), this is more intrusive as it changes the Delete API's return type.
However, it does have advantage that there is no need for another ReadOne call from reconciler, and also provides flexibility to service controller's to implement custom hook specific to delete scenario.

[EDIT]: the alternate approach that is mentioned above, is the final approach taken per the discussion/comments in this PR.
Related PR in code-generator: https://github.com/aws-controllers-k8s/code-generator/pull/114

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -349,26 +349,34 @@ func (r *resourceReconciler) cleanup(
 	defer exit(err)
 
 	rlog.Enter("rm.ReadOne")
-	latest, err := rm.ReadOne(ctx, current)
+	observed, err := rm.ReadOne(ctx, current)
 	rlog.Exit("rm.ReadOne", err)
 	if err != nil {
 		if err == ackerr.NotFound {
 			// If the aws resource is not found, remove finalizer
-			return r.setResourceUnmanaged(ctx, latest)
+			return r.setResourceUnmanaged(ctx, current)
 		}
 		return err
 	}
 	rlog.Enter("rm.Delete")
-	err = rm.Delete(ctx, latest)
+	latest, err := rm.Delete(ctx, observed)
 	rlog.Exit("rm.Delete", err)
+	if latest != nil {
+		// The Delete operation is likely asynchronous and has likely set a Status
+		// field on the returned CR to something like `deleting`. Here, we patchResource()
+		// in order to save these Status field modifications.
+		_ = r.patchResource(ctx, current, latest)
+	}
 	if err != nil {
+		// NOTE: Delete() implementations that have asynchronously-completing
+		// deletions should return a RequeueNeededAfter.
 		return err
 	}
 
 	// Now that external AWS service resources have been appropriately cleaned
 	// up, we remove the finalizer representing the CR is managed by ACK,
 	// allowing the CR to be deleted by the Kubernetes API server
-	if err = r.setResourceUnmanaged(ctx, latest); err != nil {
+	if err = r.setResourceUnmanaged(ctx, current); err != nil {
 		return err
 	}
 	rlog.Info("deleted resource")
--- a/mocks/pkg/types/aws_resource_manager.go
+++ b/mocks/pkg/types/aws_resource_manager.go
@@ -55,17 +55,26 @@ func (_m *AWSResourceManager) Create(_a0 context.Context, _a1 types.AWSResource)
 }
 
 // Delete provides a mock function with given fields: _a0, _a1
-func (_m *AWSResourceManager) Delete(_a0 context.Context, _a1 types.AWSResource) error {
+func (_m *AWSResourceManager) Delete(_a0 context.Context, _a1 types.AWSResource) (types.AWSResource, error) {
 	ret := _m.Called(_a0, _a1)
 
-	var r0 error
-	if rf, ok := ret.Get(0).(func(context.Context, types.AWSResource) error); ok {
+	var r0 types.AWSResource
+	if rf, ok := ret.Get(0).(func(context.Context, types.AWSResource) types.AWSResource); ok {
 		r0 = rf(_a0, _a1)
 	} else {
-		r0 = ret.Error(0)
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).(types.AWSResource)
+		}
 	}
 
-	return r0
+	var r1 error
+	if rf, ok := ret.Get(1).(func(context.Context, types.AWSResource) error); ok {
+		r1 = rf(_a0, _a1)
+	} else {
+		r1 = ret.Error(1)
+	}
+
+	return r0, r1
 }
 
 // ReadOne provides a mock function with given fields: _a0, _a1
--- a/pkg/types/aws_resource_manager.go
+++ b/pkg/types/aws_resource_manager.go
@@ -59,8 +59,9 @@ type AWSResourceManager interface {
 	) (AWSResource, error)
 
 	// Delete attempts to destroy the supplied AWSResource in the backend AWS
-	// service API.
-	Delete(context.Context, AWSResource) error
+	// service API, returning an AWSResource representing the
+	// resource being deleted (if delete is asynchronous and takes time)
+	Delete(context.Context, AWSResource) (AWSResource, error)
 	// ARNFromName returns an AWS Resource Name from a given string name. This
 	// is useful for constructing ARNs for APIs that require ARNs in their
 	// GetAttributes operations but all we have (for new CRs at least) is a


pkg/runtime/reconciler.go
```diff
@@ -325,6 +325,17 @@ func (r *resourceReconciler) cleanup(
 		return err
 	}
 	if err = rm.Delete(ctx, observed); err != nil {
+		var requeueNeededAfter *requeue.RequeueNeededAfter
+		var requeueNeeded *requeue.RequeueNeeded
+		if errors.As(err, &requeueNeededAfter) ||
+			errors.As(err, &requeueNeeded) {
```
A-Hilaly, line 331: I'm not in favor of using `errors.As` here, first because it's using the `reflect` library and second because it's a code that can panic if misused https://golang.org/src/errors/wrap.go?s=2494:2537#L67

`requeueNeededAfter` and `requeueNeeded` are nil pointers which will cause `errors.As` to panic

A-Hilaly: You can use Go type assertions/switches for this use case. Something similar to:

```go
_, ok := err.(*requeue.RequeueNeededAfter)
if ok {
    // ...
}
```

pkg/runtime/reconciler.go
```diff
@@ -325,6 +325,17 @@ func (r *resourceReconciler) cleanup(
...
+			// read delete progress and corresponding conditions
+			// and patch resource to surface the delete progress.
+			observed, _ := rm.ReadOne(ctx, current)
+			if observed != nil {
+				_ = r.patchResource(ctx, current, observed)
+			}
```
A-Hilaly, line 337: Why ignore the returned errors here?

jaypipes:
@kumargauravsharma if a call to DeleteReplicationGroup returns success, can the ReplicationGroup ever end up in any state other than deleted?

pkg/runtime/reconciler.go
kumargauravsharma: Updated code per the comment.

pkg/runtime/reconciler.go
kumargauravsharma: This block of code executes when Delete api returned error and requeue is requested. ReadOne error is ignored as during the next reconciler loop run (due to requeue) the correct course of action would be taken.
Error from path resources are ignored throughout in the reconciler at other places as well, thus followed the same.

kumargauravsharma:
@jaypipes no error from DeleteReplicationGroup API call set the replication group state as 'deleting'. It may take some time for it to get deleted, meanwhile the resource describe api does return the resource details as delete progresses.

jaypipes:
> @jaypipes no error from DeleteReplicationGroup API call set the replication group state as 'deleting'. It may take some time for it to get deleted, meanwhile the resource describe api does return the resource details as delete progresses.

(apologies for the late reply, I'm on PTO this week...)

@kumargauravsharma ok, thanks for the info. I just want to make sure I'm understanding your use cases clearly...you want this functionality in order to match the Kubernetes API experience to the Elasticache API experience: that the replication group resource continues to appear in a DescribeReplicationGroup API call -- and the ReplicationGroup CR will continue to appear in calls to `kubectl describe replicationgroups/{name}` with a `Status.Status` == `deleting` -- even after a DeleteReplicationGroup API call returns no error and the replication group will only ever transition to a `deleted` state (and not appear in a DescribeReplicationGroup API call). This is about allowing the ReplicationGroup CR to exist for that period of time while it is in progress of being deleted, yes?

pkg/runtime/reconciler.go
jaypipes: > This block of code executes when Delete api returned error and requeue is requested. ReadOne error is ignored as during the next reconciler loop run (due to requeue) the correct course of action would be taken.
> Error from path resources are ignored throughout in the reconciler at other places as well, thus followed the same.

@kumargauravsharma I think @A-Hilaly was referring to ignoring the returned error from `rm.ReadOne()`, not patchResource... (I think?) :)

pkg/runtime/reconciler.go
```diff
@@ -325,6 +325,16 @@ func (r *resourceReconciler) cleanup(
 		return err
 	}
 	if err = rm.Delete(ctx, observed); err != nil {
```
jaypipes, line 327: Instead of adding in the logic below to call `rm.ReadOne()`, how about changing the `rm.Delete()` method signature to return `(acktypes.AWSResource, error)` similar to the `acktypes.AWSResourceManager:ReadOne()`, `Create()` and `Update()` methods?

That way, `rm.Delete()` implementations for asynchronously-completing delete operations (like Elasticache's RGs) could simply return a modified `AWSResource` containing the fields that should be saved via `patchResource()`? Then the below code could be simplified to this:

```go
latest, err = rm.Delete(ctx, observed)
if latest != nil {
    // The Delete operation is likely asynchronous and has likely set a Status
    // field on the returned CR to something like `deleting`. Here, we patchResource()
    // in order to save these Status field modifications.
    _ = r.patchResource(ctx, latest)
}
if err != nil {
    // NOTE: Delete() implementations that have asynchronously-completing
    // deletions should return a RequeueNeededAfter.
    return err
}
```

thoughts?

jaypipes (commented):
@kumargauravsharma please see inline for a suggested alternative approach that I think would align our Delete method more with the other ReadOne/Create/Update resource manager methods...

pkg/runtime/reconciler.go
kumargauravsharma: I thought the comment was for both. 
For error during read one: 
  There are two errors available at that time a) error returned from Delete() method b) error returned from Delete call.
  The code returns the error received from Delete call as that is the primary operation being done in this method.
  ReadOne error is ignored as during the next reconciler loop run (due to requeue) the correct course of action would be taken.

A-Hilaly: IMO we should at least log these errors. Tt would be hard to track them if the metrics server is reporting few 4XX/5XX but logs are not showing any

A-Hilaly: > IMO we should at least log these errors. It would be hard to track them if the metrics server is reporting few 4XX/5XX but logs are not showing any

nvm this was added in the recent logging PR :)

pkg/runtime/reconciler.go
kumargauravsharma: I gave this approach a thought while creating this PR and mentioned its pros, cons in the PR description at that time, please refer section with text in PR description:
> Alternate approach that was considered ...

Downside of this approach was the scope of the changes that it brings: changes to the API signature of resource manager delete, sdkDelete method, and controllers would need to be regenerated with unit tests for delete and code generator templates.

However, I prefer this approach and this PR can be updated for it.

kumargauravsharma:
> > @jaypipes no error from DeleteReplicationGroup API call set the replication group state as 'deleting'. It may take some time for it to get deleted, meanwhile the resource describe api does return the resource details as delete progresses.
> 
> (apologies for the late reply, I'm on PTO this week...)
> 
> @kumargauravsharma ok, thanks for the info. I just want to make sure I'm understanding your use cases clearly...you want this functionality in order to match the Kubernetes API experience to the Elasticache API experience: that the replication group resource continues to appear in a DescribeReplicationGroup API call -- and the ReplicationGroup CR will continue to appear in calls to `kubectl describe replicationgroups/{name}` with a `Status.Status` == `deleting` -- even after a DeleteReplicationGroup API call returns no error and the replication group will only ever transition to a `deleted` state (and not appear in a DescribeReplicationGroup API call). This is about allowing the ReplicationGroup CR to exist for that period of time while it is in progress of being deleted, yes?

The PR: https://github.com/aws-controllers-k8s/elasticache-controller/pull/34 takes care of ensuring that the ReplicationGroup CR exists for the period of time while delete is in progress.
However, the ReplicationGroup CR becomes out of sync from its latest state on service side. Thus, this PR #21 was to fix that behavior of reconciler.

nit: once the ReplicationGroup gets deleted on service side, the sdkFind returns not found exception, as there is no ' deleted' status.

pkg/runtime/reconciler.go
jaypipes: > I gave this approach a thought while creating this PR and mentioned its pros, cons in the PR description at that time, please refer section with text in PR description:
> 
> > Alternate approach that was considered ...

Indeed you did :) Sorry I missed that.

> However, I prefer this approach of changing API return type and this PR can be updated for it.

OK, cool. Let's cut a release of runtime (and probably code-generator) with the enhancements made over the last couple days and then modify this PR to change the Delete() method signature.

Work for you?

kumargauravsharma: Sure, thanks Jay!

surajkota: @kumargauravsharma any plan to work on this PR soon?

kumargauravsharma: I have updated this PR with the approach that we concluded.
Corresponding code-generator changes are available in PR: https://github.com/aws-controllers-k8s/code-generator/pull/114

surajkota (approved):

jaypipes (approved):
thanks for your patience @kumargauravsharma.

PR #24, opened 2021-06-20
add common ACK pkg/condition

jaypipes:
Adds common resource condition functions to a new `pkg/condition`:

- `Synced()` returns the resource's Condition of type
ConditionTypeResourceSynced, or nil if the condition isn't found.
- `Terminal()` returns the resource's Condition of type
ConditionTypeTerminal, or nil if the condition isn't found.
- `FirstOfType()` returns the first Condition of the specified type, or
  nil if a condition of the type isn't found on the resource.
- `AllOfType()` returns a slice of `Condition` having the specified
  condition type.
- `SetSynced()` ensures that a Condition of type
  ConditionTypeResourceSynced is present in the resource's Conditions
  collection and has a specified status, message and reason
- `SetTerminal()` ensures that a Condition of type
  ConditionTypeResourceTerminal is present in the resource's Conditions
  collection and has a specified status, message and reason

**IMPORTANT COMPATIBILITY NOTE**:

Note that the `pkg/types.AWSResource` interface has been modified to be
a composition of a new `pkg/types.ConditionManager` interface which
adds a new `ReplaceConditions()` method that overwrites a resource's
Conditions collection. This is a backwards-incompatible change since the
`pkg/types.AWSResource` implementations generated by
`aws-controllers-k8s/code-generator` (the `pkg/{RESOURCE/resource.go`
files...}) will not include this new `ReplaceConditions` implementation
and thus won't compile if you attempt to regenerate a controller using
the ACK runtime after this patch is merged. A corresponding patch to the
code-generator is coming that adds the `ReplaceCondiions`
implementation.


pkg/types/manages_conditions.go
```diff
@@ -0,0 +1,28 @@
...
+
+package types
+
+import (
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+)
+
+// ManagesConditions describes a thing that can set and retrieve Condition
+// objects.
+type ManagesConditions interface {
```
A-Hilaly, line 22: nit: s/`ManagesConditions`/`ConditionsManager`/

jaypipes: > nit: s/`ManagesConditions`/`ConditionsManager`/

@A-Hilaly @RedbackThomson fine, changed the name :)

PR #26, opened 2021-06-25
Add a `pkg/webhook` package to standardize the way controllers create and register new webhooks

A-Hilaly:
Adding a new package to systematize the way we create and register new
webhooks.

Introducing two files:
- `webhook.go` which contains the webhook structure and defines the
setup method that needs to be suplied by each CRD per APIVersion.
- `register.go` similarily to what a controller
`pkg/resource/register.go` does, this files exposes functions to register
and list the registered webhooks.

### DIFF
--- a/pkg/webhook/registry.go
+++ b/pkg/webhook/registry.go
@@ -1,0 +1,48 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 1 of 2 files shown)


pkg/webhook/registry.go
```diff
@@ -0,0 +1,48 @@
...
+// the same unique identifier.
+func RegisterWebhook(w *Webhook) {
+	if webhooksRegistry == nil {
+		webhooksRegistry = make(map[string]*Webhook)
+	}
+
+	_, ok := webhooksRegistry[w.UID()]
+	if ok {
+		msg := fmt.Sprintf("webhook %s already registred.", w.UID())
+		panic(msg)
```
vijtrip2, line 45: Should this panic or just return `false` ?

vijtrip2 (approved):
Looks good but just one comment.

RedbackThomson (approved):

pkg/webhook/registry.go
A-Hilaly: @vijtrip2 Mmmm yeah i'm not very comfortable with the panic in here... Maybe we could return and error and leave it for the builder functions to panic if they need to ?

PR #28, opened 2021-06-28
Delete `--bind-port` flag

A-Hilaly:
Small rework of the `config` package that deleted `--bind-port`

NOTE: `Port` and `Host` options that are used to create a new
controller-runtime manager instance are used for the webhook server.
see https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L211-L216

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,7 +24,6 @@ import (
 )
 
 const (
-	flagBindPort             = "bind-port"
 	flagEnableLeaderElection = "enable-leader-election"
 	flagMetricAddr           = "metrics-addr"
 	flagEnableDevLogging     = "enable-development-logging"
@@ -40,7 +39,6 @@ const (
 
 // Config contains configuration otpions for ACK service controllers
 type Config struct {
-	BindPort                 int
 	MetricsAddr              string
 	EnableLeaderElection     bool
 	EnableDevelopmentLogging bool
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -35,12 +34,12 @@ const (
 	flagResourceTags         = "resource-tags"
 	flagWatchNamespace       = "watch-namespace"
 	flagEnableWebhookServer  = "enable-webhook-server"
-	flagWebhookServerAddr    = "webhook-server-addr"
+	flagWebhookServerPort    = "webhook-server-port"
+	flagWebhookServerHost    = "webhook-server-host"
```
A-Hilaly, line 38: I know this is not consistent with the all the other flag (`--metrics-addr` for example). Shall i use a single flag instead ? keeping the `--webhook-server-addr` and parse/split the host/port ?

vijtrip2: I recommend keeping the consistency with "metrics-addr".

IMO, separating port and host provides good user experience but ACK customers will not be dealing with these directly, correct? They will just install using helm charts. So I am fine with keeping it as "Addr".

A-Hilaly: Yep agreed :ok_hand:  I kept `webhook-server-addr` and just deleted `--bind-port` flag. PTAL

vijtrip2:
/approve. 
LGTM, I will let Jay take a look as well before merging.

A-Hilaly:
Just opened #29 which will help parsing `--webhook-server-addr`

PR #29, opened 2021-06-28
Add `GetHostPort` function to extract the host and port from network addresses

A-Hilaly:
This patch adds a new utility function called `GetHostPort` to help
extracting the host and port from network addresses.
This function will be used in the generated `main.go` to configure
the webhook server host and port. 

### DIFF
--- a/pkg/util/address.go
+++ b/pkg/util/address.go
@@ -1,0 +1,28 @@
+package util
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"strconv"
+)
+
+// GetHostPort extracts the host and port from a given address string.
+func GetHostPort(address string) (string, int, error) {
+	u, err := url.Parse(fmt.Sprintf("//%s", address))
+	if err != nil {
+		return "", 0, err
+	}
+
+	host, portString, err := net.SplitHostPort(u.Host)
+	if err != nil {
+		return "", 0, err
+	}
+
+	port, err := strconv.Atoi(portString)
+	if err != nil {
+		return "", 0, fmt.Errorf("cannot parse port: %v", err)
+	}
+
+	return host, port, nil
+}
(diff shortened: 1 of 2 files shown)


pkg/util/address.go
```diff
@@ -0,0 +1,27 @@
...
+	"strconv"
+	"strings"
+)
+
+// GetHostPort extracts the host and port from a given address string.
+func GetHostPort(address string) (string, int, error) {
+	parts := strings.Split(address, ":")
+	if len(parts) != 2 {
+		return "", 0, fmt.Errorf("malformed address")
+	}
```
RedbackThomson, line 14: THIS is exactly why I think the IPv6 format is doomed. How the hell do you parse `0::0:443` properly in IPv6? Is this IP `0000:0000:...:0000` port `443` or just IP `0000:0000:....:0000:0443`?

A-Hilaly: Good point, my first implementation didn't work with IPv6... so i was like let's start with IPv4 only support and improve it later

A-Hilaly: first implementation:
```go
// GetHostPort extracts the host and port from a given address string.
func GetHostPort(address string) (string, int, error) {
	u, err := url.Parse("//" + address)
	if err != nil {
		return "", 0, err
	}

	host, portString, err := net.SplitHostPort(u.Host)
	if err != nil {
		return "", 0, err
	}

	port, err := strconv.Atoi(portString)
	if err != nil {
		return "", 0, fmt.Errorf("cannot parse port: %v", err)
	}

	return host, port, nil
}


```

RedbackThomson: From the `net` documentation: `A literal IPv6 address in hostport must be enclosed in square brackets, as in "[::1]:80", "[::1%lo0]:80".`

I think your first implementation should still work?

A-Hilaly: Haaaaa! This should work i believe, thanks! :)

RedbackThomson:
Solid
/lgtm

PR #30, opened 2021-07-14
fix for #866 - nil Interface check during delete

kumargauravsharma:
Issue:
https://github.com/aws-controllers-k8s/community/issues/866

Fix:
`nil` Interface check during delete

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -16,7 +16,7 @@ package runtime
 import (
 	"context"
 	"time"
-	
+
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/runtime/reconciler.go
```diff
@@ -361,7 +362,7 @@ func (r *resourceReconciler) cleanup(
 	rlog.Enter("rm.Delete")
 	latest, err := rm.Delete(ctx, observed)
 	rlog.Exit("rm.Delete", err)
-	if latest != nil {
+	if latest != nil && !reflect.ValueOf(latest).IsNil() {
```
vijtrip2, line 365: I would recommend reusing this IsNil() method.
https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/compare/nil.go#L35

kumargauravsharma: Thanks @vijtrip2 
I have updated it to use ```IsNotNil()``` from: https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/compare/nil.go#L47

vijtrip2 (approved):

A-Hilaly (approved):

PR #31, opened 2021-07-15
Refactor reconciler.patchResource() to handle both Spec and Status patching

vijtrip2:
* Refactored `reconciler.patchResource` to patch both spec and status
* Added two separate methods patchResourceMetadataAndSpec , patchResourceStatus for granular patching
* Added new unit tests

This is a precursor change for server-side-defaults implementation. https://github.com/aws-controllers-k8s/community/issues/812

```
vijat@u872f1b87e2dc5e:~/go/src/github.com/aws-controllers-k8s/runtime$ make test
building mocks for pkg/types ... ok.
building mocks for k8s.io/apimachinery/pkg/apis/meta/v1 ... ok.
building mocks for k8s.io/apimachinery/runtime ... ok.
building mocks for k8s.io/apimachinery/runtime/schema ... ok.
building mocks for sigs.k8s.io/controller-runtime/pkg/client ... ok.
go test  ./...
?       github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1       [no test files]
?       github.com/aws-controllers-k8s/runtime/mocks/apimachinery/pkg/apis/meta/v1      [no test files]
?       github.com/aws-controllers-k8s/runtime/mocks/apimachinery/pkg/runtime   [no test files]
?       github.com/aws-controllers-k8s/runtime/mocks/apimachinery/pkg/runtime/schema    [no test files]
?       github.com/aws-controllers-k8s/runtime/mocks/controller-runtime/pkg/client      [no test files]
?       github.com/aws-controllers-k8s/runtime/mocks/pkg/types  [no test files]
ok      github.com/aws-controllers-k8s/runtime/pkg/compare      (cached)
ok      github.com/aws-controllers-k8s/runtime/pkg/condition    (cached)
?       github.com/aws-controllers-k8s/runtime/pkg/config       [no test files]
?       github.com/aws-controllers-k8s/runtime/pkg/errors       [no test files]
?       github.com/aws-controllers-k8s/runtime/pkg/metrics      [no test files]
ok      github.com/aws-controllers-k8s/runtime/pkg/requeue      (cached)
ok      github.com/aws-controllers-k8s/runtime/pkg/runtime      0.020s
ok      github.com/aws-controllers-k8s/runtime/pkg/runtime/cache        (cached)
?       github.com/aws-controllers-k8s/runtime/pkg/runtime/log  [no test files]
?       github.com/aws-controllers-k8s/runtime/pkg/types        [no test files]
[...]

PR #32, opened 2021-07-20
Reduce the number of instantiated/used `ackrtcache.Caches` objects.

A-Hilaly:
Issue: N/A

Currently, each ACK reconciler uses its own `ackrtcache.Caches`
instance. Long term this might cause some issues since each Cache
instance require its own KubeClient/shared informer etc...

This patch reduces the number of instantiated and used
`ackrtcache.Caches` to one. All the reconcilers will now use a copy of
the same cache instance since it's thread-safe.

### DIFF
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -21,8 +21,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


RedbackThomson (approved):
Fair enough. Good call

jaypipes (approved):
This is nice, thanks Amine. @vijtrip2 I will let you `/lgtm` this since it will cause you to have to rebase https://github.com/aws-controllers-k8s/runtime/pull/31 ...

A-Hilaly:
@vijtrip2 i can also rebase after #31 merge :)

A-Hilaly:
@vijtrip2 @jaypipes branch is rebased

PR #35, opened 2021-07-22
Add inline comments for deep copy patching

RedbackThomson:
Adds clarifying comments about why `DeepCopyObject` is necessary when using `Patch`.


jaypipes (approved):
nice, and thank you @RedbackThomson!

PR #36, opened 2021-07-22
Allow endpointURL to be specified in namespace annotations

ryansteakley:
related issue: https://github.com/aws-controllers-k8s/community/issues/876

We want the user to have the same level of control over `endpoint_url `as the existing account and region parameters. The current implementation it does not allow the user to specify` endpoint_url` in namespace meta data annotations when they are attempting to use cross-account or cross region resource management.


A-Hilaly (approved):
Great stuff @ryansteakley ! :+1:

jaypipes (approved):
I dig it. Nice work @ryansteakley :)

PR #39, opened 2021-07-29
simplify patching spec and status in reconciler

jaypipes:
The way that we were handling patching spec/metadata and status in the
main reconciler loop was starting to get awkward and messy. We were
calling `patchResource`, `patchResourceStatus` or
`patchResourceMetadataAndSpec` in quite a few places, the main
`resourceReconciler.Sync()` code path was getting messy and hard to
read, and we were in danger of having more subtle bugs creep into the
reconciler logic because of the duplicative patch calls and non-obvious
nature of the calls.

This patch simplifies and cleans up the logic around the core
`resourceReconciler.Sync()` code paths in the following ways:

* Breaks out the create and update code paths into their own separate
  `createResource` and `updateResource` methods.
* Renames the `cleanup` method to `deleteResource` in order to match the
  create and update methods
* Removes ALL calls to `patchResourceStatus` in favor of always ensuring
  Status patching is done in the
  `resourceReconciler.HandleReconcileError` wrapper method
* Removes the `resourceReconciler.patchResource` method entirely (which
  patched both Metadata/Spec as well as Status) and ensures that the
  `createResource`, `updateResource` and `deleteResource` methods on
  `resourceReconciler` only ever call `patchResourceMetadataAndSpec`
  (since `patchResourceStatus` is now always called by the
  `HandleReconcileError` wrapper)

These changes should make the late initialize work easier to fit into
the core reconciler logic and provide a more explicit contract about
what is responsible for patching status (**only** the
`HandleReconcileError` wrapper) and what is responsible for patching
metadata and spec (the `createResource`, `updateResource` and
`deleteResource` methods).

### DIFF
--- a/pkg/runtime/reconciler_test.go
+++ b/pkg/runtime/reconciler_test.go
@@ -154,9 +154,6 @@ func TestReconcilerUpdate(t *testing.T) {
 
 	r, kc := reconcilerMocks(rmf)
 
-	statusWriter := &ctrlrtclientmock.StatusWriter{}
-	kc.On("Status").Return(statusWriter)
-	statusWriter.On("Patch", ctx, latestRTObj, client.MergeFrom(desiredRTObj)).Return(nil)
 	kc.On("Patch", ctx, latestRTObj, client.MergeFrom(desiredRTObj)).Return(nil)
 
 	// With the above mocks and below assertions, we check that if we got a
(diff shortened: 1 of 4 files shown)


pkg/runtime/reconciler_test.go
```diff
@@ -154,23 +154,22 @@ func TestReconcilerUpdate(t *testing.T) {
...
-	err := r.Sync(ctx, rm, desired)
+	_, err := r.Sync(ctx, rm, desired)
 	require.Nil(err)
 	rm.AssertCalled(t, "ReadOne", ctx, desired)
 	rd.AssertCalled(t, "Delta", desired, latest)
 	rm.AssertCalled(t, "Update", ctx, desired, latest, delta)
+	// No changes to metadata or spec so Patch on the object shouldn't be done
 	kc.AssertNotCalled(t, "Patch", ctx, latestRTObj, client.MergeFrom(desiredRTObj))
-	statusWriter.AssertCalled(t, "Patch", ctx, latestRTObj, client.MergeFrom(desiredRTObj))
+	// Only the HandleReconcilerError wrapper function ever calls patchResourceStatus
```
vijtrip2, line 171: nit: should the test be renamed to TestSync_Update() , which is correct representation of the test.

vijtrip2 (approved):
Excellent Refactoring 🙌

pkg/types/aws_resource_reconciler.go
```diff
@@ -37,11 +38,28 @@ type AWSResourceReconciler interface {
...
-	) error
+	) (AWSResource, error)
+	// HandleReconcileError will handle errors from reconcile handlers, which
+	// respects runtime errors.
+	//
+	// If the `latest` parameter is not nil, this function will ALWAYS patch the
+	// latest Status fields back to the Kubernetes API.
+	//
+	// NOTE(jaypipes): This is really only here for dependency injection
+	// purposes in unit testing in order to simplify test setups.
```
A-Hilaly, line 58: :+1:

A-Hilaly (approved):
Nice! LGTM

PR #42, opened 2021-08-05
Optionally run the `Namespace` and `ConfigMap` caches

A-Hilaly:
Issue N/A

Description of changes:

- Run the caches only when `--watch-namespace` is not empty.
- Remove the "watch namespace" filter from the `NamespaceCache` event handlers.

### DIFF
--- a/pkg/runtime/cache/account.go
+++ b/pkg/runtime/cache/account.go
@@ -34,25 +34,13 @@ const (
 // make the changes accordingly.
 type AccountCache struct {
 	sync.RWMutex
-
-	log logr.Logger
-
-	// ConfigMap informer
-	informer k8scache.SharedInformer
+	log      logr.Logger
 	roleARNs map[string]string
 }
 
-// NewAccountCache makes a new AccountCache from a client.Interface
-// and a logr.Logger
-func NewAccountCache(clientset kubernetes.Interface, log logr.Logger) *AccountCache {
-	sharedInformer := informersv1.NewConfigMapInformer(
-		clientset,
-		currentNamespace,
-		informerResyncPeriod,
-		k8scache.Indexers{},
-	)
+// NewAccountCache instanciate a new AccountCache.
+func NewAccountCache(log logr.Logger) *AccountCache {
 	return &AccountCache{
-		informer: sharedInformer,
 		log:      log.WithName("cache.account"),
 		roleARNs: make(map[string]string),
 	}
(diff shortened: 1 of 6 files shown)


pkg/runtime/cache/account.go
```diff
@@ -65,10 +53,15 @@ func resourceMatchACKRoleAccountsConfigMap(raw interface{}) bool {
 	return ok && object.ObjectMeta.Name == ACKRoleAccountMap
 }
 
-// Run adds the default event handler functions to the SharedInformer and
-// runs the informer to begin processing items.
-func (c *AccountCache) Run(stopCh <-chan struct{}) {
-	c.informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
+// Run instantiate a new SharedInformer for ConfigMaps and runs it to begin processing items.
+func (c *AccountCache) Run(clientSet kubernetes.Interface, stopCh <-chan struct{}) {
```
RedbackThomson, line 57: Could we rename this to `Watch`? or `Subscribe`? These verbs imply that it's a continuous task of listening for differences, rather than just doing the action once.

RedbackThomson (commented):
This makes a lot of sense. Could we test this with a `Role` (instead of `ClusterRole`) and double check it solves our problem?

pkg/runtime/cache/account.go
ryansteakley: I'm working on testing these changes to see if it solves the problem

ryansteakley: Still see the same errors `E0806 04:30:23.241619       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:ack-system:ack-sagemaker-controller" cannot list resource "namespaces" in API group "" at the cluster scope`, only way I got them to not show up is by not running the cache at all but I don't think that is the intent of these changes.

pkg/runtime/service_controller.go
```diff
@@ -133,13 +133,15 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 	c.metaLock.Lock()
 	defer c.metaLock.Unlock()
 
-	clusterConfig := mgr.GetConfig()
-	clientset, err := kubernetes.NewForConfig(clusterConfig)
-	if err != nil {
-		return err
+	cache := ackrtcache.New(c.log)
+	if cfg.WatchNamespace != "" {
```
A-Hilaly, line 137:
```suggestion
	if cfg.WatchNamespace == "" {
```
Aaaaaah... :facepalm:

A-Hilaly: This is why you still see the error @ryansteakley

A-Hilaly: Fixed. Can you try again @ryansteakley?

pkg/runtime/cache/account.go
A-Hilaly: > Could we test this with a Role (instead of ClusterRole) and double check it solves our problem?

@RedbackThomson Ryan is working on a PoC using `Role` instead of `ClusterRole`

> Could we rename this to Watch? or Subscribe? These verbs imply that it's a continuous task of listening for differences, rather than just doing the action once.

I like more the `Watch` verb. How about `WatchNamespaces`/`WatchConfigMaps` ?

pkg/runtime/service_controller.go
ryansteakley: testing it out now

pkg/runtime/cache/account.go
RedbackThomson: Yes `WatchNamespaces` and `WatchConfigMaps` sounds good.

pkg/runtime/service_controller.go
ryansteakley: Worked for getting rid of the errors related to v1.Namespace, but now I'm seeing that although the resources are being created on sagemaker and I can also successfully delete them, none of this is reflected in kubectl get or kubectl describe as they don't seem to be updating.

jaypipes: @RedbackThomson please see above... is this anything to do with the GenerationPredicate bug you told me about?

RedbackThomson: I believe the `GenerationPredicate` bug has made this visible. On first reconciliation we were getting an error, it was updating the spec but not the status. On the second reconciliation, it updated the status.

PR #43, opened 2021-08-09
Call `ReadOne` as part of `createResource`

RedbackThomson:
Some status fields were not being set properly during the first reconciliation loop, as part of the Create flow. This loop called the `Create*` operation, and set any status fields that are returned as part of the output shape, and then returned successfully (without re-queuing). Previously, the reconciler would loop again to describe and set the other status fields, however we now filter out any events that do not change the resource generation. 

This fix will call `ReadOne` to describe the newly created resource and set the status fields based on that response, directly after calling `Create*`. This ensures all fields are set before completing the first reconciliation loop.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -265,11 +265,15 @@ func (r *resourceReconciler) createResource(
 	}
 
 	rlog.Enter("rm.ReadOne")
-	latest, err = rm.ReadOne(ctx, latest)
+	observed, err := rm.ReadOne(ctx, latest)
 	rlog.Exit("rm.ReadOne", err)
 	if err != nil {
 		return latest, err
 	}
+
+	// Take the status from the latest ReadOne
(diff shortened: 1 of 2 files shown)


RedbackThomson:
Tested with @A-Hilaly for DynamoDB. We see an initial reconciliation loop that sets the status as "Creating" before it requeues (based on hook logic) to eventually set it to status "Active".

A-Hilaly (approved):

RedbackThomson:
Tested with @ryansteakley that the SageMaker ModelPackageGroup has returned to normal, as well. He only sees a single reconciliation loop, and all status fields are filled.

vijtrip2 (approved):

RedbackThomson:
I have updated the PR so that we call `resource.SetStatus` so that we only update the status fields returned from the `ReadOne`. Spec fields returned from `ReadOne` should be handled by the late initializer logic, and will be introduced separately.

PR #44, opened 2021-08-12
Use endpointResolver and add serviceEndpointsID to serviceController

ryansteakley:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/896

Description of changes:
If `endpointURL` is not an empty string then we will use the `endpointresolver` and we will use this `endpointURL` when the service is equal to the `serviceEndpointID` which is retrieved from the service controller. 

### DIFF
--- a/pkg/runtime/session.go
+++ b/pkg/runtime/session.go
@@ -42,8 +42,17 @@ func (c *serviceController) NewSession(
 		Region:              aws.String(string(region)),
 		STSRegionalEndpoint: endpoints.RegionalSTSEndpoint,
 	}
-	if endpointURL != nil {
-		awsCfg.Endpoint = endpointURL
+
+	if *endpointURL != "" {
+		endpointServiceResolver := func(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
+			if service == c.ServiceEndpointsID {
+				return endpoints.ResolvedEndpoint{
+					URL: *endpointURL,
+				}, nil
+			}
+			return endpoints.DefaultResolver().EndpointFor(service, region)
+		}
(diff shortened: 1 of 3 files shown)


pkg/runtime/session.go
```diff
@@ -42,8 +42,17 @@ func (c *serviceController) NewSession(
...
 		STSRegionalEndpoint: endpoints.RegionalSTSEndpoint,
 	}
-	if endpointURL != nil {
-		awsCfg.Endpoint = endpointURL
+
+	if *endpointURL != "" {
+		endpointServiceResolver := func(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
+			if service == c.ServiceEndpointsID {
+				return endpoints.ResolvedEndpoint{
+					URL: *endpointURL,
```
RedbackThomson, line 50: There is also a "SigningRegion" field. Do you know if that is applicable, or what that's used for?

ryansteakley: There are various "signing" fields. Basing this code off of the old operators for sagemaker it was not specified. https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#ResolvedEndpoint. I believe it has to do with https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html, but I don't think it is really necessary.

RedbackThomson: Yeah, Googling the subject seems to show few results. I imagine it defaults to a sane value.

A-Hilaly (approved):

jaypipes (approved):
This is exactly what I had in mind in the original PR #7. Nice work @ryansteakley.

PR #45, opened 2021-08-17
Set Terminal condition on updating unmanaged

RedbackThomson:
ACK uses finalizers to determine whether it is currently managing a k8s resource. We set the finalizer when creating the resource and delete it when we have deleted the underlying AWS resource. The current code path also sets the finalizer if one is not applied when attempting to update a resource. If a user were to create a resource that has the same name as an existing AWS resource, it will attempt to call `updateResource` which would set the finalizer and treat the resource as if ACK created it. This behaviour circumvents the safer adoption mechanisms put in place to prevent overriding existing resources.

This pull request ensures that any resource which does not have a finalizer, when attempting to call `updateResource`, will be marked with a terminal condition. Adopted resources are now marked with the finalizer when they are created.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -562,9 +562,7 @@ func (r *resourceReconciler) failOnResourceUnmanaged(
 		return nil
 	}
 
-	notManagedMessage := "Resource already exists"
-	notManagedReason := "This resource already exists and is not managed by ACK"
-	condition.SetTerminal(res, corev1.ConditionTrue, &notManagedMessage, &notManagedReason)
+	condition.SetTerminal(res, corev1.ConditionTrue, &condition.NotManagedMessage, &condition.NotManagedReason)
 	return ackerr.Terminal
 }
 
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -28,6 +28,7 @@ import (
 
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	"github.com/aws-controllers-k8s/runtime/pkg/condition"
 	ackcfg "github.com/aws-controllers-k8s/runtime/pkg/config"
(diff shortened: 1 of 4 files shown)


vijtrip2:
looks good except Unit test failures.

vijtrip2:
/approve

I would let Jay review as well so not merging.

pkg/runtime/reconciler.go
```diff
@@ -557,20 +543,31 @@ func (r *resourceReconciler) setResourceUnmanaged(
...
+func (r *resourceReconciler) failOnResourceUnmanaged(
+	ctx context.Context,
+	res acktypes.AWSResource,
+) error {
+	if r.rd.IsManaged(res) {
+		return nil
+	}
+
+	notManagedMessage := "Resource already exists"
+	notManagedReason := "This resource already exists and is not managed by ACK"
```
jaypipes, line 566: Let's provide a little more help to the user. How about this instead?

```go
msg := "This resource already exists but is not managed by ACK. "+
    "To bring the resource under ACK management, you should explicitly adopt "+
    "the resource by creating a services.k8s.aws/AdoptedResource"
```

And since that's pretty long, why not place that text as a constant in the `pkg/conditions` package and use it here.

jaypipes (changes requested):
Looking great. One small suggestion inline, though

pkg/runtime/reconciler_test.go
```diff
@@ -393,3 +407,62 @@ func TestReconcilerUpdate_ErrorInLateInitialization(t *testing.T) {
...
+
+	_, err := r.Sync(ctx, rm, desired)
+	// Assert the error from late initialization
+	require.NotNil(err)
+	assert.Equal(ackerr.Terminal, err)
+	rm.AssertCalled(t, "ReadOne", ctx, desired)
+	rd.AssertNotCalled(t, "Delta", desired, latest)
+	rm.AssertNotCalled(t, "Update", ctx, desired, latest, delta)
+	rm.AssertNotCalled(t, "LateInitialize", ctx, latest)
+}
```
jaypipes, line 468: nice test. ++

jaypipes (approved):
Well done.

PR #46, opened 2021-08-19
Adding abstractions for exponential backoff & retry during late initialization

vijtrip2:
Issue # https://github.com/aws-controllers-k8s/community/issues/812

* Adds new pkg/annotation to handle lateInitializationAttempt annotations
* Adds new Exponential struct to handle exponential backoff and retry calculations.

### DIFF
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -14,6 +14,7 @@
 package annotation
 
 import (
+	"errors"
 	"strconv"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -21,7 +22,7 @@ import (
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 )
 
-var (
+const (
 	LateInitializationAttempt = "services.k8s.aws/late-initialization-attempt"
 )
 
@@ -46,26 +47,25 @@ func GetNumLateInitializationAttempt(metav1Obj metav1.Object) int {
 // attempts from k8s resource annotation, increments by 1, and stores back into
 // k8s resource annotation
 // This method updates the object the passed in the parameter inplace. Make sure
-//to pass a pointer object for updates to reflect after method call.
+// to pass a pointer object for updates to reflect after method call.
 // NOTE: AWSResource.MetaObject() method for all ACK resources returns a pointer
 // , so this method is safe to use without return type.
 // Also See https://github.com/aws-controllers-k8s/community/issues/905
-func IncrementNumLateInitializationAttempt(metav1Obj metav1.Object) {
+func IncrementNumLateInitializationAttempt(metav1Obj metav1.Object) error {
 	if ackcompare.IsNil(metav1Obj) {
-		return
+		return errors.New("metav1Obj parameter should not be nil")
 	}
 	annotations := metav1Obj.GetAnnotations()
 	if annotations != nil {
 		if currentNumAttemptStr, ok := annotations[LateInitializationAttempt]; ok {
 			if currentNumAttempt, err := strconv.Atoi(currentNumAttemptStr); err == nil {
-				SetNumLateInitializationAttempt(metav1Obj, currentNumAttempt+1)
-				return
+				return SetNumLateInitializationAttempt(metav1Obj, currentNumAttempt+1)
 			}
 		}
 	}
 	// If late initialization attempt annotation is not present or value is not
 	// a number, reset to 1
-	SetNumLateInitializationAttempt(metav1Obj, 1)
+	return SetNumLateInitializationAttempt(metav1Obj, 1)
 }
 
 // SetNumLateInitializationAttempt stores 'numAttempt' into k8s resource annotation
@@ -75,9 +75,9 @@ func IncrementNumLateInitializationAttempt(metav1Obj metav1.Object) {
 // NOTE: AWSResource.MetaObject() method for all ACK resources returns a pointer
 // , so this method is safe to use without return type.
 // Also See: https://github.com/aws-controllers-k8s/community/issues/905
-func SetNumLateInitializationAttempt(metav1Obj metav1.Object, numAttempt int) {
+func SetNumLateInitializationAttempt(metav1Obj metav1.Object, numAttempt int) error {
 	if ackcompare.IsNil(metav1Obj) {
-		return
+		return errors.New("metav1Obj parameter should not be nil")
 	}
 	annotations := make(map[string]string)
 	if metav1Obj.GetAnnotations() != nil {
@@ -85,6 +85,7 @@ func SetNumLateInitializationAttempt(metav1Obj metav1.Object, numAttempt int) {
 	}
 	annotations[LateInitializationAttempt] = strconv.Itoa(numAttempt)
 	metav1Obj.SetAnnotations(annotations)
+	return nil
 }
 
 // RemoveLateInitializationAttempt removes the 'services.k8s.aws/late-initialization-attempt'
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -1,0 +1,103 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package annotation
+
+import (
+	"strconv"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+)
+
+var (
+	LateInitializationAttempt = "services.k8s.aws/late-initialization-attempt"
+)
+
+// GetNumLateInitializationAttempt returns the number of late initialization
+// attempts made. This value is stored in the k8s resource annotation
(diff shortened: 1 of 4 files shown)


pkg/annotation/annotation.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"strconv"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+)
+
+var (
```
A-Hilaly, line 24:
```suggestion
const (
```

pkg/annotation/annotation_test.go
```diff
@@ -0,0 +1,89 @@
+package annotation_test
```
A-Hilaly, line 1: missing license header

pkg/types/retry_test.go
```diff
@@ -0,0 +1,128 @@
+package types_test
+
+import (
+	"testing"
+	"time"
+
+	types "github.com/aws-controllers-k8s/runtime/pkg/types"
```
A-Hilaly, line 7: This package mostly imported as `acktypes` in other runtime/controller packages

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// SetNumLateInitializationAttempt stores 'numAttempt' into k8s resource annotation
+// 'services.k8s.aws/late-initialization-attempt'
+// This method updates the object the passed in the parameter inplace. Make sure
+// to pass a pointer object for updates to reflect after method call.
+// NOTE: AWSResource.MetaObject() method for all ACK resources returns a pointer
+// , so this method is safe to use without return type.
+// Also See: https://github.com/aws-controllers-k8s/community/issues/905
+func SetNumLateInitializationAttempt(metav1Obj metav1.Object, numAttempt int) {
```
A-Hilaly, line 78: Is there a special reason you're using annotations and not conditions for this case?

vijtrip2: Conditions are also used in addition with Annotations for late initialization.
I have used `conditions` to provide status, reason, message etc about late initialization.

However to provide the metadata about late initialization attempts, `Annotations` felt like the correct place to do it with key value schema. It also hides away extra details from user (who check resource status mainly)

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,103 @@
...
+	if ackcompare.IsNil(metav1Obj) {
+		return
```
RedbackThomson, line 80: Also would return `error` here. Leave it up to the consumer to determine whether they want to ignore it.

pkg/types/retry.go
```diff
@@ -0,0 +1,53 @@
...
+
+package types
+
+import (
+	"math"
+	"time"
+)
+
+// Exponential type helps in implementing exponential backoff strategy.
+// NOTE: The delay added is in seconds.
```
RedbackThomson, line 22: Is this a relevant note? `time.Duration` handles units appropriately

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// RemoveLateInitializationAttempt removes the 'services.k8s.aws/late-initialization-attempt'
+// annotation from k8s resource annotations
+// This method updates the object the passed in the parameter inplace. Make sure
+// to pass a pointer object for updates to reflect after method call.
+// NOTE: AWSResource.MetaObject() method for all ACK resources returns a pointer
+// , so this method is safe to use without return type.
+// Also See https://github.com/aws-controllers-k8s/community/issues/905
+func RemoveLateInitializationAttempt(metav1Obj metav1.Object) {
```
jaypipes, line 98: nit: rename to `RemoveLateInitializeAttempts`

pkg/types/retry.go
```diff
@@ -0,0 +1,52 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package types
+
+import (
+	"math"
+	"time"
+)
```
jaypipes, line 19: I've used the https://github.com/cenkalti/backoff library before in previous projects and it's pretty small and simple and might reduce the amount of code we would need to keep in ACK itself.

@vijtrip2 would you mind taking a look at whether the [`github.com/cankalti/backoff.ExponentialBackoff:NextBackOff`](https://github.com/cenkalti/backoff/blob/v4/exponential.go#L116) function might be used instead of custom code?

The https://github.com/cenkalti/backoff/ library is MIT-licensed and so not concerning from an attribution perspective.

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,104 @@
...
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package annotation
```
jaypipes, line 14: after reading this PR a number of times, I think I'd prefer a `pkg/lateinitialize` package with a simplified calling interface kind of like this:

(this would be in, say, `pkg/runtime/reconciler.go` [here](https://github.com/aws-controllers-k8s/runtime/blob/a506412923e6667fbc36cc4c841f4c6efe8bd51e/pkg/runtime/reconciler.go#L210))

```go

import (
    acklateinit "github.com/aws-controllers-k8s/runtime/pkg/lateinitialize"
)

// Sync ensures that the supplied AWSResource's backing API resource
// matches the supplied desired state.
//
// It returns a copy of the resource that represents the latest observed state.
func (r *resourceReconciler) Sync(
	ctx context.Context,
	rm acktypes.AWSResourceManager,
	desired acktypes.AWSResource,
	var err error
	exit := rlog.Trace("r.Sync")
	defer exit(err)

	var latest acktypes.AWSResource // the newly created or mutated resource

	isAdopted := IsAdopted(desired)
	rlog.WithValues("is_adopted", isAdopted)

	rlog.Enter("rm.ReadOne")
	latest, err = rm.ReadOne(ctx, desired)
	rlog.Exit("rm.ReadOne", err)

	// Check to see if we attempted to late initialize this resource previously and
	// marked the resource as needing to be late initialized after some delay
	if acklateinit.RequiresInitialization(desired) {
		if latest, err = r.lateInitializeResource(ctx, rm, latest); err != nil {
			return latest, err
		}
	}
	if err != nil {
		if err != ackerr.NotFound {
			return latest, err
		}
		if isAdopted {
			return nil, ackerr.AdoptedResourceNotFound
		}
		if latest, err = r.createResource(ctx, rm, desired); err != nil {
			return latest, err
		}
	} else {
		if latest, err = r.updateResource(ctx, rm, desired, latest); err != nil {
			return latest, err
		}
	}
	// Attempt to late initialize the resource. If there are no fields to
	// late initialize, this operation will be a no-op.
	if latest, err = r.lateInitializeResource(ctx, rm, latest); err != nil {
		return latest, err
	}
	return r.handleRequeues(ctx, latest)
}
```

jaypipes (changes requested):
@vijtrip2 let's chat about this one when you're around today. I've got some suggestions to change the calling interface here as well as use a well-known library for calculating backoffs.

pkg/types/retry.go
vijtrip2: I will definitely take a look at this library and see if I can make use of it here instead of the custom code(which to be honest is very less.)

pkg/annotation/annotation.go
vijtrip2: Great Suggestion! I agree to implement it but I have two suggestions for the above code.

a) Perform re-attempt of late initialization before the readOne call inside Sync() method. i.e. Make it the very first action in Sync method. Since lateInitialization makes the readOne call, we do not need to make the extra call.

b) We save the lateInitialized desired into the same object and if more requeues are not required, we continue with the rest of Sync function as it was before.

```go
import (
    acklateinit "github.com/aws-controllers-k8s/runtime/pkg/lateinitialize"
)

// Sync ensures that the supplied AWSResource's backing API resource
// matches the supplied desired state.
//
// It returns a copy of the resource that represents the latest observed state.
func (r *resourceReconciler) Sync(
	ctx context.Context,
	rm acktypes.AWSResourceManager,
	desired acktypes.AWSResource,
	var err error
	exit := rlog.Trace("r.Sync")
	defer exit(err)

	var latest acktypes.AWSResource // the newly created or mutated resource

	isAdopted := IsAdopted(desired)
	rlog.WithValues("is_adopted", isAdopted)

        // Check to see if we attempted to late initialize this resource previously and
	// marked the resource as needing to be late initialized after some delay
	if acklateinit.RequiresInitialization(desired) {
                // This method only updates the late initialized spec fields if they are
                // not provided by the user but added by the AWS service. Which makes
                // it safe to capture the result as desired state of k8s resource.
		if desired, err = r.lateInitializeResource(ctx, rm, desired); err != nil {
			return desired, err
		}
	}

	rlog.Enter("rm.ReadOne")
	latest, err = rm.ReadOne(ctx, desired)
	rlog.Exit("rm.ReadOne", err)

	if err != nil {
		if err != ackerr.NotFound {
			return latest, err
		}
		if isAdopted {
			return nil, ackerr.AdoptedResourceNotFound
		}
		if latest, err = r.createResource(ctx, rm, desired); err != nil {
			return latest, err
		}
	} else {
		if latest, err = r.updateResource(ctx, rm, desired, latest); err != nil {
			return latest, err
		}
	}
	// Attempt to late initialize the resource. If there are no fields to
	// late initialize, this operation will be a no-op.
	if latest, err = r.lateInitializeResource(ctx, rm, latest); err != nil {
		return latest, err
	}
	return r.handleRequeues(ctx, latest)
}
```

pkg/types/retry.go
jaypipes: @vijtrip2 actually, look what I found :) https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L380

PR #47, opened 2021-08-19
Add `Status` when adopting resource

RedbackThomson:
`r.kc.Create()` does not add the `Status` fields from `ReadOne` to the newly created resource. Some resources (like SageMaker's `ModelPackage`) require these resource fields as their identifier for the `ReadOne` call. It is necessary to patch the `Status` fields so the consequent update loop is able to find the resource.


vijtrip2 (approved):
No unit test updates needed?

jaypipes (approved):
:+1: sensible change. Agree with @vijtrip2 that we should have some unit tests for this functionality (the adopted resource reconciler in general...) but we can add that in a followup PR.

PR #49, opened 2021-09-02
Use APIReader for reading desired object directly from k8s apiserver

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/894

Description of changes:
* Use `APIReader` for making reading desired state of k8s resource directly from apiserver and avoid any client caches.
Also see: https://github.com/kubernetes-sigs/controller-runtime/issues/585#issuecomment-528102351

Tested locally with apigatewayv2 controller and this change resolves the duplicate API creation issue.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -40,12 +40,13 @@ import (
 
 // reconciler describes a generic reconciler within ACK.
 type reconciler struct {
-	sc      acktypes.ServiceController
-	kc      client.Client
-	log     logr.Logger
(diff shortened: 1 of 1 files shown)


RedbackThomson (approved):
Seems simple enough. Interested to see what effect this has when there are many custom resources being reconciled at once.

vijtrip2:
> Seems simple enough. Interested to see what effect this has when there are many custom resources being reconciled at once.

By default our controller only reconciles one object per GVK at any given time, so I do not think it will overload the API Server. But yes we'll see.

PR #50, opened 2021-09-07
do not patch status if resource is not managed

surajkota:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/931

Description of changes:
Solution 1 in the issue linked above - `return nil, nil` instead of `resource, nil` after finalizer is removed 

Testing:
`make test`
Manually using sagemaker controller

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -474,11 +474,14 @@ func (r *resourceReconciler) deleteResource(
 	if err != nil {
 		if err == ackerr.NotFound {
 			// If the aws resource is not found, remove finalizer
-			// and finish reconciliation
 			if err = r.setResourceUnmanaged(ctx, current); err != nil {
 				return current, err
 			}
 			rlog.Info("deleted resource")
+
+			// Return nil resource so that patchStatus attempt is not made.
+			// Underlying K8s resource gets deleted after removing finalizer string and
+			// patchStatus call will return NotFound error.
 			return nil, nil
 		}
 		return current, err
@@ -511,8 +514,9 @@ func (r *resourceReconciler) deleteResource(
 	}
 	rlog.Info("deleted resource")
 
-	// Since the finalizer has been removed, the controller should successfully
-	// reconcile
+	// Return nil resource so that patchStatus attempt is not made.
+	// Underlying K8s resource gets deleted after removing finalizer string and
+	// patchStatus call will return NotFound error.
 	return nil, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/runtime/reconciler.go
```diff
@@ -505,7 +510,10 @@ func (r *resourceReconciler) deleteResource(
 		return latest, err
 	}
 	rlog.Info("deleted resource")
-	return latest, nil
+
+	// Since the finalizer has been removed, the controller should successfully
```
vijtrip2, line 514: Please update this comment to say exactly what is happening here.
"Return 'nil' resource so that patchStatus attempt is not made. Underlying K8s resource gets deleted after removing finalizer string and patchStatus call will return NotFound error."

vijtrip2 (approved):

pkg/runtime/reconciler.go
surajkota: done!

vijtrip2 (approved):

vijtrip2:
Waiting for other ACK core team members to take a look as well. 
I will merge this PR tomorrow morning, if it does not receive new comments.

pkg/runtime/reconciler.go
```diff
@@ -505,7 +513,11 @@ func (r *resourceReconciler) deleteResource(
 		return latest, err
 	}
 	rlog.Info("deleted resource")
-	return latest, nil
+
+	// Return nil resource so that patchStatus attempt is not made.
+	// Underlying K8s resource gets deleted after removing finalizer string and
+	// patchStatus call will return NotFound error.
+	return nil, nil
```
jaypipes, line 520: Instead of all of the above changes, would it not be easier to just change [this one line](https://github.com/aws-controllers-k8s/runtime/blob/24e0b8ddeaaf7df4bf55ef1a75e4608ea5ec01fb/pkg/runtime/reconciler.go#L647)

from this:

```go
	if ackcompare.IsNotNil(latest) {
```

to this:

```go
	if ackcompare.IsNotNil(latest) && r.rd.IsManaged(latest) {
```

surajkota: that was my 3rd solution on the issue. What do you think about doing this in addition to the current change?
so logically the return from deleteResource also matches the condition change you suggested

jaypipes: @surajkota the return from the Delete operation should be the latest observed state of the resource. Not nil.

surajkota: oh right, done!

jaypipes (approved):
:+1:

PR #51, opened 2021-09-08
Use APIReader for reading adopted resource. Stop using deepcopy for kc.Patch calls

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/935

Description of changes:
* Mostly adding the latest improvements from reconciler into adoption_reconciler
    * Use APIReader to read k8s resource instead of kc.Client for avoiding caching problems
    * Do not use DeepCopy for 'kc.Patch' calls because the object metadata gets lost.
    * Preserve resource status after kc.Patch call.
    * Only create resource if it does not already exists. Making adoption reconciler idempotent.

TODOs (https://github.com/aws-controllers-k8s/community/issues/939)
* Handle 'resource modified underneath' error when resource gets modified by resource reconciler while adopted reconciler is still patching status.
* Add unit tests for the adoption_reconciler.go file
* Refactor 'PatchMetadataAndSpec' , 'PatchStatus' method to be reusable between 'reconciler.go' & 'adoption_reconciler.go'

Testing:
* Sagemaker adoption_reconciler e2e tests passed

### DIFF
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -22,6 +22,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	ctrlrt "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	k8sctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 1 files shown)


surajkota:
Can you help me understand how this PR is solving the issue in 935?

> Do not use DeepCopy for 'kc.Patch' calls because the object metadata gets lost
if this is the reason, wouldn't the Status.patch fail everytime?

vijtrip2:
> Can you help me understand how this PR is solving the issue in 935?
> 
> 
> 
> > Do not use DeepCopy for 'kc.Patch' calls because the object metadata gets lost
> 
> 
> 
> if this is the reason, wouldn't the Status.patch fail everytime?
> 
> 

This PR should help fix the initial problem that you saw where the error was that 'resource has been modified'. 

We solved this problem in reconciler.go and I am porting over the gold from reconciler.go into adoption_reconciler.go .

pkg/runtime/adoption_reconciler.go
```diff
@@ -481,6 +486,45 @@ func (r *adoptionReconciler) getRegion(
...
+// for reconciler and adoptionReconciler
+func (r *adoptionReconciler) patchMetadataAndSpec(
+	ctx context.Context,
+	res *ackv1alpha1.AdoptedResource,
+	base *ackv1alpha1.AdoptedResource,
+) error {
+	// k8s Client Patch call updates the status of original object with the
+	// content returned from apiserver.
+	// Keep a copy of status field to reset the status of 'res' after patch call
+	resStatusCopy := res.DeepCopy().Status
```
A-Hilaly, line 502: +++

pkg/runtime/adoption_reconciler.go
```diff
@@ -344,39 +361,27 @@ func (r *adoptionReconciler) getTargetResourceGroupKind(
 }
 
 // markManaged places the supplied resource under the management of ACK.
+// It adds the finalizer string, patches the object in etcd and updates
+// the object 'res' in parameter with latest metadata.
 func (r *adoptionReconciler) markManaged(
 	ctx context.Context,
-	res ackv1alpha1.AdoptedResource,
+	res *ackv1alpha1.AdoptedResource,
```
A-Hilaly, line 368: 👍

A-Hilaly (approved):
Very good stuff @vijtrip2 ! 👍

jaypipes (approved):
Thanks for bringing the adopted reconciler up to speed, @vijtrip2 :)

PR #52, opened 2021-09-08
remove duplicate log fields

jaypipes:
The use of `logr.Logger.WithValues` in the
`pkg/runtime/log.AdoptResource` function was not actually overwriting
previously set key/value pairs -- for instance, for the resource's kind,
generation or namespace. Instead, these structured log record fields
were simply being appended to the end of the `[]interface{}` that
`logr.Logger` uses to track log record fields.

See
https://pkg.go.dev/github.com/go-logr/logr#readme-why-key-value-pairs-and-not-a-map
for information on why a `[]interface{}` is used instead of a
`map[string]interface{}`, which would have avoided this bug/problem.

This patch removes the use of `logr.Logger.WithValues` in the
`ResourceLogger.Debug`, `ResourceLogger.Enter|Exit` and
ResourceLogger.Info` methods and replaces with direct passing of the
core resource fields into the `logr.Logger.Info` method.

Fixes Issue aws-controllers-k8s/runtime#938

### DIFF
--- a/pkg/runtime/log/resource.go
+++ b/pkg/runtime/log/resource.go
@@ -44,7 +44,8 @@ func (rl *ResourceLogger) Debug(
 	msg string,
 	additionalValues ...interface{},
 ) {
-	AdaptResource(rl.log, rl.res, additionalValues...).V(1).Info(msg)
+	vals := expandResourceFields(rl.res, additionalValues...)
+	rl.log.V(1).Info(msg, vals...)
 }
 
 // Info writes a supplied log message about a resource that includes a
@@ -53,7 +54,8 @@ func (rl *ResourceLogger) Info(
 	msg string,
 	additionalValues ...interface{},
 ) {
-	AdaptResource(rl.log, rl.res, additionalValues...).V(0).Info(msg)
+	vals := expandResourceFields(rl.res, additionalValues...)
+	rl.log.V(0).Info(msg, vals...)
 }
 
 // Enter logs an entry to a function or code block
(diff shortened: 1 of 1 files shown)


vijtrip2:
>This patch removes the use of logr.Logger.WithValues and replaces with
direct passing of the core resource fields into the logr.Logger.Info
method.

What am i missing here? I still see "log.WithValues(vals...)".
I only see code being refactored into separate method. Apologies if I had an oversight.

jaypipes:
> > This patch removes the use of logr.Logger.WithValues and replaces with
> > direct passing of the core resource fields into the logr.Logger.Info
> > method.
> 
> What am i missing here? I still see "log.WithValues(vals...)".
> I only see code being refactored into separate method. Apologies if I had an oversight.

Ugh, sorry :( Serves me right for trying to rush in a PR before a PE review... will fix shortly.

jaypipes:
@vijtrip2 thanks for the catch. fixed.

pkg/runtime/log/resource.go
```diff
@@ -163,28 +150,38 @@ func InfoResource(
...
 	}
 	if len(additionalValues) > 0 {
 		vals = append(vals, additionalValues...)
 	}
+	return vals
+}
+
+// AdaptAdoptedResource returns a logger with log values set for the adopted
+// resource's kind, namespace, name, etc
+func AdaptAdoptedResource(
```
vijtrip2, line 179: femto nit: Maybe these logging methods can be removed now since we log directly now.
Can be done in future PRs.

vijtrip2 (approved):

PR #53, opened 2021-09-09
Add unit tests for adoption_reconciler.go

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/939

Description of changes:
* Add an interface to test Sync method of adoption_reconciler similar to resource_reconciler
* Write unit tests for Sync method of AdoptionReconciler
* Add a new factory method to allow dependency injection of k8sClient and Reader.


pkg/runtime/adoption_reconciler_test.go
```diff
@@ -0,0 +1,505 @@
...
+func setupMockClient(kc *ctrlrtclientmock.Client, statusWriter *ctrlrtclientmock.StatusWriter, ctx context.Context, adoptedRes *ackv1alpha1.AdoptedResource) {
+	kc.On("Status").Return(statusWriter)
+	statusWriter.On("Patch", ctx, adoptedRes, mock.AnythingOfType("*client.mergeFromPatch")).Return(nil)
+	kc.On("Patch", ctx, adoptedRes, mock.AnythingOfType("*client.mergeFromPatch")).Return(nil)
+}
+
+func setupMockAwsResource(res *ackmocks.AWSResource, adoptedRes *ackv1alpha1.AdoptedResource) {
+	res.On("SetIdentifiers", adoptedRes.Spec.AWS).Return(nil)
+	res.On("RuntimeObject").Return(&k8srtmocks.Object{})
+	res.On("SetObjectMeta", mock.AnythingOfType("ObjectMeta")).Run(func(args mock.Arguments) {})
```
RedbackThomson, line 98: Empty mock method? Should this be actually setting the meta for it to mock correctly?

pkg/runtime/adoption_reconciler_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+	// Call
+	err := r.Sync(ctx, descriptor, manager, adoptedRes)
+
+	// Assertions
+	// error occured
+	require.NotNil(err)
+	require.Equal("unable to set Identifier", err.Error())
+	// Identifiers are set from AdoptedResource into AWSResource
+	res.AssertCalled(t, "SetIdentifiers", adoptedRes.Spec.AWS)
```
RedbackThomson, line 172: nit: Fix comment

pkg/runtime/adoption_reconciler_test.go
```diff
@@ -0,0 +1,505 @@
...
+	apiReader.AssertNotCalled(t, "Get", ctx, types.NamespacedName{
+		Namespace: Namespace,
+		Name:      Name,
+	}, res.RuntimeObject())
+	// Create AWSResource should not happen
+	kc.AssertNotCalled(t, "Create", ctx, res.RuntimeObject())
+	// Update status of AWSResource should not happen
+	statusWriter.AssertNotCalled(t, "Update", ctx, res.RuntimeObject())
+	// AdoptedResource did not get marked as managed
+	kc.AssertNotCalled(t, "Patch", ctx, adoptedRes, mock.AnythingOfType("*client.mergeFromPatch"))
```
RedbackThomson, line 185: Hmm as per your code comment above. This should probably be marked as managed by this point, because we have already attempt to manage it - even though it was unsuccessful.

pkg/types/adopted_resource_reconciler.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+)
+
+// AdoptedResourceReconciler is responsible for reconciling an adopted resource
+// that represent AWS service API resource.
+// It implements the upstream controller-runtime `Reconciler`
+// interface.
+type AdoptedResourceReconciler interface {
```
RedbackThomson, line 26: I think this interface should remain in the `adoption_reconciler.go` file.

RedbackThomson (changes requested):
These are well overdue on my behalf, so I really appreciate the time you've put into this. I will continue to expand them out as I make bug fixes to the reconciler in the future. As for feedback:

I think there is quite a bit of duplication in assertion logic between the tests. Would be good to split them into more properly named assertion methods, like `assertAdoptedCondition(bool)`, `assertIsManaged(bool)`, etc.

pkg/runtime/adoption_reconciler_test.go
vijtrip2: No, res is mock object. We cannot save anything there.
This empty method is needed to set expected behavior for this call otherwise an ugly warning is printed that mock does not know what to do on this method call.
There is also no return type on `SetObjectMeta`, so cannot simply return `nil`

pkg/runtime/adoption_reconciler_test.go
vijtrip2: Yeah, that's a bug in the actual code. I have added a TODO for it in the source.
In this PR, i will just add test for whatever functionality is there as is.

pkg/types/adopted_resource_reconciler.go
vijtrip2: I just placed it in same directory as AWSResourceReconciler, because I think these two belong together as top level type in same directory(`types`).
Can you provide reasoning for your suggestion? (Do you think AdoptedResourceReconciler is not a first class type in ACK runtime, unlike AWSResourceReconciler ?)

RedbackThomson: Ohhhh yeah I see what you mean. Great

RedbackThomson:
Great!
/approve
/lgtm

PR #54, opened 2021-09-28
use sts GetCallerIdentity to find AWS AccountID

vijtrip2:
Description of changes:

Breaking changes : Removing --aws-account-id flag from controller binary. This flag can allow users to enter wrong account_id during controller installation which can cause incorrect controller logs and confusion

* If `aws-region` flags are missing, default to 'AWS_REGION' env variable
* Add new method `Config.SetAccountID()` to use sts `GetCallerIdentity` for detecting aws account-id.

------------------

* TODO: code-gen update to remove `--aws-account-id` flag, community repo docs update and test-infra update to drop `aws-account-id` from `test-helm.sh` & `kind-build-test.sh`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.37.10
 	github.com/go-logr/logr v0.1.0
 	github.com/google/go-cmp v0.3.1
+	github.com/jaypipes/envutil v1.0.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.1.0
 	github.com/spf13/pflag v1.0.5
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -15,6 +15,7 @@ package config
 
 import (
 	"errors"
+	"fmt"
 	"net/url"
 
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -30,7 +31,6 @@ const (
 	flagEnableLeaderElection = "enable-leader-election"
 	flagMetricAddr           = "metrics-addr"
 	flagEnableDevLogging     = "enable-development-logging"
-	flagAWSAccountID         = "aws-account-id"
 	flagAWSRegion            = "aws-region"
 	flagAWSEndpointURL       = "aws-endpoint-url"
 	flagLogLevel             = "log-level"
@@ -38,6 +38,7 @@ const (
 	flagWatchNamespace       = "watch-namespace"
 	flagEnableWebhookServer  = "enable-webhook-server"
 	flagWebhookServerAddr    = "webhook-server-addr"
+	envVarAWSRegion          = "AWS_REGION"
 )
 
 // Config contains configuration otpions for ACK service controllers
@@ -85,13 +86,8 @@ func (cfg *Config) BindFlags() {
 			"otherwise a Zap production config will be used (encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error), sampling).",
 	)
 	flag.StringVar(
-		&cfg.AccountID, flagAWSAccountID,
-		envutil.WithDefault("AWS_ACCOUNT_ID", ""),
-		"The AWS Account ID in which the service controller will create resources",
-	)
-	flag.StringVar(
 		&cfg.Region, flagAWSRegion,
-		envutil.WithDefault("AWS_REGION", ""),
+		envutil.WithDefault(envVarAWSRegion, ""),
 		"The AWS Region in which the service controller will create its resources",
 	)
 	flag.StringVar(
@@ -137,27 +133,29 @@ func (cfg *Config) SetupLogger() {
 	ctrlrt.SetLogger(zap.New(zap.UseFlagOptions(&zapOptions)))
 }
 
-// PopulateAccountIdIfMissing uses sts GetCallerIdentity API to find
-// AWS AccountId when Config.AccountId is empty
-func (cfg *Config) PopulateAccountIdIfMissing() error {
-	if cfg.AccountID == "" {
-		// use sts to find AWS AccountId
-		session := session.Must(session.NewSession())
-		client := sts.New(session)
-		res, err := client.GetCallerIdentity(&sts.GetCallerIdentityInput{})
-		if err == nil {
-			cfg.AccountID = *res.Account
-		}
-		return err
+// SetAWSAccountID uses sts GetCallerIdentity API to find AWS AccountId and set
+// in Config
+func (cfg *Config) SetAWSAccountID() error {
+	// use sts to find AWS AccountId
+	session, err := session.NewSession()
+	if err != nil {
+		return fmt.Errorf("unable to create session: %v", err)
 	}
+	client := sts.New(session)
+	res, err := client.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		return fmt.Errorf("unable to get caller identity: %v", err)
+	}
+	cfg.AccountID = *res.Account
 	return nil
 }
 
 // Validate ensures the options are valid
 func (cfg *Config) Validate() error {
-	if cfg.AccountID == "" {
-		return errors.New("unable to start service controller as account ID is missing. Please pass --aws-account-id flag or set AWS_ACCOUNT_ID environment variable")
+	if err := cfg.SetAWSAccountID(); err != nil {
+		return errors.New("unable to determine account ID. Please make sure AWS credentials are setup in controller pod")
 	}
+
 	if cfg.Region == "" {
 		return errors.New("unable to start service controller as AWS region is missing. Please pass --aws-region flag or set AWS_REGION environment variable")
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"net/url"
 
+	"github.com/jaypipes/envutil"
 	flag "github.com/spf13/pflag"
 	"go.uber.org/zap/zapcore"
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.37.10
 	github.com/go-logr/logr v0.1.0
 	github.com/google/go-cmp v0.3.1
+	github.com/jaypipes/envutil v1.0.0
```
jaypipes, line 9: LOL, I was mostly only kidding in referring to my amazing envutil library, Vijay :)

pkg/config/config.go
```diff
@@ -134,13 +137,29 @@ func (cfg *Config) SetupLogger() {
 	ctrlrt.SetLogger(zap.New(zap.UseFlagOptions(&zapOptions)))
 }
 
+// PopulateAccountIdIfMissing uses sts GetCallerIdentity API to find
+// AWS AccountId when Config.AccountId is empty
+func (cfg *Config) PopulateAccountIdIfMissing() error {
```
jaypipes, line 142: Consider a simpler name of `defaultAWSAccountID` (also note it's `ID` not `Id` to be properly Gopherish ;)

pkg/config/config.go
```diff
@@ -83,12 +86,12 @@ func (cfg *Config) BindFlags() {
 	)
 	flag.StringVar(
 		&cfg.AccountID, flagAWSAccountID,
-		"",
+		envutil.WithDefault("AWS_ACCOUNT_ID", ""),
```
RedbackThomson, line 89: Please declare `AWS_ACCOUNT_ID` and `AWS_REGION` as constants at the top of this file

pkg/config/config.go
RedbackThomson, line 142: Please change `Id` to `ID` everywhere applicable
```suggestion
func (cfg *Config) PopulateAccountIDIfMissing() error {
```

pkg/config/config.go
```diff
@@ -134,13 +137,29 @@ func (cfg *Config) SetupLogger() {
...
+		session := session.Must(session.NewSession())
+		client := sts.New(session)
+		res, err := client.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+		if err == nil {
+			cfg.AccountID = *res.Account
+		}
+		return err
+	}
+	return nil
+}
```
jaypipes, line 154: Instead of requiring a change to the code-generator, how about just calling `defaultAWSAccountID` from the `Validate()` method below?

pkg/config/config.go
```diff
@@ -134,13 +137,29 @@ func (cfg *Config) SetupLogger() {
...
 }
 
+// PopulateAccountIdIfMissing uses sts GetCallerIdentity API to find
+// AWS AccountId when Config.AccountId is empty
+func (cfg *Config) PopulateAccountIdIfMissing() error {
+	if cfg.AccountID == "" {
+		// use sts to find AWS AccountId
+		session := session.Must(session.NewSession())
+		client := sts.New(session)
+		res, err := client.GetCallerIdentity(&sts.GetCallerIdentityInput{})
```
RedbackThomson, line 147: I just checked, and this is *always* allowed. So we are safe 👍🏼 

> No permissions are required to perform this operation. If an administrator adds a policy to your IAM user or role that explicitly denies access to the sts:GetCallerIdentity action, you can still perform this operation

pkg/config/config.go
```diff
@@ -134,13 +137,29 @@ func (cfg *Config) SetupLogger() {
...
+		if err == nil {
+			cfg.AccountID = *res.Account
```
jaypipes, line 149: is `res.Account` guaranteed to be non-nil? Not sure.. just curious.

RedbackThomson (changes requested):
Seems simple enough. Please see inline changes.

go.mod
vijtrip2: haha, I did not catch the joke but I think this works fine unless you want me to remove it. : )

pkg/config/config.go
```diff
@@ -81,14 +84,9 @@ func (cfg *Config) BindFlags() {
 		"Configures the logger to use a Zap development config (encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn, no sampling), "+
 			"otherwise a Zap production config will be used (encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error), sampling).",
 	)
-	flag.StringVar(
-		&cfg.AccountID, flagAWSAccountID,
-		"",
-		"The AWS Account ID in which the service controller will create resources",
-	)
```
jaypipes, line 88: @RedbackThomson @vijtrip2 we use the `--aws-account-id` flag when setting up `make kind-test` using kustomize and non-IRSA setups. Is there a reason to remove this?

vijtrip2: @jaypipes, in those tests we inject the temp aws creds(AWS_ACCESS_KEY_ID , ...)
And if we keep doing that, this aws-account-id will be correctly determined(using sts call) without needing to pass it as flag value

As pointed out by Nick, this flag can cause inconsistencies and confusion between actual credentials being used vs account-id printed in logs

vijtrip2: >"we use the --aws-account-id flag when setting up make kind-test using kustomize and non-IRSA setups"

I did add TODO's in the PR description to remove this flag from the test-infra repo

RedbackThomson: I've been thinking about this and yeah I don't see a reason to need an `aws-account-id` flag. The only place where we use this account ID is when we add the "OwnerAccountID" field to an object. That is, if we haven't assumed another role as part of the configmap annotation. So it's safe to say that whatever credentials we use with `sts get-caller-identity` will be correct.

pkg/config/config.go
```diff
@@ -134,13 +132,27 @@ func (cfg *Config) SetupLogger() {
 	ctrlrt.SetLogger(zap.New(zap.UseFlagOptions(&zapOptions)))
 }
 
+// SetAWSAccountID uses sts GetCallerIdentity API to find AWS AccountId and set
+// in Config
+func (cfg *Config) SetAWSAccountID() error {
+	// use sts to find AWS AccountId
+	session := session.Must(session.NewSession())
```
A-Hilaly, line 139: `session.Must` will panic if `session.NewSession` returns an error. Maybe just return a wrapped error like: `fmt.Errorf("cannot create session: %v", err)`

pkg/config/config.go
```diff
@@ -134,13 +132,27 @@ func (cfg *Config) SetupLogger() {
...
+	client := sts.New(session)
+	res, err := client.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err == nil {
+		cfg.AccountID = *res.Account
+	}
+	return err
```
A-Hilaly, line 145: nit: handle `err != nil` first

pkg/config/config.go
A-Hilaly: `aws-account-id` is used in CARM to keep trace of which resource is create in which account (mainly in the logs of resource reconciliation)

pkg/config/config.go
A-Hilaly: Better would be to wrap the error like: `fmt.Errorf("cannot get caller identity: %v", err)`

pkg/config/config.go
vijtrip2: > this will definitely impact the way we CARM config/patterns in logs/configMaps/Annotations

Can you elaborate on this, @A-Hilaly ? 
All this PR is doing is just removing option for customers and automatically detect account-id. aws-account-id will still be part of Config and there are no changes in any types. I wonder how it will impact any functionalities for CARM.

RedbackThomson: CARM assumes the `--aws-account-id` as the default AWS Account ID for any resources that do not have overrides in the namespace or on the resource annotation. This pull request will infer `--aws-account-id` from the credentials, rather than rely on it being hardcoded as an argument. All of the CARM logic will remain the same, and any other reconciliation logic will too.

I think it would be a great addition (not for this PR) to have a log statement when the controller starts that prints the default account ID and role that is being used by the controller. That is, if there are no overrides, what role is creating resources. [@komish](https://github.com/komish) pointed out that this is not clear at the moment, and makes it difficult to debug when you are trying to use CARM.

pkg/config/config.go
```diff
@@ -134,13 +133,31 @@ func (cfg *Config) SetupLogger() {
...
+	cfg.AccountID = *res.Account
+	return nil
+}
+
 // Validate ensures the options are valid
 func (cfg *Config) Validate() error {
-	if cfg.AccountID == "" {
-		return errors.New("unable to start service controller as account ID is nil. Please pass --aws-account-id flag")
+	if err := cfg.SetAWSAccountID(); err != nil {
+		return errors.New("unable to determine account ID. Please make sure AWS credentials are setup in controller pod. See: https://aws-controllers-k8s.github.io/community/docs/user-docs/irsa/")
```
RedbackThomson, line 156: I would probably remove the link to the documentation. I think our docs are going to change in the next few months, and I would prefer to not have dead links in the code.

RedbackThomson (approved):
Cool! Invent and simplify!

A-Hilaly (approved):
Neat! /lgtm

brycahta:
Dang, I actually liked being able to set the Account_Id explicitly as I use a couple different accounts when developing locally.

> This flag can allow users to enter wrong account_id during controller installation which can cause incorrect controller logs and confusion

If the account_id is malformed or missing, then using your env-var-fallback as a default is a good strategy, but is the opportunity to mistype a value a good enough reason to remove a flag? Does this imply mistyping other controller flags will **not** lead to incorrect controller logs and confusion?

PR #58, opened 2021-10-13
add clarity to validate error message

acornett21:
Issue #, if available:
N/A

Description of changes:
With the removal of `--aws-account-id` the validate error message that gets propagated is less clear then it used to be. This PR aims to add more clarity to the error message that gets propagated when `ack_user_secrets` does not exist in the namespace that a given controller runs in.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -153,7 +153,7 @@ func (cfg *Config) SetAWSAccountID() error {
 // Validate ensures the options are valid
 func (cfg *Config) Validate() error {
 	if err := cfg.SetAWSAccountID(); err != nil {
-		return errors.New("unable to determine account ID. Please make sure AWS credentials are setup in controller pod")
+		return fmt.Errorf("unable to determine account ID: %v", err)
 	}
 
 	if cfg.Region == "" {


pkg/config/config.go
```diff
@@ -153,7 +153,7 @@ func (cfg *Config) SetAWSAccountID() error {
 // Validate ensures the options are valid
 func (cfg *Config) Validate() error {
 	if err := cfg.SetAWSAccountID(); err != nil {
-		return errors.New("unable to determine account ID. Please make sure AWS credentials are setup in controller pod")
+		return errors.New("unable to determine account ID. Please make sure ack_user_secrets is configured in the ack_system namespace")
```
vijtrip2, line 156: Sorry this message is very specific and may not apply to all use cases like
a) What if user is not using `ack_system` namespace, this message will completely throw them off
b) `ack_user_secrets`(assuming it is a secret), is not the recommended way to inject AWS creds in controller pod. Our docs recommend using IAM Role for Service Accounts

jaypipes: @acornett21 I agree with @vijtrip2 here. As a compromise, I might be persuaded to add to the error message above a link to our documentation for properly setting up AWS credentials for the controller pod:

https://aws-controllers-k8s.github.io/community/docs/user-docs/irsa/

Better yet, we could move this to be an actual package-level object in [`github.com/aws-controllers-k8s/runtime/pkg/errors`](https://github.com/aws-controllers-k8s/runtime/tree/main/pkg/errors)

jaypipes (changes requested):

pkg/config/config.go
acornett21: @vijtrip2 @jaypipes It seems to me that there are a few different approaches in regards to AWS credentials, and just wanted to start the discussion on what that looks like for all approaches. This line was confusing to me, since I was aware of `aws-account-id ` and `ack_user_secrets` and had to message @RedbackThomson to clarify what this new error was actually complaining about.  

Since OLM will always installs the ACK controllers in the `ack_system` namespace since this is specificed in the [deployment](https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/config/controller/deployment.yaml#L12), I assumed that was the case in the helm world, since this namespace comes from the files on disk that are used to generate the bundle. I think a link to the documentation makes it less clear in the OpenShift world since we would not be using IRSA and will be using secrets as the only option.

I have no strong opinions one way or the other, but I feel more clarity on this message is needed. However, maybe clarity for one type of deployment, makes the error message more confusing for another type of deployment and there is no all encompassing error message. I'll defer to you all since you know all the possible use cases. Thanks for the input.

A-Hilaly: We could also keep/wrap the returned error by `cfg.SetAccountID` using something like:
```go
fmt.Errorf("unable to determine account ID: %v", err)
```

vijtrip2 (approved):

A-Hilaly (approved):
:+1:

PR #59, opened 2021-10-19
Add conditional for starting adoption reconciler

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1006

Description of changes:
Use cluster introspection to determine whether the `AdoptedResource` CRD has been installed, starting the adoption reconciler only if they have been. This will allow cluster operators to intentionally leave the `AdoptedResource` uninstalled, if they wish to handle this separately.

I have tested these changes by uninstalling the `AdoptedResource` CRD and then running the controller. I see the following line in the logs to prove it is working:
```
2021-10-19T14:52:42.975-0700    INFO    adoption  AdoptedResource CRD not installed. The adoption reconciler will not be started
```

### DIFF
--- a/pkg/runtime/service_controller.go
+++ b/pkg/runtime/service_controller.go
@@ -199,13 +199,13 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 	}
 
 	adoptionInstalled, err := c.GetAdoptedResourceInstalled(mgr)
-	errLogger := c.log.WithName("adoption-setup")
+	adoptionLogger := c.log.WithName("adoption")
 	if err != nil {
-		errLogger.Error(err, "unable to determine if the AdoptedResource CRD is installed in the cluster")
+		adoptionLogger.Error(err, "unable to determine if the AdoptedResource CRD is installed in the cluster")
 	} else if !adoptionInstalled {
-		errLogger.Info("AdoptedResource CRD not installed. The adoption reconciler will not be started")
+		adoptionLogger.Info("AdoptedResource CRD not installed. The adoption reconciler will not be started")
 	} else {
-		rec := NewAdoptionReconciler(c, c.log, cfg, c.metrics, cache)
+		rec := NewAdoptionReconciler(c, adoptionLogger, cfg, c.metrics, cache)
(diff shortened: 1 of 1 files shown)


pkg/runtime/service_controller.go
```diff
@@ -87,6 +92,44 @@ func (c *serviceController) GetResourceManagerFactories() map[string]acktypes.AW
...
+	}
+
+	gv := schema.GroupVersion{
+		Group:   ackv1alpha1.GroupVersion.Group,
+		Version: ackv1alpha1.GroupVersion.Version,
+	}
+
+	// Ensure GV is supported
+	if err = discovery.ServerSupportsVersion(clientSet, gv); err != nil {
+		return false, nil
```
A-Hilaly, line 111: Maybe i'm missing something but i think this should be a return `return false, err`

pkg/runtime/service_controller.go
```diff
@@ -153,11 +198,20 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 		c.reconcilers = append(c.reconcilers, rec)
 	}
 
-	rec := NewAdoptionReconciler(c, c.log, cfg, c.metrics, cache)
-	if err := rec.BindControllerManager(mgr); err != nil {
-		return err
+	adoptionInstalled, err := c.GetAdoptedResourceInstalled(mgr)
+	errLogger := c.log.WithName("adoption-setup")
```
A-Hilaly, line 202: nit: 
```suggestion
	adoptionLogger := c.log.WithName("adoption-setup")
```

A-Hilaly: You could also pass it to the adoption reconciler `rec := NewAdoptionReconciler(c, adoptionLogger, cfg, c.metrics, cache)`

pkg/runtime/service_controller.go
acornett21: I think for where this method is called, some errors are blocking and some are non-blocking, based on the `if/else` logic where this is called. Maybe, inline comments in this method are needed to avoid confusion and state which error blocks and which does now.

PR #61, opened 2021-11-10
clear conditions on loop start and ensure synced

jaypipes:
This is the first patch that looks at cleaning up our handling of
Condition objects for resources. We make a change to the common runtime
to clear all Condition objects from the desired resource's
Status.Conditions collection at the beginning of the reconciliation loop
and add a deferred call to a function that ensures that an
ACK.ResourceSynced condition has been placed on the resource if the
resource manager has failed to add one.

Future patches will add functionality to the code generator to allow
programmatic control over when generated resource managers set
ACK.ResourceSynced conditions on their resources.

Issue: aws-controllers-k8s/community#1030

### DIFF
--- a/pkg/condition/condition.go
+++ b/pkg/condition/condition.go
@@ -157,3 +157,10 @@ func LateInitializationInProgress(subject acktypes.ConditionManager) bool {
 	c := LateInitialized(subject)
 	return c != nil && c.Status == corev1.ConditionFalse
 }
+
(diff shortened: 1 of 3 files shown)


vijtrip2:
/approve

This looks good! But this still leaves individual resource managers to implement the `ACK.ResourceSynced` Status as `True/False` , which is currently missing.
And we should be aiming for that in future changes in code-generator. Correct?

PR #64, opened 2022-01-11
attempt to patchStatus regardless of finalizer

vijtrip2:
Description of changes:

* ACK reconciler marks a resource as `Managed` before performing the `CREATE` call
* ResourceReference in ACK reconciler is resolved before `READ_ONE` call which precedes `CREATE` call
* When ResourceReference resolution fails, ACK reconciler needs to set `Status.Condition` indicating reference resolution failed
* Hence the status patching needs to be done for unmanaged resources too
* `NotFound` error while status patching of deleted resource is caught and not logged in controller logs

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -507,12 +507,15 @@ func (r *resourceReconciler) patchResourceStatus(
 		latest.RuntimeObject(),
 		client.MergeFrom(desired.DeepCopy().RuntimeObject()),
 	)
-	rlog.Exit("kc.Patch (status)", err)
-	if err != nil {
-		return err
+	if err == nil {
+		rlog.Debug("patched resource status")
+	} else if apierrors.IsNotFound(err) {
+		// reset the NotFound error so it is not printed in controller logs
+		// providing false positive error
+		err = nil
 	}
-	rlog.Debug("patched resource status", "latest", latest)
-	return nil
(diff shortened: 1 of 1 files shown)


jaypipes (changes requested):
Please see https://github.com/aws-controllers-k8s/community/issues/931. We will need to figure out a different solution to this, because this conditional check on whether the resource is under management was specifically added in to prevent the issue described in 931.

pkg/runtime/reconciler.go
```diff
@@ -507,12 +507,17 @@ func (r *resourceReconciler) patchResourceStatus(
...
 		client.MergeFrom(desired.DeepCopy().RuntimeObject()),
 	)
-	rlog.Exit("kc.Patch (status)", err)
-	if err != nil {
-		return err
+	if err == nil {
+		rlog.Debug("patched resource status")
+	} else if apierrors.IsNotFound(err) {
+		rlog.Debug("skipping resource status patching because " +
+			"the resource no longer exists")
```
jaypipes, line 514: I would just remove the above debug log message entirely. :) It's not particularly useful and kind of reminds me of the error message that was originally problematic

jaypipes (changes requested):
femto-nit inline, otherwise I'm good with this!

jaypipes (approved):
:+1:

PR #66, opened 2022-01-19
ensure Status.Conditions before patching resource status

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1133

Description of changes:
* the logic for ensuring conditions was never been executed because the resource parameter(`latest`) passed in the defer call was always nil . See [Why](https://stackoverflow.com/a/42703862)
* If the `desired` and `latest` state have no delta, set `ResourceSynced` condition to `True`
* Set `ResourceSynced` condition to `False` when there is error in `LateInitialization`
* For any reconciler error, except Terminal, if ResourceSynced condition is missing, set it to `False`
* If there are no reconciler error, and ResourceSynced condition is missing, set it to `Unknown`
* Revert aws-sdk-go upgrade and handle that in separate release.
---

Tested these changes locally for iam-controller and tests pass successfully validating ResourceSynced condition.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -244,6 +244,15 @@ func (r *resourceReconciler) Sync(
 	// Attempt to late initialize the resource. If there are no fields to
 	// late initialize, this operation will be a no-op.
 	if latest, err = r.lateInitializeResource(ctx, rm, latest); err != nil {
+		// TODO(vijtrip2): move this condition handling to generated
+		// AWSResourceManager.LateInitialize() method
+
+		// Whenever late initialization fails for a resource, set ACK.ResourceSynced
+		// condition explicitly to "False"
+		// Setting this explicitly to False is required because ACK.ResourceSynced
+		// condition can be True due to successful Create/Update call OR no
+		// Create/Update call in reconciler loop
+		ackcondition.SetSynced(latest, corev1.ConditionFalse, nil, nil)
 		return latest, err
 	}
 	return r.handleRequeues(ctx, latest)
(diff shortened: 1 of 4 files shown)


pkg/runtime/reconciler.go
```diff
@@ -208,7 +208,6 @@ func (r *resourceReconciler) Sync(
 	var latest acktypes.AWSResource // the newly created or mutated resource
 
 	r.resetConditions(ctx, desired)
-	defer r.ensureConditions(ctx, latest, err)
```
A-Hilaly, line 211: Would something like below behave differently?
```go
defer func() {
    r.ensureConditions(ctx, latest, err)
}()
```

jaypipes: Yes. TIL...

jaypipes: The irony here is that if we'd passed a *pointer* to an error and AWSResource interface, this would have worked :)

pkg/runtime/reconciler.go
```diff
@@ -718,6 +717,10 @@ func (r *resourceReconciler) HandleReconcileError(
 	err error,
 ) (ctrlrt.Result, error) {
 	if ackcompare.IsNotNil(latest) {
+		// ensure that required conditions are set in latest resource before
+		// patching resource status
+		r.ensureConditions(ctx, latest, err)
```
jaypipes, line 722: My only worry about this is that `handleRequeues` runs *before* `handleReconcileError` and if we aren't ensuring a ResourceSynced condition is present (with either a False or Unknown condition value) on the `latest` resource by the time that `handleRequeues` runs, then the request won't be requeued...

jaypipes (changes requested):
I'm curious to see the behaviour difference between the approach @A-Hilaly suggests and the approach in this patch of calling `ensureConditions` in `HandleReconcileError`. I think (see inline comment) that because `handleRequeues` happens *before* `HandleReconcileError`, the requeues won't occur properly because ResourceSynced condition won't exist on the `latest` resource.

jaypipes (approved):
:+1: nice work on this and thank you Vijay!

PR #71, opened 2022-02-03
Use original status when making api-server `Status.Update` call in adoption reconciler

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1161

Description of changes:
* The Status update from AdoptionReconciler was not useful because the Create call before Status.Update was resetting status of the CustomResource.
* Earlier this similar kind of problem was also present in `reconciler.go` and it was fixed by keeping an original copy of Status before making Create/Update calls
* I used the same solution in AdoptionReconciler.

----------

* Discovered this issue while debugging SageMaker ModelPackage adoption test. ModelPackage uses ARN as identifier which needs to be set inside `Status.ACKResourceMetadata.ARN` field of CustomResource.
* Since the Identifier was never getting correctly set, the `ReadOne` call inside `reconciler.go` was failing while reconciling the CustomResource.

--------------

* Validated by running SageMaker end-to-end tests successfully.

### DIFF
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -224,10 +224,16 @@ func (r *adoptionReconciler) Sync(
 	}, described.RuntimeObject()); err != nil {
 		if apierrors.IsNotFound(err) {
 			// If Adopted AWS resource was not found in k8s, create it.
+
+			// Before creation, Keep the copy of original described object
+			// because after the create call, Status gets set to empty
+			describedCopy := described.DeepCopy()
 			if err := r.kc.Create(ctx, described.RuntimeObject()); err != nil {
 				return r.onError(ctx, desired, err)
 			}
-
+			// reset the status of described object to original value before
+			// making the Status Update call
+			described.SetStatus(describedCopy)
 			if err := r.kc.Status().Update(ctx, described.RuntimeObject()); err != nil {
 				return r.onError(ctx, desired, err)
 			}
--- a/pkg/runtime/adoption_reconciler_test.go
+++ b/pkg/runtime/adoption_reconciler_test.go
@@ -72,13 +72,14 @@ func mockReconciler() (acktypes.AdoptedResourceReconciler, *ctrlrtclientmock.Cli
 	), kc, apiReader
 }
 
-func mockDescriptorAndAWSResource() (*ackmocks.AWSResourceDescriptor, *ackmocks.AWSResource) {
+func mockDescriptorAndAWSResource() (*ackmocks.AWSResourceDescriptor, *ackmocks.AWSResource, *ackmocks.AWSResource) {
 	des := &ackmocks.AWSResourceDescriptor{}
 	emptyRuntimeObject := &ctrlrtclientmock.Object{}
 	res := &ackmocks.AWSResource{}
+	resDeepCopy := &ackmocks.AWSResource{}
 	des.On("EmptyRuntimeObject").Return(emptyRuntimeObject)
 	des.On("ResourceFromRuntimeObject", emptyRuntimeObject).Return(res)
-	return des, res
+	return des, res, resDeepCopy
 }
 
 func mockManager() *ackmocks.AWSResourceManager {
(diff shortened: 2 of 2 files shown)


ryansteakley (approved):

pkg/runtime/adoption_reconciler.go
```diff
@@ -224,10 +224,16 @@ func (r *adoptionReconciler) Sync(
 	}, described.RuntimeObject()); err != nil {
 		if apierrors.IsNotFound(err) {
 			// If Adopted AWS resource was not found in k8s, create it.
+
+			// Before creation, Keep the copy of original described object
+			// because after the create call, Status gets set to empty
+			describedCopy := described.DeepCopy()
 			if err := r.kc.Create(ctx, described.RuntimeObject()); err != nil {
```
RedbackThomson, line 231: Why not just use the described copy in the create?

pkg/runtime/adoption_reconciler_test.go
```diff
@@ -435,9 +437,13 @@ func assertAWSResourceCreation(
 ) {
 	if expectedCreation {
 		kc.AssertCalled(t, "Create", ctx, res.RuntimeObject())
+		res.AssertCalled(t, "DeepCopy")
+		res.AssertCalled(t, "SetStatus", res)
 		statusWriter.AssertCalled(t, "Update", ctx, res.RuntimeObject())
 	} else {
 		kc.AssertNotCalled(t, "Create", ctx, res.RuntimeObject())
+		res.AssertNotCalled(t, "DeepCopy")
+		res.AssertNotCalled(t, "SetStatus", res)
```
RedbackThomson, line 446: Not to be picky, but I don't think these tests really *test* the functionality. It's the order in which these are called, and how we manipulate the underlying pointers, that is essential to this PR doing what it intends to do. Simply seeing it call `DeepCopy` does not validate that it's successful.

RedbackThomson (commented):
Left some comments inline

pkg/runtime/adoption_reconciler.go
vijtrip2: describedCopy is used only for keeping a copy of Status.

* The create call will modify the `described` object because it's passed by reference
* We use the same modified "described" object for status patching
* But before status patching, we copy the original status from `describedCopy`

Does it answer your question? LMK if you were asking for something else.

pkg/runtime/adoption_reconciler_test.go
vijtrip2: Fair point, I have updated the test to at least use a different object for DeepCopy.
The new assertions test that SetStatus is invoked after retrieving the DeepCopy object.

pkg/runtime/adoption_reconciler.go
```diff
@@ -224,10 +224,16 @@ func (r *adoptionReconciler) Sync(
 	}, described.RuntimeObject()); err != nil {
 		if apierrors.IsNotFound(err) {
 			// If Adopted AWS resource was not found in k8s, create it.
+
+			// Before creation, Keep the copy of original described object
+			// because after the create call, Status gets set to empty
+			describedCopy := described.DeepCopy()
```
A-Hilaly, line 230: How can a described object have a non empty .Status but does not exist in the k8s cluster?

vijtrip2: If the custom resource for adopted resource does not exist in K8s cluster, the create call is made. Since we pass the `described` object in the create call, it's status gets updated to empty to match etcd copy of custom resource.

A-Hilaly: Ohhh nvm i didn't realize this was the adoption controller

A-Hilaly:
Looks good for me :+1: 
/approve

RedbackThomson (approved):
Thank you for adding those new tests. 

/lgtm

pkg/runtime/adoption_reconciler.go
```diff
@@ -224,10 +224,16 @@ func (r *adoptionReconciler) Sync(
...
 			if err := r.kc.Create(ctx, described.RuntimeObject()); err != nil {
 				return r.onError(ctx, desired, err)
 			}
-
+			// reset the status of described object to original value before
+			// making the Status Update call
+			described.SetStatus(describedCopy)
 			if err := r.kc.Status().Update(ctx, described.RuntimeObject()); err != nil {
```
jaypipes, line 237: Why are we even calling `r.kc.Status().Update()` here? We just created the CR in the call above to `r.kc.Create()` so why are we immediately updating it here?

surajkota: afaik, spec and status are not applied at same time. you need separate call

RedbackThomson: Adoption of resources that use the ARN as their primary identifier mean we have to place the status onto the object before it can be reconciled. We first call `.Create()`, with which we can only add the spec to the object. Then we need to call `.Status().Update()` to set the ARN.

vijtrip2: We need to update the status of new resource as well because for some resources the identifier is stored inside `Status.ACKResourceMetadata.ARN` . 
The Create call creates new Custom Resource with Spec and when `reconciler.go` tries to perform `ReadOne`, it fails with `NotFoundError`. NotFoundError is expected because ReadOne call expects a `Status.ACKResourceMetadata.ARN`. 

In non-adoption cases, this will result in resource creation but reconciler.go has a check that if NotFoundError occurs for AdoptedResource, Fail !!
That's why Sagemaker AdoptedModelPackage was never able reconcile from `reconciler.go` even when successfully reconciled inside `adoption_reconciler.go`

jaypipes: thanks all. I always seem to forget the fact that there is no REPLACE operation on both spec and status at once...

PR #73, opened 2022-02-18
handle requeues after setting Synced condition

vijtrip2:
Issue: https://github.com/aws-controllers-k8s/community/issues/1186

Description of changes:
* Call `handleRequeues` method after setting the `ACK.ResourceSynced` condition
* Use `AWSResourceManager.IsSynced` method to determine the `ACK.ResourceSynced` condition
* Only set `ACK.ResourceSynced` condition inside `ensureConditions` function when the condition is originally missing
* Update existing unit-tests to include `rm.IsSynced()` call and Add a new unit-test when `rm.IsSynced` returns false
------

* Also tested locally using Sagemaker controller, whose e2e tests verify that SyncedCondition is properly set in success and failure scenarios. This ensures there is no breakage in existing functionality.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -293,23 +293,32 @@ func (r *resourceReconciler) ensureConditions(
 	if ackcondition.Synced(res) == nil {
 		condStatus := corev1.ConditionFalse
 		synced := false
+		condMessage := ackcondition.NotSyncedMessage
+		var condReason string
 		rlog.Enter("rm.IsSynced")
 		if synced, err = rm.IsSynced(ctx, res); err == nil && synced {
 			condStatus = corev1.ConditionTrue
+			condMessage = ackcondition.SyncedMessage
+		} else if err != nil {
+			condReason = err.Error()
 		}
 		rlog.Exit("rm.IsSynced", err)
+
 		if reconcileErr != nil {
+			condReason = reconcileErr.Error()
 			if reconcileErr == ackerr.Terminal {
 				// A terminal condition by its very nature indicates a stable state
 				// for a resource being synced. The resource is considered synced
 				// because its state will not change.
 				condStatus = corev1.ConditionTrue
+				condMessage = ackcondition.SyncedMessage
 			} else {
 				// For any other reconciler error, set synced condition to false
 				condStatus = corev1.ConditionFalse
+				condMessage = ackcondition.NotSyncedMessage
 			}
 		}
-		ackcondition.SetSynced(res, condStatus, nil, nil)
+		ackcondition.SetSynced(res, condStatus, &condMessage, &condReason)
 	}
 }
 
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -346,20 +346,6 @@ func (r *resourceReconciler) createResource(
 		return nil, err
 	}
 
-	// Once the desired resource is marked as managed (i.e. Finalizer is added
-	// inside the resource metadata and the resource is patched inside etcd), the
-	// patch operation updates the desired resource spec with spec present in
-	// etcd. This means if any references were resolved, they get reset after
-	// the patch call. So we resolve the references again before calling the
-	// rm.Create() method.
-	rlog.Enter("rm.ResolveReferences")
-	resolvedRefDesired, err := rm.ResolveReferences(ctx, r.apiReader, desired)
-	rlog.Exit("rm.ResolveReferences", err)
-	if err != nil {
-		return resolvedRefDesired, err
-	}
-	desired = resolvedRefDesired
-
 	rlog.Enter("rm.Create")
 	latest, err = rm.Create(ctx, desired)
 	rlog.Exit("rm.Create", err)
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -343,6 +343,20 @@ func (r *resourceReconciler) createResource(
 		return nil, err
 	}
 
+	// Once the desired resource is marked as managed (i.e. Finalizer is added
+	// inside the resource metadata and the resource is patched inside etcd), the
+	// patch operation updates the desired resource spec with spec present in
+	// etcd. This means if any references were resolved, they get reset after
+	// the patch call. So we resolve the references again before calling the
+	// rm.Create() method.
+	rlog.Enter("rm.ResolveReferences")
+	resolvedRefDesired, err := rm.ResolveReferences(ctx, r.apiReader, desired)
+	rlog.Exit("rm.ResolveReferences", err)
+	if err != nil {
+		return resolvedRefDesired, err
+	}
+	desired = resolvedRefDesired
+
 	rlog.Enter("rm.Create")
 	latest, err = rm.Create(ctx, desired)
 	rlog.Exit("rm.Create", err)
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -188,7 +188,11 @@ func (r *resourceReconciler) reconcile(
 		res, _ = rm.ResolveReferences(ctx, r.apiReader, res)
 		return r.deleteResource(ctx, rm, res)
 	}
-	return r.Sync(ctx, rm, res)
+	latest, err := r.Sync(ctx, rm, res)
+	if err != nil {
+		return latest, err
+	}
+	return r.handleRequeues(ctx, latest)
 }
 
 // Sync ensures that the supplied AWSResource's backing API resource
@@ -209,7 +213,7 @@ func (r *resourceReconciler) Sync(
 
 	r.resetConditions(ctx, desired)
 	defer func() {
-		r.ensureConditions(ctx, latest, err)
+		r.ensureConditions(rm, ctx, latest, err)
 	}()
 
 	isAdopted := IsAdopted(desired)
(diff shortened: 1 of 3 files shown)


pkg/runtime/reconciler.go
```diff
@@ -279,6 +274,7 @@ func (r *resourceReconciler) resetConditions(
 // ensureConditions examines the supplied resource's collection of Condition
 // objects and ensures that an ACK.ResourceSynced condition is present.
 func (r *resourceReconciler) ensureConditions(
+	rm acktypes.AWSResourceManager,
```
A-Hilaly, line 277: `ctx context.Context` should always be the first parameter of a function

pkg/runtime/reconciler.go
```diff
@@ -244,18 +248,9 @@ func (r *resourceReconciler) Sync(
...
-		// Whenever late initialization fails for a resource, set ACK.ResourceSynced
-		// condition explicitly to "False"
-		// Setting this explicitly to False is required because ACK.ResourceSynced
-		// condition can be True due to successful Create/Update call OR no
-		// Create/Update call in reconciler loop
-		ackcondition.SetSynced(latest, corev1.ConditionFalse, nil, nil)
 		return latest, err
 	}
-	return r.handleRequeues(ctx, latest)
+	return latest, nil
```
A-Hilaly, line 253: nit:
```suggestion
	return r.lateInitializeResource(ctx, rm, latest)
```

pkg/runtime/reconciler.go
```diff
@@ -292,12 +288,13 @@ func (r *resourceReconciler) ensureConditions(
...
-		rlog.Debug("resource missing ACK.ResourceSynced condition")
-		// only the resource manager will know whether the resource is in a
-		// stable sync state. Even if we got no error back from the
-		// create/update operations, we can only set this to Unknown.
-		condStatus := corev1.ConditionUnknown
+	// If the ACK.ResourceSynced condition is not set using the custom hooks,
+	// determine the Synced condition using "rm.IsSynced" method
+	if ackcondition.Synced(res) == nil {
+		condStatus := corev1.ConditionFalse
+		if synced, _ := rm.IsSynced(ctx, res); synced {
```
A-Hilaly, line 295: I know that generated `IsSynced` function returns only nil `err` for now. Still we should handle this error properly?

A-Hilaly (commented):
Sweet! Left few comments below

pkg/runtime/reconciler.go
vijtrip2: This was intentional because we have defer exit(err) call in Sync function. So I wanted to assign the error into "err" variable.

pkg/runtime/reconciler.go
vijtrip2: I did think about it before sending PR and my thought process was that we want to mark `ResourceSynced` condition False when `rm.IsSynced` function returns false or non-nil error. And the code in PR did that.

Saying that I also agree to your point and I have updated the code to handle error from `rm.IsSynced` function. Following are controller logs for error vs no-error case now.

Error:
```
1.645553470691295e+09   DEBUG   ackrt   >> r.ensureConditions   {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2}
1.6455534706913338e+09  DEBUG   ackrt   >>> rm.IsSynced {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2}
1.645553470691346e+09   DEBUG   ackrt   <<< rm.IsSynced {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2, "error": "error from IsSynced"}
1.645553470691356e+09   DEBUG   ackrt   << r.ensureConditions   {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2}

```

No-Error:
```
1.6455532315793998e+09  DEBUG   ackrt   >> r.ensureConditions   {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2}
1.645553231579408e+09   DEBUG   ackrt   >>> rm.IsSynced {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2}
1.645553231579413e+09   DEBUG   ackrt   <<< rm.IsSynced {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2}
1.645553231579419e+09   DEBUG   ackrt   << r.ensureConditions   {"account": "******", "role": "", "region": "us-west-2", "kind": "API", "namespace": "default", "name": "api-ref-test", "is_adopted": false, "generation": 2}
```

pkg/runtime/reconciler.go
A-Hilaly: Make sense! Thanks for clarifying @vijtrip2

pkg/runtime/reconciler_test.go
```diff
@@ -169,8 +165,8 @@ func TestReconcilerUpdate(t *testing.T) {
 		conditions := args.Get(0).([]*ackv1alpha1.Condition)
 		assert.Equal(t, 1, len(conditions))
 		cond := conditions[0]
-		assert.Equal(t, cond.Type, ackv1alpha1.ConditionTypeResourceSynced)
-		assert.Equal(t, cond.Status, corev1.ConditionUnknown)
+		assert.Equal(t, ackv1alpha1.ConditionTypeResourceSynced, cond.Type)
```
A-Hilaly, line 168: 👍

A-Hilaly (approved):
LGTM overall. Great testing @vijtrip2
/approve

pkg/runtime/reconciler.go
jaypipes: The call to `exit(err)` doesn't actually work correctly anyway, because of this:

https://stackoverflow.com/questions/42703707/when-defer-func-evaluates-its-parameters/42703862#42703862

We need to change the `exit` function to take `*error` instead of `error`.

jaypipes (changes requested):
@vijtrip2 please separate this into two PRs, one for the code changes to call `IsSynced` in `ensureConditions` and another for the resource reference resolution bits. It's difficult to review this PR as it stands because these two code changes are unrelated to each other.

pkg/runtime/reconciler.go
```diff
@@ -346,6 +346,20 @@ func (r *resourceReconciler) createResource(
 		return nil, err
 	}
 
+	// Once the desired resource is marked as managed (i.e. Finalizer is added
```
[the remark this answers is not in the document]
vijtrip2: God idea! I will consider this in the separate PR as advised. 
But in this PR, I did not want to introduce any major change in reconciler by moving the `setResourceManaged` call and the extra API server calls did not seem a lot since this code only runs once during lifecycle of a resource(once during create).

pkg/runtime/reconciler.go
```diff
@@ -292,12 +288,16 @@ func (r *resourceReconciler) ensureConditions(
...
+	// If the ACK.ResourceSynced condition is not set using the custom hooks,
+	// determine the Synced condition using "rm.IsSynced" method
+	if ackcondition.Synced(res) == nil {
+		condStatus := corev1.ConditionFalse
+		synced := false
+		rlog.Enter("rm.IsSynced")
+		if synced, err = rm.IsSynced(ctx, res); err == nil && synced {
+			condStatus = corev1.ConditionTrue
+		}
+		rlog.Exit("rm.IsSynced", err)
```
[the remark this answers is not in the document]
vijtrip2: I will also create a separate issue for this.

pkg/runtime/reconciler.go
```diff
@@ -292,12 +288,16 @@ func (r *resourceReconciler) ensureConditions(
...
 
-	if syncedCond := ackcondition.Synced(res); syncedCond == nil {
-		rlog.Debug("resource missing ACK.ResourceSynced condition")
-		// only the resource manager will know whether the resource is in a
-		// stable sync state. Even if we got no error back from the
-		// create/update operations, we can only set this to Unknown.
-		condStatus := corev1.ConditionUnknown
+	// If the ACK.ResourceSynced condition is not set using the custom hooks,
+	// determine the Synced condition using "rm.IsSynced" method
+	if ackcondition.Synced(res) == nil {
```
RedbackThomson, line 293: Do we want to check for `Unknown` as well?

RedbackThomson (approved):
Just one question, otherwise good to go

pkg/runtime/reconciler.go
vijtrip2: Before this line in code, only custom hooks set the `ACK.ResourceSynced` condition, and I am sure none of them set is as `Unknown`. Now the recommendation is to use `IsSynced` method rather than custom hooks, so I do not think we need to check for `Unknown` status.

pkg/runtime/reconciler.go
A-Hilaly: Most of the errors returned by error packages are pointers e.g: https://github.com/aws/aws-sdk-go/blob/1f3b1b4bd14f4187976f81ea5bea1f9279a57122/aws/awserr/types.go#L50, and `error` type is an interface. I'm not sure if `*error` will work but curious to see it in action

pkg/runtime/reconciler.go
```diff
@@ -292,12 +288,16 @@ func (r *resourceReconciler) ensureConditions(
...
+		condStatus := corev1.ConditionFalse
+		synced := false
+		rlog.Enter("rm.IsSynced")
+		if synced, err = rm.IsSynced(ctx, res); err == nil && synced {
```
A-Hilaly, line 297: Should we also store the err in a condition if `IsSynced` fails?

vijtrip2: Good idea, it is helpful from user perspective to see the error message in Condition message.

jaypipes (approved):
Rock on, thanks for separating the PRs :)

PR #74, opened 2022-02-23
Move `fieldpath` from `code-generator`

RedbackThomson:
Description of changes:
Move the `fieldpath` package from `code-generator` into runtime so the logic can be reused for runtime operations needed for the field export implementation.


RedbackThomson:
/hold
Looking into alternative methods for field export. This may not be necessary.

PR #75, opened 2022-02-25
Add `FieldExport` CRD and field export reconciler

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/740

Description of changes:
Creates a new `FieldExport` CRD as outlined [this proposal](https://github.com/aws-controllers-k8s/community/pull/1183). Also creates a new reconciler that reconciles changes to `FieldExport` and changes to any ACK resource recognised by the current controller.

The path for reconciling a `FieldExport`:
1. Ensure the `FieldExport` `Source` resource group matches the current controller's group
2. Mark as managed (add finalizer)
3. Describe the source object
3a. If no source object is found, stop reconciling
4. Use the [`gojq` library](https://github.com/itchyny/gojq) to parse the unstructured source object 
5. Write to `ConfigMap`/`Secret`

The path when any ACK resource is updated:
1. Ensure the resource has the synced condition set to true
2. List and filter all `FieldExport` types in the namespace that reference the current resource as its `Source`
3. For each matching `FieldExport` do steps 3 and 4 from the previous path

The reconciler converts all primitive types into strings (ConfigMap is a string-string map, Secret is a string-byte[] map). If the jq query returns a list of results, we will only take the first result (this PR does not support exporting slices/maps). If the jq query returns a struct, we will not take any action, treating the query as a failure. 

The reconciler will not delete any values from the ConfigMap or Secret. If the referenced field is removed from the object, the reconciler will no-op, rather than overwriting the data with an empty string. When the `FieldExport` is deleted, the ConfigMap or Secret will also be left untouched - meaning it contains whatever was last written to it.

Upon creating a `FieldExport` CR, the reconciler will initially attempt to export the field. If it fails, it will not attempt to retry. After the first sync, the reconciler will only wait until the `ResourceVersion` of the source object changes for it to trigger a new update.


apis/core/v1alpha1/adopted_resource.go
```diff
@@ -20,7 +20,7 @@ import (
 // AdoptedResourceSpec defines the desired state of the AdoptedResource.
 type AdoptedResourceSpec struct {
 	// +kubebuilder:validation:Required
-	Kubernetes *TargetKubernetesResource `json:"kubernetes"`
+	Kubernetes *AdoptedResourceTarget `json:"kubernetes"`
```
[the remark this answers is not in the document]
RedbackThomson: It's just been renamed. The only thing that changes are the docstrings

apis/core/v1alpha1/identifiers.go
```diff
@@ -26,14 +30,34 @@ type AWSIdentifiers struct {
...
-type TargetKubernetesResource struct {
-	// +kubebuilder:validation:Required
-	Group string `json:"group"`
-	// +kubebuilder:validation:Required
-	Kind     string             `json:"kind"`
-	Metadata *PartialObjectMeta `json:"metadata,omitempty"`
+// NamespacedTargetKubernetesResource provides all the values necessary to identify an ACK
+// resource of a given type (within the same namespace).
+type NamespacedTargetKubernetesResource struct {
+	metav1.GroupKind `json:""`
```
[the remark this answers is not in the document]
RedbackThomson: Other way around. Including the json tag inlines it.

apis/core/v1alpha1/adopted_resource.go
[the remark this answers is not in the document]
jaypipes: Existing AdoptedResource CRs in an existing installation of a controller would fail to be read into the controller, no? Since the over-wire format of the schema would be different...

apis/core/v1alpha1/identifiers.go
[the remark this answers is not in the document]
jaypipes: Not according to this: https://stackoverflow.com/questions/43077665/how-do-i-inline-fields-when-marshalling-json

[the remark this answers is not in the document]
RedbackThomson: Not sure what to say haha. The `AdoptedResource` CRD remains the same if I do it this way - you can check the diff to verify.

apis/core/v1alpha1/adopted_resource.go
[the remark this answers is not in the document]
RedbackThomson: The types are still identical, and would share the same bit packing... I believe. Definitely something I could look into. I would like to think that changing the Go type name is irrelevant :(

apis/core/v1alpha1/identifiers.go
```diff
@@ -26,14 +30,34 @@ type AWSIdentifiers struct {
...
+
+// ResourceFieldSelector provides the values necessary to identify an individual
+// path on an individual K8s resource.
+type ResourceFieldSelector struct {
+	Resource NamespacedTargetKubernetesResource `json:"resource"`
+	Path     *string                            `json:"path"`
+}
+
+// FieldExportOutputSelector provides the values necessary to identify the
+// output path for a field export (within the same namespace).
```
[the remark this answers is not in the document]
RedbackThomson: Yeah this comment is incorrect. The output can be into a separate namespace. The default is the same namespace (which is why `namespace` is optional).

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,673 @@
...
+	// Look up the rmf for the given target resource GVK
+	rmf, ok := (r.sc.GetResourceManagerFactories())[sourceGK.String()]
+	if !ok {
+		return ackerr.ResourceManagerFactoryNotFound
+	}
+
+	sourceObject, err := r.getSourceResource(ctx, rmf.ResourceDescriptor(), sourceName)
+	if err != nil {
+		return r.onError(ctx, res, err)
+	}
```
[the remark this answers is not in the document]
RedbackThomson: > Is this the user experience we want?

Nope. In fact this is what I was trying to avoid, but I got caught up in the `err != nil` copypasta. I will replace this with a return success. No need to requeue it, since the field export reconciler will run again once the resource has been created and export it (now that we have marked this field export as managed).

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,673 @@
...
+	ackrtlog.InfoFieldExport(r.log, desired, "patching target secret")
+	err = r.kc.Patch(ctx, secret, patch)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *fieldExportReconciler) FilterAllExports(
```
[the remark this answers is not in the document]
RedbackThomson: Docstring for this method is placed on the interface definition. I didn't want to duplicate it. 
I also wanted it to be public so we could unit test it.

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,673 @@
...
+	}
+
+	var requeueNeeded *requeue.RequeueNeeded
+	if errors.As(err, &requeueNeeded) {
+		r.log.V(1).Info(
+			"requeue needed error",
+			"error", requeueNeeded.Unwrap(),
+		)
+		return ctrlrt.Result{Requeue: true}, nil
+	}
```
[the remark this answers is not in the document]
RedbackThomson: No, in fact I don't want to ever requeue the `FieldExport` resource unless it explicitly throws an error. If it's in a recoverable condition, either:
- the user needs to manually patch it to bring it back to working order
- the resource needs to be created/updated for it to become valid again

Either of these conditions will automatically prompt a requeue

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,673 @@
...
+	UnstructuredConverter runtime.UnstructuredConverter = runtime.DefaultUnstructuredConverter
+)
+
+// fieldExportReconciler is responsible for reconciling the state of any field
+// export CRs that target resources supported by the current controller.
+// It implements the upstream controller-runtime `Reconciler` interface.
+type fieldExportReconciler struct {
+	reconciler
+	// rd is only used if binding to an ACK resource (not `FieldExport`)
+	rd *acktypes.AWSResourceDescriptor
```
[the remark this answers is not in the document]
RedbackThomson: Yeah I wanted to ask you about the best practices for something like this. I wanted to use this as some sort of flag that the reconciler is acting on a particular AWS resource (not a `FieldExport` type). I was using the fact that a pointer can be nil-able as this flag. Do you think it would be better to just explicitly put a boolean, instead?

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	// Update the field
+	patch := client.StrategicMergeFrom(cm.DeepCopy())
+	if cm.Data == nil {
+		cm.Data = make(map[string]string, 1)
+	}
+	cm.Data[key] = sourceValue
+
+	ackrtlog.DebugFieldExport(r.log, desired, "patching target config map")
```
vijtrip2, line 379: nit: including the name of configmap into the log message

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+	return err
+}
+
+// onSuccess will patch the FieldExport with a synced condition and
+// return any errors that occurred while patching
+func (r *fieldExportReconciler) onSuccess(
+	ctx context.Context,
+	res *ackv1alpha1.FieldExport,
+) error {
+	r.clearConditions(ctx, res)
```
vijtrip2, line 482: * Should we use `ACK.ResourceSynced` condition for FieldExport resource as well or create a new Condition `ACK.FieldsExported` ?

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+		recoverableCondition = &ackv1alpha1.Condition{
+			Type: ackv1alpha1.ConditionTypeRecoverable,
+		}
+		res.Status.Conditions = append(res.Status.Conditions, recoverableCondition)
+	}
+
+	var errMessage string
+	if err != nil {
+		errMessage = err.Error()
+		recoverableCondition.Status = corev1.ConditionFalse
```
vijtrip2, line 525: If the error is not Terminal error then recoverable should be set to true, correct?

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+		recoverableCondition.Message = &errMessage
+	} else {
+		recoverableCondition.Message = nil
```
vijtrip2, line 528: If the error is nil, recoverable condition should be removed, correct?

vijtrip2 (commented):
Excellent work! Left some minor comments.

pkg/runtime/log/resource.go
```diff
@@ -225,3 +225,63 @@ func expandAdoptedResourceFields(
...
+
+// DebugFieldExport writes a supplied log message about a field export that
+// includes a set of standard log values for the resource's kind, namespace, name, etc
+func DebugFieldExport(
+	log logr.Logger,
+	res *v1alpha1.FieldExport,
+	msg string,
+	additionalValues ...interface{},
+) {
+	AdaptFieldExport(log, res, additionalValues...).V(1).Info(msg)
```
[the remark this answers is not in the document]
RedbackThomson: This is the same as the `DebugAdoptedResource` implementation. I believe we are using `V(1)` to denote debug.

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+		// Read only field export objects
+		&ackv1alpha1.FieldExport{},
+	).WithEventFilter(
+		predicate.GenerationChangedPredicate{},
+	).Complete(r)
+}
+
+// BindServiceResourceToManager binds a given AWS resource descriptor to the
+// controller manager
+func (r *fieldExportReconciler) BindServiceResourceManager(mgr ctrlrt.Manager) error {
```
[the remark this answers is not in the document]
RedbackThomson: I love that, yes!

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+		Namespace: req.Namespace,
+	}
+
+	// Check if the target API group matches with the controller
+	var controllerRMF acktypes.AWSResourceManagerFactory
+	for _, v := range r.sc.GetResourceManagerFactories() {
+		controllerRMF = v
+		break
+	}
+	if sourceGK.Group != controllerRMF.ResourceDescriptor().GroupKind().Group {
```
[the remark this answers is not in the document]
RedbackThomson: This only fails if `GetResourceManagerFactories` returns an empty list. In which case we have a controller without any managers haha.

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+
+	// Attempt an initial export
+	return r.Sync(ctx, sourceObject, *feObject)
+}
+
+// reconcileSourceResource handles updates to any other (not `FieldExport`) ACK
+// resources
+func (r *fieldExportReconciler) reconcileSourceResource(ctx context.Context, req ctrlrt.Request) error {
+	res, err := r.getSourceResource(ctx, *r.rd, req.NamespacedName)
+	if err != nil {
```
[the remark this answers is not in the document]
RedbackThomson: Great catch. It was actually raising reconciler errors when delete any resources.

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+	for _, v := range r.sc.GetResourceManagerFactories() {
+		controllerRMF = v
+		break
+	}
+	if sourceGK.Group != controllerRMF.ResourceDescriptor().GroupKind().Group {
+		ackrtlog.DebugFieldExport(r.log, feObject, "target resource API group is not of this service. no-op")
+		return nil
+	}
+
+	if feObject.DeletionTimestamp != nil {
```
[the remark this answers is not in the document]
RedbackThomson: I specifically put it after the group check because I only want the controller that manages FieldExports for this type to be able to clean it up. I realise now that someone could change the spec and the service identifier in the finalizer would not match the service in the spec. Probably need to rethink this logic a little bit.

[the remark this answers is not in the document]
RedbackThomson: Okay so I've just added a check inside `markUnmanaged` that only attempts to remove the finalizer if it's detected (that is, the field export has previously been managed by the current controller). Then I moved this line to above the sourceGK check, so every reconciler will do this check. Essentially all controllers will see the deletion timestamp and attempt to remove their respective finalizers if they see it.

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,674 @@
...
+	path string,
+) (*string, error) {
+	obj, err := UnstructuredConverter.ToUnstructured(from.RuntimeObject())
+	if err != nil {
+		return nil, err
+	}
+
+	query, err := jq.Parse(path)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to parse path")
```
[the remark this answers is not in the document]
RedbackThomson: This error message will eventually be placed into the `FieldExport` conditions. That information is already available in the spec. For the controller logs, they go through `AdaptFieldExport` and have that information included in that, too.

pkg/runtime/field_export_reconciler.go
RedbackThomson: The reason why I didn't: I see conditions as actions taken by the controller on the given object, or the state of that object itself. The field export reconciler takes an action on the configmap, and simply uses the fieldexport object as the configuration. 

That being said, I'm happy to add this in a future revision if people want to see the last time it was synced, for instance.

pkg/runtime/field_export_reconciler.go
RedbackThomson: Oh wow I had this logic backwards. If there is an error, set `Recoverable` to `true` and set the message to that error message. If there is not error, set `Recoverable` to `false` and remove any message.

pkg/runtime/field_export_reconciler.go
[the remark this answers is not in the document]
jaypipes: > No, in fact I don't want to ever requeue the `FieldExport` resource unless it explicitly throws an error. If it's in a recoverable condition, either:
> 
> * the user needs to manually patch it to bring it back to working order

This is the behaviour for `ACK.Terminal`, which means "unless you change the desired spec, this resource won't sync". `ACK.Recoverable` on the other hand means "this resource may sync in a future reconciliation loop without the user changing anything about the desired state"

pkg/runtime/field_export_reconciler.go
vijtrip2: Yeah. For FieldExport resource, it will be useful to  
* easily find the last Synced time during debugging
* Knowing whether FieldExport is still in process of being reconciled if application fails to start

pkg/runtime/field_export_reconciler.go
vijtrip2: >If there is not error, set Recoverable to false and remove any message.

Why not clear the recoverable condition in this case?

pkg/runtime/field_export_reconciler.go
```diff
@@ -60,18 +59,22 @@ var (
 // It implements the upstream controller-runtime `Reconciler` interface.
 type fieldExportReconciler struct {
 	reconciler
+	// resourceReconciler indicates that the reconciler should be reconciling
+	// AWS resources
+	resourceReconciler bool
 	// rd is only used if binding to an ACK resource (not `FieldExport`)
-	rd *acktypes.AWSResourceDescriptor
+	rd acktypes.AWSResourceDescriptor
```
vijtrip2, line 66: zero value for the interface is nil. 
I do not think you need the "resourceReconciler" boolean. You can just use nil check for rd

pkg/runtime/field_export_reconciler.go
```diff
@@ -299,7 +301,7 @@ func (r *fieldExportReconciler) getSourcePathFromResource(
 ) (*string, error) {
 	obj, err := UnstructuredConverter.ToUnstructured(from.RuntimeObject())
 	if err != nil {
-		return nil, err
+		return nil, nil
```
vijtrip2, line 304: if the error is being ignored, logging before ignoring will be a better practice.

RedbackThomson: Ah crap no this is a mistake I made when trying to change error handling elsewhere.

pkg/runtime/field_export_reconciler.go
RedbackThomson: Okay. I might add a follow-up PR with these changes, because I think we are already pretty close with this one.

pkg/runtime/field_export_reconciler.go
RedbackThomson: We won't ever reach the code path where `err == nil` in this reconciler anyway. We only call this method when we have an error. I'll just remove that branch so it's not confusing.

pkg/runtime/field_export_reconciler.go
RedbackThomson: When I tested that it wasn't really working. It was being instantiated as `acktypes.AWSResourceDescriptor{}` (the empty object). The nil check was failing because of that. I'm not very familiar with nillable interfaces in Go

pkg/runtime/field_export_reconciler.go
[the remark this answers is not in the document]
RedbackThomson: I see! Okay I will make some errors into `Terminal`, then.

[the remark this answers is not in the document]
RedbackThomson: Hmm there's actually only one place where the error would be `Terminal` - if the `spec.path` cannot be parsed by `gojq`. Anything else could be considered `Recoverable`:
- If the path is not found on the object, it's possible that the object just hasn't added that status or spec yet
- If it cannot find the output configmap/secret, it's possible the output resource hasn't been created yet

pkg/runtime/field_export_reconciler.go
vijtrip2: I just tested by creating a small struct and the value of AWSResourceDescriptor was nil when not initialized in the struct. Can you please double check your implementation again? Maybe try using `ackCompare.IsNil` function.

I used following test to verify the nil check.

```
type mystruct struct {
	mystring string
	resDes   acktypes.AWSResourceDescriptor
}

func TestMyStruct(t *testing.T) {
	myS := mystruct{mystring: "myString"}
	if myS.resDes == nil {
		fmt.Println("resDes is nil")
	} else {
		fmt.Println("resDes is not nil")
	}
	fmt.Println(myS.mystring)
}
```

pkg/types/field_export_reconciler.go
```diff
@@ -33,7 +33,7 @@ type FieldExportReconciler interface {
 		context.Context,
 		AWSResource,
 		ackv1alpha1.FieldExport,
-	) error
+	) (ackv1alpha1.FieldExport, error)
```
vijtrip2, line 36: curious: was this return type updated just for the unit testing?

vijtrip2 (approved):

pkg/types/field_export_reconciler.go
RedbackThomson: Yeah, to be in line with other reconcilers, for unit testing and since we now maintain a "desired" and a "latest" - we are not directly editing the passed argument.

pkg/runtime/field_export_reconciler.go
```diff
@@ -216,31 +218,41 @@ func (r *fieldExportReconciler) Sync(
...
-	r.clearConditions(ctx, &desired)
+) (ackv1alpha1.FieldExport, error) {
+	var err error
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("r.Sync")
+	defer exit(err)
+
+	latest := desired.DeepCopy()
+
+	r.resetConditions(ctx, &desired)
```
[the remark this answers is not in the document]
RedbackThomson: `desired.DeepCopy()` returns a pointer, actually:
```go
func (*ackv1alpha1.FieldExport).DeepCopy() *ackv1alpha1.FieldExport
```

pkg/runtime/field_export_reconciler.go
```diff
@@ -581,7 +592,22 @@ func (r *fieldExportReconciler) patchTerminalCondition(
 		terminalCondition.Message = &errMessage
 	}
 
-	return r.patchStatus(ctx, res, base)
+	return nil
+}
+
+// ensureConditions examines the supplied resource's collection of Condition
+// objects and ensures that an ACK.ResourceSynced condition is present.
+func (r *fieldExportReconciler) ensureConditions(
```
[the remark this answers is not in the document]
RedbackThomson: Oh by `ensure` I meant "ensure that they're present in the status of the object". I think you're right though, apart from logging, there is no need for this to be a separate method.

pkg/runtime/field_export_reconciler.go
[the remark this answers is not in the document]
vijtrip2: gotcha. I think implicit conversion is happening when you call desired.DeepCopy method.

vijtrip2 (approved):

apis/core/v1alpha1/identifiers.go
```diff
@@ -26,14 +30,34 @@ type AWSIdentifiers struct {
...
+// ResourceFieldSelector provides the values necessary to identify an individual
+// path on an individual K8s resource.
+type ResourceFieldSelector struct {
+	Resource NamespacedTargetKubernetesResource `json:"resource"`
+	Path     *string                            `json:"path"`
+}
+
+// FieldExportOutputSelector provides the values necessary to identify the
+// output path for a field export.
+type FieldExportOutputSelector struct {
```
jaypipes, line 56: Now, this particular struct *is* specific to a FieldExport struct, so I would move it to the `field_export.go` file and I'd recommend renaming it `FieldExportTarget` because that's what it describes.

pkg/requeue/requeue.go
```diff
@@ -44,6 +52,31 @@ func NeededAfter(
 	}
 }
 
+// An error to instruct the ACK runtime to process an error, but not requeue the
```
jaypipes, line 55: docstring should begin "// NoRequeue is a struct that contains an error to instruct.."

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,785 @@
...
+
+const (
+	fieldExportFinalizerString = "finalizers.services.k8s.aws/FieldExport"
+)
+
+var (
+	pathDoesNotExistError = errors.New("path does not exist in this object")
+)
+
+// Remove the global to make it easier to mock
```
jaypipes, line 52: Remove the global?

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,785 @@
...
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
+	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+)
+
+const (
+	fieldExportFinalizerString = "finalizers.services.k8s.aws/FieldExport"
+)
+
+var (
+	pathDoesNotExistError = errors.New("path does not exist in this object")
```
jaypipes, line 49: Did you want to place this in [`pkg/errors`](https://github.com/aws-controllers-k8s/runtime/tree/main/pkg/errors) and export it?

pkg/runtime/service_controller.go
```diff
@@ -92,9 +100,9 @@ func (c *serviceController) GetResourceManagerFactories() map[string]acktypes.AW
 	return c.rmFactories
 }
 
-// GetAdoptedResourceInstalled returns whether the AdoptedResource CRD has been
-// installed into the cluster, and is accessible by the service controller.
-func (c *serviceController) GetAdoptedResourceInstalled(mgr ctrlrt.Manager) (bool, error) {
+// GetAdoptedResourceInstalled returns whether the given resource plural has
```
jaypipes, line 103: need to match the docstring with the new name of this method.

pkg/types/field_export_reconciler.go
```diff
@@ -0,0 +1,54 @@
...
+	// to the controller manager
+	BindControllerManagerForAWSResource(
+		ctrlrt.Manager,
+	) error
+	// FilterAllExports will list all FieldExport CRs and filter them based on
+	// whether they contain a reference to the given AWS resource.
+	FilterAllExports(
+		context.Context,
+		metav1.GroupKind,
+		types.NamespacedName,
```
jaypipes, line 52: Instead of separate GK and NamespacedName parameters, how about using the new `FieldExportOutputSelector` (that I think you should rename to `FieldExportTarget` type instead?

jaypipes (changes requested):
Getting there, @RedbackThomson thank you! Left a number of suggestions inline for ya.

pkg/types/field_export_reconciler.go
```diff
@@ -0,0 +1,54 @@
...
+		ctrlrt.Manager,
+	) error
+	// BindControllerManagerForAWSResource binds a given AWS resource descriptor
+	// to the controller manager
+	BindControllerManagerForAWSResource(
+		ctrlrt.Manager,
+	) error
+	// FilterAllExports will list all FieldExport CRs and filter them based on
+	// whether they contain a reference to the given AWS resource.
+	FilterAllExports(
```
jaypipes, line 49: Can we name this `GetFieldExportsForResource`?

apis/core/v1alpha1/adopted_resource.go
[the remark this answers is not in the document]
RedbackThomson: I reached out to Ellis about this:
> [Nick] I feel that if the CRD is identical, there shouldn't be a difference?
> [Ellis] The API Server serializes it using the CRD
> [Ellis] Exactly
> [Ellis] The APIServer isn't aware of your go struct, just the CRD
> [Ellis] We do this regularly in karpenter w/ our go types

So we don't need to worry about etcd serialization. Just that we are maintaining the same contract with the API server, which is the CRD. I have confirmed that the CRD remains untouched.

pkg/requeue/requeue.go
RedbackThomson: Copied from the error docstrings below, but I'll fix all of them 👍🏼

pkg/runtime/field_export_reconciler.go
RedbackThomson: There are a number of error strings in this package. I see resource references have their own error file. Should I move everything into one like that?

pkg/types/field_export_reconciler.go
RedbackThomson: Unfortunately can't re-use that since the `Kind` it's referring to in `FieldExportTarget` is the `"configmap"|"secret"` enum, it's not any ACK type.

jaypipes: ah, yeah.. never mind then! :)

pkg/runtime/field_export_reconciler.go
jaypipes: Up to you! Feel free to create a field_export.go file in pkg/errors or add to the errors.go file. Whatever you feel is best.

pkg/errors/field_export.go
```diff
@@ -0,0 +1,39 @@
...
+	// FieldExportInvalidPath indicates there was an error when executing the
+	// JQ query
+	FieldExportQueryFailed = fmt.Errorf("unable to execute query")
+	// FieldExportMissingConfigMap indicates there was an error when trying
+	// to get the target configmap
+	FieldExportMissingConfigMap = fmt.Errorf("unable to get existing configmap")
+	// FieldExportMissingSecret indicates there was an error when trying
+	// to get the target secret
+	FieldExportMissingSecret = fmt.Errorf("unable to get existing secret")
+)
```
jaypipes, line 39: :+1:

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,777 @@
...
+	if feObject.DeletionTimestamp != nil {
+		return r.cleanup(ctx, feObject)
+	}
+
+	// Check if the target API group matches with the controller
+	var controllerRMF acktypes.AWSResourceManagerFactory
+	for _, v := range r.sc.GetResourceManagerFactories() {
+		controllerRMF = v
+		break
+	}
```
jaypipes, line 145: Why do we need to do this when `fieldExportReconciler` contains an `rd` field which is the AWSResourceDescriptor that is the same as what is returned by `controllerRMF.ResourceDescriptor()` in the line below?

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,777 @@
...
+		return err
+	}
+
+	// Look up the rmf for the given target resource GVK
+	rmf, ok := (r.sc.GetResourceManagerFactories())[sourceGK.String()]
+	if !ok {
+		return ackerr.ResourceManagerFactoryNotFound
+	}
+
+	sourceObject, err := r.getSourceResource(ctx, rmf.ResourceDescriptor(), sourceName)
```
jaypipes, line 161: It seems to me that because the controller's Role or ClusterRole is the thing that is reading the source resource, this would be an easy way for a Kubernetes user to subvert Kubernetes RBAC. If a Kubernetes user only has read and write permissions on the `FieldExport` Kind and does not have read permissions on the source resource Kind, the functionality here essentially allows them to read any information (by specifying a path on that CRD) on that source resource Kind.

pkg/runtime/field_export_reconciler.go
RedbackThomson: > `fieldExportReconciler` contains an `rd` field

This is only true if we have bound the manager to an ACK resource. If we are binding to `FieldExport` then we leave that property as nil. In that case, this logic is the same as the adoption reconciler, where we simply fetch the first resource manager and use it to find the group that the controller is in charge of.

jaypipes (changes requested):
@RedbackThomson the whole `fieldExportReconciler` single reconciler struct for both source resource and FieldExport resources is still rubbing me the wrong way. Please see my comment inline about having two reconciler structs, one for FieldExports and another for source resources.

pkg/runtime/field_export_reconciler.go
```diff
@@ -0,0 +1,777 @@
...
+	UnstructuredConverter runtime.UnstructuredConverter = runtime.DefaultUnstructuredConverter
+)
+
+// fieldExportReconciler is responsible for reconciling the state of any field
+// export CRs that target resources supported by the current controller.
+// It implements the upstream controller-runtime `Reconciler` interface.
+type fieldExportReconciler struct {
+	reconciler
+	// rd is only used if binding to an ACK resource (not `FieldExport`)
+	rd acktypes.AWSResourceDescriptor
```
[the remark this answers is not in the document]
RedbackThomson: Great idea. I will admit I still don't fully understand how composition works in Go, and didn't know if this was possible. But my initial tests seem to show this works really well.

jaypipes (approved):
There's enough good here that I'm willing to address a few things about the fieldExportReconciler struct in followup PRs. :+1:

acornett21:
@RedbackThomson @jaypipes Sorry I'm late on this one, but is this a CRD that would be packaged with every controller just like the `AdoptedResource` correct? If so then mitigation work in the runtime and in the `code-generator` should be accounted for to not package this CRD. If I'm miss understanding this apologies.

RedbackThomson:
> @RedbackThomson @jaypipes Sorry I'm late on this one, but is this a CRD that would be packaged with every controller just like the `AdoptedResource` correct? If so then mitigation work in the runtime and in the `code-generator` should be accounted for to not package this CRD. If I'm miss understanding this apologies.

Correct. This CRD is handled exactly in the same way as `AdoptedResource`. The mitigation work we put in for OLM to not package it, I think we did that for the common CRD directory? Hopefully that code already removes this new CRD as well.

acornett21:
@RedbackThomson Looking again at this PR and the code we have [here](https://github.com/aws-controllers-k8s/code-generator/blob/main/scripts/olm-create-bundle.sh#L107-L108) I think adding this would be covered by whats in place. Can you please take a look and confirm? Thanks!

RedbackThomson:
> @RedbackThomson Looking again at this PR and the code we have [here](https://github.com/aws-controllers-k8s/code-generator/blob/main/scripts/olm-create-bundle.sh#L107-L108) I think adding this would be covered by whats in place. Can you please take a look and confirm? Thanks!

Exactly! I tried to future proof it, and I think this still holds.

acornett21:
Yeah, It looks that way, missed it on the first pass of looking at this PR.

PR #79, opened 2022-03-21
Add `Region()` to `ResourceMetadata` struct and `AWSResourceIdentifiers`

A-Hilaly:
interface

Currently when a resource is created without setting a region in the
annotations, the controller uses either the namespace annotation or the
controller region flag. This causes the controller to miss manage the
resource if they are restart with a different region, or if run a second
controller using a different region flag.

This patch adds a new field to `ResourceMetadata` struct and
`AWSResourceIdentifiers` interface to store the AWS region in which
the resource is created/should be managed. This patch also updates the
`reconciler.getRegion` method to lookup for a region in the
status.ResourceMetadata before trying to get one from the
resource/namespace annotations or the controller configuration.

Addresses https://github.com/aws-controllers-k8s/community/issues/1220

This patch will also require some changes to the code-generator.


apis/core/v1alpha1/resource_metadata.go
```diff
@@ -30,4 +30,7 @@ type ResourceMetadata struct {
 	// OwnerAccountID is the AWS Account ID of the account that owns the
 	// backend AWS service API resource.
 	OwnerAccountID *AWSAccountID `json:"ownerAccountID"`
+	// AWSRegion is the AWS region in which the resource is expected to be
+	// managed.
+	AWSRegion *AWSRegion
```
jaypipes: Let's just call this field `Region`, since it is assumed to be the AWS region...

apis/core/v1alpha1/resource_metadata.go
```diff
@@ -30,4 +30,7 @@ type ResourceMetadata struct {
 	// OwnerAccountID is the AWS Account ID of the account that owns the
 	// backend AWS service API resource.
 	OwnerAccountID *AWSAccountID `json:"ownerAccountID"`
+	// AWSRegion is the AWS region in which the resource is expected to be
+	// managed.
```
jaypipes: instead of "which the resource is expected to be managed", this should read "which the resource exists". `Status.ACKResourceMetadata.Region` should only be set when a successful `Create` operation has occurred.

pkg/runtime/reconciler.go
```diff
@@ -827,14 +827,20 @@ func (r *resourceReconciler) getRoleARN(
...
+// If the CR have a region associated with it, it is used. Otherwise
 // we look for the namespace associated region, if that is set we use it. Finally
-// if none of these annotations are set we use the use the region specified in the
-// configuration is used
+// if none of these annotations are set we use the region specified in the
+// controller configuration.
 func (r *resourceReconciler) getRegion(
 	res acktypes.AWSResource,
 ) ackv1alpha1.AWSRegion {
+	// first try to get the owner account id from the status.resourceMetadata
```
jaypipes, line 838: s/owner account id/region/

pkg/runtime/reconciler.go
```diff
@@ -827,14 +827,20 @@ func (r *resourceReconciler) getRoleARN(
...
 
-// getRegion returns the AWS region that the given resource is in or should be
-// created in. If the CR have a region associated with it, it is used. Otherwise
+// getRegion returns the AWS region where the resource should be managed.
+// If the CR have a region associated with it, it is used. Otherwise
 // we look for the namespace associated region, if that is set we use it. Finally
-// if none of these annotations are set we use the use the region specified in the
-// configuration is used
+// if none of these annotations are set we use the region specified in the
+// controller configuration.
```
jaypipes, line 834: I would reword this docstring to say something like this:

```
// getRegion returns the region the resource exists in, or if the resource
// has yet to be created, the region the resource *should* be created in.
//
// If the resource has not yet been created, we look for the AWS region
// in the following order of precedence:
//  - The resource's `services.k8s.aws/region` annotation, if present
//  - The resource's Namespace's `services.k8s.aws/region` annotation, if present
//  - The controller's `--aws-region` CLI flag
```

pkg/types/aws_resource_identifiers.go
```diff
@@ -27,4 +27,7 @@ type AWSResourceIdentifiers interface {
 	// this means the resource has not yet been created in the backend AWS
 	// service.
 	ARN() *ackv1alpha1.AWSResourceName
+	// AWSRegion is the AWS region in which the resource is expected to be
+	// managed.
```
jaypipes, line 31: s/AWSRegion/Region/

s/is expected to be managed/exists or will exist/

jaypipes (changes requested):
@A-Hilaly thank you! a few minor comments about naming/wording, nothing major :)

apis/core/v1alpha1/resource_metadata.go
A-Hilaly: Changed few seconds after the first push :)

jaypipes (approved):
:+1: thanks Amine!

PR #80, opened 2022-03-23
Resolve references again only when resource is patched with finalizer

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1234

Description of changes:

* Keeps existing functionality but adds a condition to check for finalizer before adding the finalizer and resolving the references again.
* If the resource is not getting patched again, there is no need for resolving the references again
* Adds Unit test cases for Create behavior of both managed and unmanaged resources

----

* Original issue [#1187](https://github.com/aws-controllers-k8s/community/issues/1187) explaining why references are resolved again during resource creation.


jaypipes (approved):
Tricky one. Thanks for the fix, @vijtrip2.

PR #81, opened 2022-03-30
Add a retry mechanism for `ReadOne` calls in `rm.createResource`

A-Hilaly:
Issue #, if available:

Description of changes:

In some rare cases and with some specific AWS APIs, calling `ReadOne`
right after a `rm.Create` can return a `NotFound` error. We want to
retry calling `rm.ReadOne` in hopes of receiving a correct response.

This patch adds a backoff/retry mechanism around `ReadOne` call in
`rm.createResource`.

### DIFF
--- a/ATTRIBUTION.md
+++ b/ATTRIBUTION.md
@@ -38,6 +38,7 @@ License version 2.0, we include the full text of the package's License below.
 * `k8s.io/client-go`
 * `sigs.k8s.io/controller-runtime`
 * `sigs.k8s.io/controller-tools`
+* `github.com/cenkalti/backoff`
 
 ### github.com/aws/aws-sdk-go
 
@@ -1388,3 +1389,29 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+# github.com/cenkalti/backoff
+
+The MIT License (MIT)
+
+Copyright (c) 2014 Cenk Altı
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+Subdependencies: N/A
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -15,8 +15,10 @@ package runtime
 
 import (
 	"context"
+	"fmt"
 	"time"
 
+	backoff "github.com/cenkalti/backoff/v4"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -37,6 +39,10 @@ import (
 	ackrtcache "github.com/aws-controllers-k8s/runtime/pkg/runtime/cache"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+)
+
+const (
+	backoffReadOneTimeout = 10 * time.Second
 )
 
 // reconciler describes a generic reconciler within ACK.
@@ -380,7 +386,21 @@ func (r *resourceReconciler) createResource(
 	observed, err := rm.ReadOne(ctx, latest)
 	rlog.Exit("rm.ReadOne", err)
 	if err != nil {
-		return latest, err
+		if err == ackerr.NotFound {
+			// Some eventually-consistent APIs return a 404 from a
+			// ReadOne operation immediately after a successful
+			// Create operation. In these exceptional cases
+			// we retry the ReadOne operation with a backoff
+			// until we get the expected 200 from the ReadOne.
+			rlog.Enter("rm.delayedReadOneAfterCreate")
+			observed, err = r.delayedReadOneAfterCreate(ctx, rm, latest)
+			rlog.Exit("rm.delayedReadOneAfterCreate", err)
+			if err != nil {
+				return latest, err
+			}
+		} else {
+			return latest, err
+		}
 	}
 
 	// Take the status from the latest ReadOne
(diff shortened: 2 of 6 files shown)


vijtrip2 (commented):
Can you please also add some unit tests for this by mocking the ReadOne call and verifying that right number of ReadOne calls are being made in unit test assertions.

---------

Not against this solution, since it leaves the reconciler behavior as is for services that do not suffer from this ReadOne(after Create) behavior. However i would love to hear your thoughts on following,

a) Do you think it is worth to make this retry-backoff behavior configurable for service teams in generator.yaml? (Have default values when not specified)

b) Why was existing solution of requeueing and complete reconciler loop retry not sufficient enough? (Just add in the PR description)

pkg/runtime/reconciler.go
```diff
@@ -370,10 +376,29 @@ func (r *resourceReconciler) createResource(
 		return latest, err
 	}
 
-	rlog.Enter("rm.ReadOne")
-	observed, err := rm.ReadOne(ctx, latest)
-	rlog.Exit("rm.ReadOne", err)
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxElapsedTime = backoffReadOneTimeout
```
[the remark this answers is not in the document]
A-Hilaly: >Does this library automatically stop and the loop breaks when backoff time reaches max backoff seconds?

Yes exactly. I can also put a MaxRetries i guess 5 should be Okey in this case?

pkg/runtime/reconciler.go
```diff
@@ -370,10 +376,29 @@ func (r *resourceReconciler) createResource(
...
+	ticker := backoff.NewTicker(bo)
+	attempts := 0
+
+	var observed acktypes.AWSResource
+
+	for range ticker.C {
+		rlog.Enter(fmt.Sprintf("rm.ReadOne (attempt %d)", attempts+1))
+		observed, err = rm.ReadOne(ctx, latest)
+		rlog.Exit("rm.ReadOne (attempt %d)", err)
+		if err == ackerr.NotFound {
```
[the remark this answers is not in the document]
A-Hilaly: Yes, I purposefully separated them. Because I thought that maybe I should add a special log when an API returns `NotFound`.

A-Hilaly:
awaiting for soak tests results with ECR
/hold

A-Hilaly:
> a) Do you think it is worth to make this retry-backoff behavior configurable for service teams in generator.yaml? (Have default values when not specified)

@vijtrip2 Yes i do believe that it can be useful for some portion of the controllers at least. Were thinking about adding a flag to make this retry/backoff configurable?

pkg/runtime/reconciler.go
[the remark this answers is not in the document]
A-Hilaly: Squashed the err statements

ATTRIBUTION.md
```diff
@@ -1388,3 +1389,29 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+Subdependencies: N/A
```
jaypipes, line 1417: :+1:

pkg/runtime/reconciler.go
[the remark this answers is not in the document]
jaypipes: @vijtrip2 I don't think there's a need to *both* a max timeout AND a number of retries... the way the backoff library works is that if a timeout is configured (as is the case here) the ticker stops. I think it adds unnecessary code to add the max retries in addition to the timeout.

pkg/runtime/reconciler.go
```diff
@@ -376,10 +383,30 @@ func (r *resourceReconciler) createResource(
...
+		rlog.Exit(fmt.Sprintf("rm.ReadOne (attempt %d)", attempts+1), err)
+		if err != nil {
+			if attempts+1 > backoffReadOneMaxRetries {
+				ticker.Stop()
+				break
+			}
+			continue
+		}
+		ticker.Stop()
+		break
```
jaypipes, line 405: We need to handle non-404 errors differently than 404s. Remove the max retries stuff and update this block to this instead:

```go
if err == nil {
    break
}
if err != ackerr.NotFound {
    return latest, nil
}
```

jaypipes: 1) Place all of the retry w/ backoff loop into a separate helper function called `delayedReadOneAfterCreate()`.

2) Use the `rlog.Trace()` functionality to mark the entrance and exit to this helper function.

3) The `delayedReadOneAfterCreate` helper function represents an exceptional edge case and as such we should only call the `delayedReadOneAfterCreate` helper function if the original code's `ReadOne` call returns an `err` that is not `nil` and is a NotFound.

In other words, the original code was this:

```go
	rlog.Enter("rm.Create")
	rlog.Exit("rm.Create", err)
	if err != nil {
		return latest, err
	}

	rlog.Enter("rm.ReadOne")
	observed, err := rm.ReadOne(ctx, latest)
	rlog.Exit("rm.ReadOne", err)
	if err != nil {
		return latest, err
	}
```

and I'd like to see it look like this:

```go
	rlog.Enter("rm.Create")
	rlog.Exit("rm.Create", err)
	if err != nil {
		return latest, err
	}

	rlog.Enter("rm.ReadOne")
	observed, err := rm.ReadOne(ctx, latest)
	rlog.Exit("rm.ReadOne", err)
	if err != nil {
		if err == ackerr.NotFound {
			// Some eventually-consistent APIs return a 404 from a
			// ReadOne operation immediately after a successful
			// Create operation. In these exceptional cases
			// we retry the ReadOne operation with a backoff
			// until we get the expected 200 from the ReadOne.
			observed, err = r.delayedReadOneAfterCreate(ctx, latest)
			if err != nil {
				return latest, err
			}
		} else {
			return latest, err
		}
	}
```

jaypipes (changes requested):
Thanks @A-Hilaly! Good start on this :) Left some comments about changes I'd like to see..

pkg/runtime/reconciler.go
RedbackThomson: In the case of a server 5xx error, do we want to immediately return that? I could imagine a hiccup in the service (or maybe a throttling exception) could cause our backoff to false negative and spit out an error?

A-Hilaly: I like the idea of the `delayedReadOneAfterCreate`. Changed the code to adopt that.

A-Hilaly: > In the case of a server 5xx error, do we want to immediately return that? I could imagine a hiccup in the service (or maybe a throttling exception) could cause our backoff to false negative and spit out an error?

I think it's really hard to generalize the code to be able to handle 5XX errors across all the AWS APIs. Also 5xx errors can occur at any CRUD operation

pkg/runtime/reconciler.go
```diff
@@ -39,6 +41,11 @@ import (
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
 )
 
+const (
+	backoffReadOneTimeout    = 10 * time.Second
+	backoffReadOneMaxRetries = 5
```
vijtrip2, line 46: remove the unused constant.?

pkg/runtime/reconciler.go
```diff
@@ -397,6 +418,46 @@ func (r *resourceReconciler) createResource(
...
+
+		rlog.Enter(fmt.Sprintf("rm.ReadOne (attempt %d)", attempts))
+		observed, err = rm.ReadOne(ctx, res)
+		rlog.Exit(fmt.Sprintf("rm.ReadOne (attempt %d)", attempts), err)
+		if err == nil {
+			ticker.Stop()
+			break
+		}
+		if err != ackerr.NotFound {
+			return res, nil
```
vijtrip2, line 452: ?
```suggestion
			return res, err
```

pkg/runtime/reconciler.go
```diff
@@ -397,6 +418,46 @@ func (r *resourceReconciler) createResource(
...
+		}
```
vijtrip2, line 453:
```suggestion
		if err == nil || err != ackerr.NotFound {
			ticker.Stop()
			break
		}
```

pkg/runtime/reconciler_test.go
```diff
@@ -139,6 +139,58 @@ func managerFactoryMocks(
...
+	).Return()
+
+	rm := &ackmocks.AWSResourceManager{}
+	rm.On("ResolveReferences", ctx, nil, desired).Return(
+		desired, nil,
+	).Times(2)
+	rm.On("ReadOne", ctx, desired).Return(
+		latest, ackerr.NotFound,
+	).Once()
+	rm.On("ReadOne", ctx, latest).Return(
```
vijtrip2, line 170: 👍

PR #82, opened 2022-04-05
Add `New` method to `TerminalError`

nmvk:
This method helps to throw terminal error from the controller.

Controller code can throw errors like

```
	if *latestStatus == "create-failed" {
		return nil, ackerr.TerminalError.New(errors.New("cluster cannot be updated as its status is not 'available'."))
	}
```

Code generator changes - https://github.com/aws-controllers-k8s/code-generator/pull/318

### DIFF
--- a/pkg/errors/error.go
+++ b/pkg/errors/error.go
@@ -92,6 +92,10 @@ type TerminalError struct {
 	err error
 }
 
+func (e TerminalError) New(terminalError error) *TerminalError {
+	return &TerminalError{err: terminalError}
+}
+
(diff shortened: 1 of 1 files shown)


RedbackThomson:
I don't quite understand. Is this any different from `&TerminalError{err: fmt.Errorf("my error message")}`?

nmvk:
> I don't quite understand. Is this any different from `&TerminalError{err: fmt.Errorf("my error message")}`?

This would throw `Unexported field 'err' usage`  since we want to throw the error outside of the package?

PR #84, opened 2022-04-22
Add `SecretKeyReference` comparison functions to `pkg/compare`

A-Hilaly:
Issue https://github.com/aws-controllers-k8s/community/issues/1239

Currently, the code generator doesn't correctly generate delta comparison
functions for secret fields, because the runtime doesn't contain any
helper functions that can help in computing the delta of
`SecretKeyReference` and `SecretKeyReference` slices.

This patch adds two functions to help compare two instances or slices
of `SecretKeyReference`.

### DIFF
--- a/pkg/compare/secret_reference.go
+++ b/pkg/compare/secret_reference.go
@@ -1,0 +1,104 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package compare
+
+import (
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+)
+
+// SecretKeyReferenceEqual returns true if the supplied secret key references
+// are equal.
+func SecretKeyReferenceEqual(a, b *ackv1alpha1.SecretKeyReference) bool {
+	if HasNilDifference(a, b) {
+		return false
+	}
+	return a.Name == b.Name &&
+		a.Namespace == b.Namespace &&
+		a.Key == b.Key
+}
+
+// SliceSecretKeyReferenceEqual returns true if the supplied slices of secret
+// key reference pointers contain the exact same elements.
+func SliceSecretKeyReferenceEqual(
+	a, b []*ackv1alpha1.SecretKeyReference,
+) bool {
+	equal, _, _ := CompareSecretKeyReferences(a, b)
+	return equal
+}
+
+// CompareSecretKeyReference returns true if the supplied slices of secret
+// key reference pointers contain the exact same elements. It will also return 2
+// slices containing elements contained in a that aren't in b, and elements
+// contained in b that aren't in a, respectively. The comparison doesn't
+// take into consideration the order of elements.
+// Duplicated elements will not impact the behaviour of this function, meaning
+// that the function will always return only unique instances of the added/removed
+// elements.
+func CompareSecretKeyReferences(
+	a, b []*ackv1alpha1.SecretKeyReference,
+) (equal bool, added, removed []*ackv1alpha1.SecretKeyReference) {
+	// finding the removed elements
+	for _, aVal := range a {
+		found := false
+		for _, bVal := range b {
+			if SecretKeyReferenceEqual(aVal, bVal) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			removed = append(removed, aVal)
+			continue
+		}
+	}
+	// finding the added elements
+	for _, bVal := range b {
+		found := false
+		for _, aVal := range a {
+			if SecretKeyReferenceEqual(aVal, bVal) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			added = append(added, bVal)
+		}
+	}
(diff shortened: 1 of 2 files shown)


pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,95 @@
...
+// SliceStringPEqual returns true if the supplied slices of secret key reference
+// pointers contain the exact same elements. It will also return 2 slices containing
+// the added and removed elements. The comparison doesn't take into consideration
+// the order of elements.
+// Duplicated elements will not impact the behaviour of this function, meaning that
+// the function will always return only unique instances of the added/removed elements.
+//
+// TODO(a-hilaly): Ideally we will need to return another array containing secret
+// that got updated. To do that we will need to pass a k8s client to the newResourceDelta
+// function and use it here to retrieve/compare the secret values.
```
jaypipes, line 40: I actually don't want to compare the values of the secrets. Just the keys/namespaces I think is good.

pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,95 @@
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			removed = append(removed, aVal)
+			continue
+		}
+	}
+	// finding the deleted elements
```
jaypipes, line 58: s/deleted/added/

pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	if len(added) == 0 && len(removed) == 0 {
+		equal = true
+	}
+	return equal, cleanUpDuplicateSecretReferences(added), cleanUpDuplicateSecretReferences(removed)
+}
+
+// cleanUpDuplicateSecretReferences removes duplicate elements from a given slice
+// of secret key references.
+func cleanUpDuplicateSecretReferences(a []*ackv1alpha1.SecretKeyReference) (uniqueSecretReferences []*ackv1alpha1.SecretKeyReference) {
```
jaypipes, line 79: for long function (>79 chars) signatures like this, please use multiple lines, like so:

```go
func cleanUpDuplicateSecretReferences(
    a []*ackv1alpha1.SecretKeyReference,
) (uniqueSecretReferences []*ackv1alpha1.SecretKeyReference) {
```

pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	return a.Name == b.Name &&
+		a.Namespace == b.Namespace &&
+		a.Key == b.Key
+}
+
+// SliceStringPEqual returns true if the supplied slices of secret key reference
+// pointers contain the exact same elements. It will also return 2 slices containing
+// the added and removed elements. The comparison doesn't take into consideration
+// the order of elements.
```
jaypipes, line 34: @A-Hilaly none of the `XXXEqual` functions in `pkg/compare` return the elements that are different. They only return a boolean to indicate that there is a difference. I would like to keep that consistent for secret key reference types, please.

jaypipes (changes requested):
@A-Hilaly very cool! I have a few changes to request inline but this is a great start!

pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,95 @@
...
+func SecretKeyReferenceEqual(a, b *ackv1alpha1.SecretKeyReference) bool {
+	if HasNilDifference(a, b) {
+		return false
+	}
+	return a.Name == b.Name &&
+		a.Namespace == b.Namespace &&
+		a.Key == b.Key
+}
+
+// SliceStringPEqual returns true if the supplied slices of secret key reference
```
jaypipes, line 31: Just noticed the above should be `SliceSecretKeyReferenceEqual` instead of `SliceStringPEqual`

jaypipes (commented):
How are we doing on this one, @A-Hilaly?

pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,103 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
RedbackThomson, line 1: nit: Rename this file. Maybe we could move it under `pkg/compare/secrets/reference.go`? Then when we do value comparison that can go under `value.go`?

pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,103 @@
...
+func SliceSecretKeyReferenceEqual(
+	a, b []*ackv1alpha1.SecretKeyReference,
+) bool {
+	equal, _, _ := compareSecretKeyReferenceEqual(a, b)
+	return equal
+}
+
+// compareSecretKeyReferenceEqual returns true if the supplied slices of secret
+// key reference pointers contain the exact same elements. It will also return
+// 2 slices containing the added and removed elements. The comparison doesn't
```
RedbackThomson, line 42: nit: Maybe add `It will also return 2 slices containing elements contained in a that aren't in b, and elements contained in b that aren't in a, respectively.`

"added and removed" depends on the order of elements, so this would just clear up the confusion.

pkg/compare/secret_key_reference.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if len(added) == 0 && len(removed) == 0 {
+		equal = true
+	}
+	return equal, cleanUpDuplicateSecretReferences(added), cleanUpDuplicateSecretReferences(removed)
+}
+
+// cleanUpDuplicateSecretReferences removes duplicate elements from a given slice
+// of secret key references.
+func cleanUpDuplicateSecretReferences(
```
RedbackThomson, line 85: Ew, I hate that this is the best way to do this using vanilla Go. We really should move to Go 1.18 and use https://github.com/samber/lo

jaypipes:
@A-Hilaly how we doing on this one?

A-Hilaly:
@RedbackThomson @jaypipes updated the PR pleasr retake a look

PR #86, opened 2022-05-12
add Delta.DifferentExcept() method

jaypipes:
I need the ability to detect whether there are any differences in a
resource's fields *other than* a particular one. Basically, in doing the
tagging work for DBInstance, I came across a problem where we were
calling the ModifyDBInstance API call even though there were no changes
to DBInstance Spec fields *other than* the Tags collection.

So, I need to add the following in an sdk_update_pre_build_request
template for DBInstance:

```go
if delta.DifferentAt("Spec.Tags") {
     if err = rm.SyncTags(ctx, desired, latest); err != nil {
         return nil, err
     }
}
if !delta.DifferentExcept("Spec.Tags") {
     // We don't want to proceed to call the ModifyDBInstance API since
     // no other resource fields have changed.
     return desired, nil
}
```

This PR adds the `delta.DifferentExcept()` implementation to enable the
above code.

Issue aws-controllers-k8s/community#1276

### DIFF
--- a/pkg/compare/delta.go
+++ b/pkg/compare/delta.go
@@ -34,6 +34,49 @@ func (d *Delta) DifferentAt(subject string) bool {
 	return false
 }
 
+// DifferentExcept returns true if the delta contains any differences *other*
+// than any of the supplied path strings.
+//
+// This method is useful if you have a scenario where you don't want to proceed
+// with a modification action if certain fields in a resource have not been
+// changed.
+//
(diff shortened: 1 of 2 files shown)


RedbackThomson:
Cute little helper method

/lgtm

brucegucode:
not too familiar with this code but code logic looks good to me 
/lgtm

PR #87, opened 2022-05-19
update for default resource tags

vijtrip2:

Description of changes:
* This PR changes the default ACK resource tags from empty list to []string{"services.k8s.aws/controller-version=%CONTROLLER_VERSION%" ,"services.k8s.aws/namespace=%K8S_NAMESPACE%"}.

* Instead of statically passing the controller release version from release artifacts, i made it a dynamic `%CONTROLLER_VERSION%` variable. The dynamic variable will allow customers to use this variable in their custom tags as well.

* I refactored the various service controller metadata fields into a new `ServiceControllerMetadata` type as well.

Tested locally that the resource tags get resolved as expected.

### DIFF
--- a/pkg/runtime/tags.go
+++ b/pkg/runtime/tags.go
@@ -14,17 +14,26 @@
 package runtime
 
 import (
+	"fmt"
 	"strings"
-	"time"
+
+	rtclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	ackconfig "github.com/aws-controllers-k8s/runtime/pkg/config"
-	rtclient "sigs.k8s.io/controller-runtime/pkg/client"
+	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+)
+
+const (
+	// MissingImageTagValue is the placeholder value when ACK controller
+	// image tag(release semver) cannot be determined.
+	MissingImageTagValue = "unknown"
 )
 
 // GetDefaultTags provides Default tags (key value pairs) for given resource
 func GetDefaultTags(
 	config *ackconfig.Config,
 	object rtclient.Object,
+	md acktypes.ServiceControllerMetadata,
 ) map[string]string {
 	if object == nil || config == nil || len(config.ResourceTags) == 0 {
 		return nil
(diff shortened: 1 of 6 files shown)


pkg/runtime/tags.go
```diff
@@ -52,18 +61,34 @@ func GetDefaultTags(
...
 	if value == nil || obj == nil {
 		return nil
 	}
 	var expandedValue string = ""
 	switch *value {
-	case "%UTCNOW%":
-		expandedValue = time.Now().UTC().String()
-	case "%KUBERNETES_NAMESPACE%":
+	case "%CONTROLLER_VERSION%":
+		expandedValue = generateControllerVersion(md)
```
RedbackThomson, line 72: Could we split this into two different variables? Perhaps `%CONTROLLER_SERVICE%` and `%CONTROLLER_VERSION%`? This way users can, with more granularity, choose their tag formats.

RedbackThomson (commented):
Left a suggestion inline.

I think it would be good to support a few more variables, even if they aren't used in the default. My suggestions would be:
- `%OWNER_ACCOUNT_ID%` - taken from `Status`
- `%K8S_RESOURCE_NAME%` - taken from `metadata.name`
- `%K8S_CRD_VERSION%` - the CRD version (`v1alpha1`)

pkg/runtime/tags.go
vijtrip2: how about i tackle this in a followup PR?

This PR can deal with refactor of `serviceController` and addition of a new tag field. Keeping existing expandValue method as is.

Next PR will deal with supporting more reserved keywords, and find + replace keywords in all resource tags, not just the ones that match the switch case.

RedbackThomson: The change I suggested will not be backward compatible. If we delay that change but push out this functionality, there is a chance that one of our GA controllers will be released between now and then with breaking compatibility. I'd prefer that if we are going to start enabling this feature, that we start off with the format we want to support in the long-term.

vijtrip2: Yeah, by followup PR, i meant next PR(before releasing this feature).

And I am curious to know what will be the backward incompatible change. I did not get that part.

RedbackThomson: Oh the backward incompatibility would be `%CONTROLLER_VERSION%` changing from `s3-v1.0.0` to `v1.0.0`. But I guess as long as we get both of these out before cutting a new runtime release then I'm fine with it.

RedbackThomson (approved):
/lgtm

pkg/runtime/tags.go
vijtrip2: After supporting these different tag variables, I was planning on changing the default tag to `services.k8s.aws/controller-version=%CONTROLLER_SERVICE%-%CONTROLLER_VERSION%` and then the output tag value will stay the same.

Also I wanted to do inplace substring replacement instead complete string replacement in the next PR.

PR #88, opened 2022-05-22
support multiple ACK resource tag formats

vijtrip2:

Description of changes:
* support '%CONTROLLER_SERVICE%' , '%CONTROLLER_VERSION%', '%K8S_NAMESPACE%' and '%K8S_RESOURCE_NAME%' as the resource tag format
* support tag format as substring, and not just complete tag value
* As per the [comment](https://github.com/aws-controllers-k8s/runtime/pull/87#pullrequestreview-979427535) , I did not add 'OWNER_ACCOUNT_ID' because any AWS related information will be known to customer when describing tags or viewing the resource in AWS console.  
* I also did not add 'K8S_CRD_VERSION' right now and waiting for multi version support , to find out the best place to find the k8s resource crd version.


pkg/runtime/tags.go
```diff
@@ -23,19 +22,59 @@ import (
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
 )
 
+// expandTagFormat is a function which returns the resolved value for an
+// ACK resource tag format. Ex: %CONTROLLER_SERVICE% -> s3
+type expandTagFormat func(rtclient.Object, acktypes.ServiceControllerMetadata) string
```
A-Hilaly, line 27: What is `rtclient.Object` used for?

A-Hilaly: If this function calls `rtclient.Object` for future ResourceTagFormat we should also return an `error`

pkg/runtime/tags.go
```diff
@@ -23,19 +22,59 @@ import (
...
 
+// ACKResourceTagFormats is map of ACK resource tag formats to it's
+// expandTagFormat function.
+//
+// To add a new ACKResourceTag format, include it in this map, along with the
+// expandTagFormat function and expandTagValue() method will start
+// expanding the new resource tag format.
+var ACKResourceTagFormats = map[string]expandTagFormat{
+	"%CONTROLLER_SERVICE%": func(obj rtclient.Object,
+		md acktypes.ServiceControllerMetadata) string {
```
A-Hilaly, line 43: nit:
```suggestion
	"%CONTROLLER_SERVICE%": func(
	obj rtclient.Object,
	md acktypes.ServiceControllerMetadata,
) string {
```

pkg/runtime/tags.go
A-Hilaly, line 27: I am also not super convinced with the name `expandTagFormat`, maybe something like `resolveTagValues` makes more sense?

vijtrip2: `rtClient.Object` is used to retrieve `Name` and `Namespace` fields from the Object metadata.

I chose to stay away from error handling when resolving Tag formats, and instead use the pattern of injecting default values if the field is missing. Ex: `MissingImageTagValue` for the controller-version field.

vijtrip2: changed it to `resolveTagFormat`

A-Hilaly (approved):
:+1:

PR #90, opened 2022-05-25
add new AWSResourceManager.EnsureControllerTags() method

vijtrip2:

Description of changes:
* Adds new `AWSResourceManager.EnsureControllerTags()` method
* `EnsureControllerTags` method will be called before invoking `reconcile` method to update the desired resource with the ACK controller tags
* This change also introduces a new `ResourceTags` shape, which will be used as mediator/hub to merge AWS tags represented using different go types.
* code-generator will generate the logic of conversion to/from `ResourceTags`

### DIFF
--- a/pkg/tags/resource_tags.go
+++ b/pkg/tags/resource_tags.go
@@ -1,0 +1,52 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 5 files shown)


pkg/tags/resource_tags.go
```diff
@@ -0,0 +1,52 @@
...
+// permissions and limitations under the License.
+
+package tags
+
+// ResourceTags represents the AWS tags which will be added to the AWS resource.
+// Inside aws-sdk-go, Tags are represented using multiple types, Ex: map of
+// string, list of structs etc...
+// ResourceTags type will be used as a hub/mediator to merge tags represented
+// using different types.
+type ResourceTags struct {
```
jaypipes, line 21: How about just calling this `Tags`? I don't think there is value in having ResourceTags versus some other type of Tags... it's all just a map[string]string anyway :)

jaypipes (changes requested):
Agree with @RedbackThomson's suggestion. Also another naming suggestion inline from me...

jaypipes (approved):
thanks for that explanation (again :) ), @vijtrip2!

PR #92, opened 2022-06-07
Patch metadata and spec when handling errors

RedbackThomson:
Description of changes:
If any update hook code tried to return a requeue error, the reconciler would short circuit and never patch the metadata and spec. As a result, fields returned by the API server were never being persisted back into the K8s API - which in turn lead to the delta finding differences every loop. This is the root cause for https://github.com/aws-controllers-k8s/community/issues/1251. 

This pull request adds patching into the `HandleReconcileError` method, which sits at the end of the call stack where we handle the re-queuing errors. Therefore, even if an error is returned, as long as the error is not terminal, the spec will be patched.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -814,6 +814,8 @@ func (r *resourceReconciler) HandleReconcileError(
 	err error,
 ) (ctrlrt.Result, error) {
 	if ackcompare.IsNotNil(latest) {
+		// Create a copy so we don't override the spec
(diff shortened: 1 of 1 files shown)


brucegucode:
Thanks! @RedbackThomson

jaypipes (approved):
Excellent sleuthing (again) @RedbackThomson :)

jaypipes:
@RedbackThomson feel free to cut a release of runtime.

PR #94, opened 2022-06-15
Terminal condition sets ResourceSynced=False

jaypipes:
This patch updates the `resourceReconciler.ensureConditions()` method to
set the `ACK.ResourceSynced` Condition value to `False` when a Terminal
error has been returned from the resource manager. Previously, the
`ACK.ResourceSynced` Condition value was erroneously being set to `True`
due to a misunderstanding of what `ACK.ResourceSynced` means.

`ACK.ResourceSynced` is a Condition that informs the Kubernetes user
*whether the desired state of a resource matches the latest observed
state of the resource*. In the case of a Terminal error, the desired
state of a resource will *never* match the latest observed state of the
resource because there is something invalid about the desired resource
state.

Further, this patch changes the default value for `ACK.ResourceSynced`
Condition to be `Unknown` instead of `False` for any non-Terminal
reconciler error. The reasoning behind this change is that for
non-Terminal errors, we simply do not know whether the desired resource
state matches the latest observed state or not.

### DIFF
--- a/pkg/condition/condition.go
+++ b/pkg/condition/condition.go
@@ -29,8 +29,9 @@ var (
 	NotManagedReason  = "This resource already exists but is not managed by ACK. " +
 		"To bring the resource under ACK management, you should explicitly adopt " +
 		"the resource by creating a services.k8s.aws/AdoptedResource"
-	NotSyncedMessage = "Resource not synced"
-	SyncedMessage    = "Resource synced successfully"
+	UnknownSyncedMessage = "Unable to determine if desired resource state matches latest observed state"
+	NotSyncedMessage     = "Resource not synced"
+	SyncedMessage        = "Resource synced successfully"
 )
 
 // Synced returns the Condition in the resource's Conditions collection that is
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -326,15 +326,19 @@ func (r *resourceReconciler) ensureConditions(
 		if reconcileErr != nil {
 			condReason = reconcileErr.Error()
 			if reconcileErr == ackerr.Terminal {
-				// A terminal condition by its very nature indicates a stable state
-				// for a resource being synced. The resource is considered synced
-				// because its state will not change.
-				condStatus = corev1.ConditionTrue
-				condMessage = ackcondition.SyncedMessage
-			} else {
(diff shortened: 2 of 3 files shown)


pkg/condition/condition.go
```diff
@@ -29,8 +29,9 @@ var (
 	NotManagedReason  = "This resource already exists but is not managed by ACK. " +
 		"To bring the resource under ACK management, you should explicitly adopt " +
 		"the resource by creating a services.k8s.aws/AdoptedResource"
-	NotSyncedMessage = "Resource not synced"
-	SyncedMessage    = "Resource synced successfully"
+	UnknownSyncedMessage = "Unable to determine if desired resource state matches latest observed state"
```
RedbackThomson, line 32: Instead of `desired resource state` do we want to just plainly say `spec`?

RedbackThomson (approved):
I just checked and I see that nowhere do we explicitly check `Synced == false`. We always check for `Synced != true`. Happy with these changes - just had a question about our messaging.

pkg/condition/condition.go
jaypipes: > Instead of `desired resource state` do we want to just plainly say `spec`?

The issue with saying "spec" is that the spec can contain both the desired and latest state :) We compare full AWSResource objects to determine desired versus latest because some fields representing the latest state are set in the spec...

pkg/runtime/reconciler.go
```diff
@@ -326,15 +326,19 @@ func (r *resourceReconciler) ensureConditions(
...
-				// for a resource being synced. The resource is considered synced
-				// because its state will not change.
-				condStatus = corev1.ConditionTrue
+				// A terminal condition is a stable state for a resource.
+				// Terminal conditions indicate that without changes to the
+				// desired state of a resource, the resource's desired state
+				// will never match the latest observed state. Thus,
+				// ACK.ResourceSynced must be False.
+				condStatus = corev1.ConditionFalse
 				condMessage = ackcondition.SyncedMessage
```
vijtrip2, line 335:
```suggestion
				condMessage = ackcondition.NotSyncedMessage
```

vijtrip2 (changes requested):
Looks good to me. Nicely explained.
This may spur some test failures in various controller's e2e assertions that we will have to update upon the new release.

PR #95, opened 2022-06-16
log the JSON patch document when updating resource

jaypipes:
This patch updates the `resourceReconciler.patchResourceMetadataAndSpec` and
`resourceReconciler.patchResourceStatus` methods in a couple ways.
First, I ensure that the desired and latest resource parameters are
*never* modified by either method by ensuring that only copies of
desired and latest are used when calling Patch against the API server.
Second, I added some debug-level logging of the exact JSON patch
document that is sent over the wire to the Kubernetes API server when
patching either the metadata/spec *or* the status.

Here is what this looks like in the debug logging output:

metadata+spec:

```
1.65540769296847e+09	DEBUG	ackrt	>>>> kc.Patch (metadata + spec)	{"account": "750630568209", "role": "", "region": "us-west-2", "kind": "Repository", "namespace": "default", "name": "ecr-repository-u8loelgfy", "is_adopted": false, "generation": 1}
1.6554076929743059e+09	DEBUG	ackrt	patched resource metadata + spec	{"account": "750630568209", "role": "", "region": "us-west-2", "kind": "Repository", "namespace": "default", "name": "ecr-repository-u8loelgfy", "is_adopted": false, "generation": 1, "json": "{\"metadata\":{\"generation\":2,\"resourceVersion\":\"1049\"},\"spec\":{\"encryptionConfiguration\":{\"encryptionType\":\"AES256\"},\"imageScanningConfiguration\":{\"scanOnPush\":false},\"imageTagMutability\":\"MUTABLE\",\"registryID\":\"750630568209\",\"tags\":null},\"status\":{\"conditions\":null}}"}
1.6554076929743316e+09	DEBUG	ackrt	<<<< kc.Patch (metadata + spec)	{"account": "750630568209", "role": "", "region": "us-west-2", "kind": "Repository", "namespace": "default", "name": "ecr-repository-u8loelgfy", "is_adopted": false, "generation": 1}
```

status:

```
1.6554077304127564e+09	DEBUG	ackrt	>> kc.Patch (status)	{"account": "750630568209", "role": "", "region": "us-west-2", "kind": "Repository", "namespace": "default", "name": "ecr-repository-yq327yxdo", "is_adopted": false, "generation": 5}
1.6554077304208796e+09	DEBUG	ackrt	patched resource status	{"account": "750630568209", "role": "", "region": "us-west-2", "kind": "Repository", "namespace": "default", "name": "ecr-repository-yq327yxdo", "is_adopted": false, "generation": 5, "json": "{\"metadata\":{\"resourceVersion\":\"1129\"},\"spec\":{\"tags\":[{\"key\":\"k1\",\"value\":\"v1\"}]},\"status\":{\"conditions\":[{\"lastTransitionTime\":\"2022-06-16T19:28:50Z\",\"message\":\"Late initialization successful\",\"reason\":\"Late initialization successful\",\"status\":\"True\",\"type\":\"ACK.LateInitialized\"},{\"lastTransitionTime\":\"2022-06-16T19:28:50Z\",\"message\":\"Resource synced successfully\",\"reason\":\"\",\"status\":\"True\",\"type\":\"ACK.ResourceSynced\"}]}}"}
1.6554077304209113e+09	DEBUG	ackrt	<< kc.Patch (status)	{"account": "750630568209", "role": "", "region": "us-west-2", "kind": "Repository", "namespace": "default", "name": "ecr-repository-yq327yxdo", "is_adopted": false, "generation": 5}
1.655407730420916e+09	DEBUG	ackrt	< r.patchResourceStatus	{"account": "750630568209", "role": "", "region": "us-west-2", "kind": "Repository", "namespace": "default", "name": "ecr-repository-yq327yxdo", "is_adopted": false, "generation": 5}
```

I specifically strip out all of the Kubernetes server-side apply
annotations out of the patch document because they are super verbose and
are not helpful since we don't use server-side apply.

I'm hoping this extra bit of logging information will be useful when
attempting to diagnose exactly when and what changes are being applied
to resources in the Kubernetes API.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -15,6 +15,7 @@ package runtime
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -580,8 +581,46 @@ func (r *resourceReconciler) lateInitializeResource(
 	return lateInitializedLatest, err
 }
 
-// patchResourceMetadataAndSpec patches the custom resource in the Kubernetes API to match the
-// supplied latest resource's metadata and spec.
+// getPatchDocument returns a JSON string containing the object that will be
+// patched in the Kubernetes API server.
+//
+// NOTE(jaypipes): Because the Kubernetes API server's server-side apply
+// functionality introduces an enormous amount of verbose annotations in the
+// resource metadata, and because those annotations are pretty unhelpful to
+// ACK, we strip all that stuff out of the returned patch document.
+func getPatchDocument(
+	patch client.Patch,
+	obj client.Object, // the diff of this will be represented in the patch
+) string {
+	js, _ := patch.Data(obj)
+	var m map[string]interface{}
+	_ = json.Unmarshal(js, &m)
+	if md, ok := m["metadata"]; ok {
+		// Strip out managedFields stuff, since it's super verbose and
+		// doesn't offer any value to us (since we don't use server-side
+		// apply
+		if mv, ok := md.(map[string]interface{}); ok {
+			if _, ok := mv["managedFields"]; ok {
+				delete(mv, "managedFields")
+			}
+		}
+	}
+	js, _ = json.Marshal(m)
+	return string(js)
+}
+
+// patchResourceMetadataAndSpec patches the custom resource in the Kubernetes
+// API to match the supplied latest resource's metadata and spec.
+//
+// NOTE(jaypipes): The latest parameter is *mutated* by this method: the
+// resource's metadata.resourceVersion is incremented in the process of calling
+// Patch. This is intentional, because without updating the resource's
+// metadata.resourceVersion, the resource cannot be passed to Patch again later
+// in the reconciliation loop if Patch is called with the Optimistic Locking
+// option.
+//
+// See https://github.com/kubernetes-sigs/controller-runtime/blob/165a8c869c4388b861c7c91cb1e5330f6e07ee16/pkg/client/patch.go#L81-L84
+// for more information.
 func (r *resourceReconciler) patchResourceMetadataAndSpec(
 	ctx context.Context,
 	desired acktypes.AWSResource,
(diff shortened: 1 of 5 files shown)


pkg/runtime/reconciler.go
```diff
@@ -604,27 +636,25 @@ func (r *resourceReconciler) patchResourceMetadataAndSpec(
...
-	rlog.Exit("kc.Patch (metadata + spec)", err)
-
-	if err != nil {
-		return err
+	dobj := desired.DeepCopy().RuntimeObject()
+	lobj := latest.DeepCopy().RuntimeObject()
+	patch := client.MergeFrom(dobj)
+	err = r.kc.Patch(ctx, lobj, patch)
+	if err == nil {
+		if rlog.IsDebugEnabled() {
```
RedbackThomson, line 644: Why is this condition necessary? Can we not just print to Debug and let the logger determine the log level of the output?

pkg/runtime/reconciler.go
```diff
@@ -604,27 +636,25 @@ func (r *resourceReconciler) patchResourceMetadataAndSpec(
...
-	)
-	// Reset the status of latest object after patching.
-	latest.SetStatus(latestCopy)
-	rlog.Exit("kc.Patch (metadata + spec)", err)
-
-	if err != nil {
-		return err
+	dobj := desired.DeepCopy().RuntimeObject()
+	lobj := latest.DeepCopy().RuntimeObject()
+	patch := client.MergeFrom(dobj)
```
RedbackThomson, line 641: We have tried doing something like this, I think. When we patch spec or metadata, it updates the `resourceVersion` and any of our existing references cannot be used for patching.

pkg/runtime/reconciler.go
```diff
@@ -638,13 +668,15 @@ func (r *resourceReconciler) patchResourceStatus(
...
 
 	rlog.Enter("kc.Patch (status)")
-	err = r.kc.Status().Patch(
-		ctx,
-		latest.RuntimeObject(),
-		client.MergeFrom(desired.DeepCopy().RuntimeObject()),
-	)
+	dobj := desired.DeepCopy().RuntimeObject()
+	lobj := latest.DeepCopy().RuntimeObject()
+	patch := client.MergeFrom(dobj)
```
RedbackThomson, line 673: It may be fine to do this, because I don't believe patching status updates the `resourceVersion`

RedbackThomson (commented):
I have concerns about this method - I have a feeling that this is similar to what I tried the first time we "fixed" patching

vijtrip2 (commented):
I agree with all the comments made by @RedbackThomson .
This feels like completing a full circle for patchMetadataAndSpec. PatchStatus looks good.

I am also curious if you tested this refactoring on some controller's e2e tests. I remember from past experience that Sagemaker and Elasticache logic and assertions surface issues with any patching updates.

pkg/runtime/reconciler.go
jaypipes: It's necessary because `getPatchDocument()` is a relatively expensive operation (marshals, unmarshals, marshals again JSON) and there's no point executing it if debug logging isn't enabled.

pkg/runtime/reconciler.go
jaypipes: @RedbackThomson I was under the impression that would only impact us if we were using optimistic locking option:

https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/client/patch.go#L81-L84

That said, I'm happy to revert this part of the patch if you're concerned about it. All e2e tests were passing in the ecr controller local build-kind tests I ran.

jaypipes:
@RedbackThomson @vijtrip2 OK, reverted the `patchResourceMetadataAndSpec()` back to reset the Status of `latest` after calling `Patch`.

pkg/runtime/reconciler.go
```diff
@@ -580,8 +581,46 @@ func (r *resourceReconciler) lateInitializeResource(
...
+	obj client.Object, // the diff of this will be represented in the patch
+) string {
+	js, _ := patch.Data(obj)
+	var m map[string]interface{}
+	_ = json.Unmarshal(js, &m)
+	if md, ok := m["metadata"]; ok {
+		// Strip out managedFields stuff, since it's super verbose and
+		// doesn't offer any value to us (since we don't use server-side
+		// apply
+		if mv, ok := md.(map[string]interface{}); ok {
```
A-Hilaly, line 602: Is it necessary to check the `md` type here? i suppose it's always going to be a `map[string]interface{}`

jaypipes: actually `md` is `interface{}` not `map[string]interface{}` since `md` is an element of the top-level `map[string]interface{}` :)

vijtrip2:
Happy to merge this.
@RedbackThomson , if you also have same conclusion, please merge away.

RedbackThomson:
Still waiting for the patching code changes to be removed

pkg/runtime/reconciler.go
```diff
@@ -604,27 +643,29 @@ func (r *resourceReconciler) patchResourceMetadataAndSpec(
...
+			js := getPatchDocument(patch, lorig.RuntimeObject())
+			rlog.Debug("patched resource metadata + spec", "json", js)
+		}
 	}
-	rlog.Debug("patched resource metadata and spec", "latest", latest)
-	return nil
+	// The call to Patch() above ends up setting the latest variable's Status
+	// to the value of the desired variable's Status. We do not want this
+	// behaviour; instead, we want to keep latest's original Status value.
+	latest.SetStatus(lorig)
```
jaypipes, line 659: @RedbackThomson I reverted this back to the original code (see this line above).

RedbackThomson (approved):
/lgtm

PR #97, opened 2022-06-22
By default requeue the resource with 10 hour delay

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1355

Description of changes:
* implement ReSync using `requeue.NeededAfter` instead of upstream controlle-runime's `SyncPeriod`
* ACK controllers do not use cached go-client for reading objects from APIServer, hence upstream controller-runime ReSync was not working. Upstream controller-runtime ReSync only works for cached objects.
* As part of this change, whenever the `RequeueOnSuccess` duration for an ACK resource is 0 , requeue it with a 10 hour delay to trigger ReSync.
* I did not change the default value of `RequeueOnSuccessSeconds` from 0 to 10 hours because RequeueOnSuccessSeconds = 0 still means resource will not be requeued again to the developers. This 10 hour ReSync is an ACK runtime behavior and hence i only made change in ACK runtime. There will no change needed in existing ACK controllers. 


pkg/runtime/reconciler.go
```diff
@@ -830,11 +834,32 @@ func (r *resourceReconciler) handleRequeues(
 			}
 			// The code below only executes for "ConditionTypeResourceSynced"
 			if condition.Status == corev1.ConditionTrue {
-				if duration := r.rmf.RequeueOnSuccessSeconds(); duration > 0 {
+				duration := r.rmf.RequeueOnSuccessSeconds()
+				if duration > 0 {
```
RedbackThomson, line 838: I think I prefer the original condition. If `duration < 0`, we should just assume it `= 0`.

pkg/runtime/reconciler.go
```diff
@@ -830,11 +834,32 @@ func (r *resourceReconciler) handleRequeues(
...
 					rlog.Debug(
 						"requeueing resource after resource synced condition true",
 					)
 					return latest, requeue.NeededAfter(nil, time.Duration(duration)*time.Second)
+				} else if duration == 0 {
+					// If there is no RequeueOnSuccess, then requeue the resource
+					// with "resyncPeriod" to perform drift detection and
```
RedbackThomson, line 845: nit
```suggestion
					// with "resyncPeriod" to perform periodic drift detection and
```

RedbackThomson (commented):
Just two nits. PTAL

pkg/runtime/reconciler.go
```diff
@@ -835,6 +839,26 @@ func (r *resourceReconciler) handleRequeues(
 						"requeueing resource after resource synced condition true",
 					)
 					return latest, requeue.NeededAfter(nil, time.Duration(duration)*time.Second)
+				} else {
```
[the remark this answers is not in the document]
vijtrip2: Oh, come on! ;)

PR #98, opened 2022-06-23
Update logger to ISO8601

RedbackThomson:
Description of changes:
The current logs are written in Unix nanosecond timestamp, but are awkwardly represented with scientific notation (eg. `1.6559302967712986e+09`). These timestamps are almost impossible for a human to read and because of their notation are also not suitable for log ingestion.

This pull request updates the logger to use the ISO8601 standard format (eg. `2022-06-23T23:19:35.040Z`). This should make it much clearer to developers (and end users) how much time elapsed between events. ISO8601 is a standard format for timestamps that is properly supported by log ingest tools, as well.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -53,6 +53,7 @@ var (
 			acktags.NamespaceTagFormat,
 		),
 	}
+	defaultLogLevel = zapcore.InfoLevel
 )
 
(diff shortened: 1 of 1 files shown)


RedbackThomson:
One second, just updating the command line argument descriptions

vijtrip2:
Sick!
/lgtm

PR #99, opened 2022-06-29
update log message when requeueing with no error

vijtrip2:
Description of changes:
* When a resource was being requeued without any error, the debug log message reads `requeue needed after error {... , "error": null, "after": "10h0m0s"...}` , which is a red herring for users when debugging controller issues.
* This PR updates the log message correctly to not say `requeue needed after error`, when the resource is being requeued with no error.
* Resource can be requeued without any error for drift remediation.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -949,20 +949,28 @@ func (r *resourceReconciler) HandleReconcileError(
 	var requeueNeededAfter *requeue.RequeueNeededAfter
 	if errors.As(err, &requeueNeededAfter) {
 		after := requeueNeededAfter.Duration()
-		rlog.Debug(
-			"requeue needed after error",
-			"error", requeueNeededAfter.Unwrap(),
-			"after", after,
-		)
+		if wrappedError := requeueNeededAfter.Unwrap(); wrappedError != nil {
+			rlog.Debug(
+				"requeue needed after error",
+				"error", wrappedError,
+				"after", after,
+			)
+		} else {
+			rlog.Debug("requeueing", "after", after)
+		}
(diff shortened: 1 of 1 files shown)


pkg/runtime/reconciler.go
```diff
@@ -949,20 +949,28 @@ func (r *resourceReconciler) HandleReconcileError(
...
-		)
+		if requeueNeededAfter.Unwrap() == nil {
+			rlog.Debug("requeueing", "after", after)
+		} else {
+			rlog.Debug(
+				"requeue needed after error",
+				"error", requeueNeededAfter.Unwrap(),
+				"after", after,
+			)
+		}
```
RedbackThomson, line 960:
```suggestion
		if wrapped := requeueNeededAfter.Unwrap(); wrapped != nil {
			rlog.Debug(
				"requeue needed after error",
				"error", wrapped,
				"after", after,
			)
		} else {
			rlog.Debug("requeueing", "after", after)
		}
```

pkg/runtime/reconciler.go
```diff
@@ -949,20 +949,28 @@ func (r *resourceReconciler) HandleReconcileError(
...
 	}
 
 	var requeueNeeded *requeue.RequeueNeeded
 	if errors.As(err, &requeueNeeded) {
-		rlog.Debug(
-			"requeue needed error",
-			"error", requeueNeeded.Unwrap(),
-		)
+		if requeueNeeded.Unwrap() == nil {
+			rlog.Debug("requeueing")
```
RedbackThomson, line 967: I think this would be clearer if it said `requeueing immediately`?

jaypipes (approved):
Very nice, thank you @vijtrip2! :)

PR #100, opened 2022-08-23
Add optional key to the FieldExport target

baijum:
FieldExport generates the key name with an output structure like
`<namespace>.<FieldExport-resource-name>`. This default naming structure
creates non-conflicting names as the keys. However, many applications
expect short names as keys. This is important, especially when the
Secret/ConfigMap is mounted as files; the file names are nothing but
the key names.

This PR adds an optional attribute (.spec.to.key) to specify the key
name in the FieldExport resource. If this attribute is specified,
use this key name instead of the default.

Signed-off-by: Baiju Muthukadan <baiju.m.mail@gmail.com>

Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1410


apis/core/v1alpha1/field_export.go
```diff
@@ -24,6 +24,7 @@ type FieldExportTarget struct {
 	// Namespace is marked as optional, so we cannot compose `NamespacedName`
 	Namespace *string               `json:"namespace,omitempty"`
 	Kind      FieldExportOutputType `json:"kind"`
+	Key       *string               `json:"key,omitempty"`
```
jaypipes, line 27: Go ahead and add a docstring above here so that our API reference docs will include an explanation for the field.

pkg/runtime/field_export_reconciler.go
```diff
@@ -296,6 +296,9 @@ func (r *fieldExportReconciler) writeToConfigMap(
 ) error {
 	// Construct the data key
 	key := fmt.Sprintf("%s.%s", desired.Namespace, desired.Name)
+	if desired.Spec.To.Key != nil {
+		key = *desired.Spec.To.Key
```
jaypipes, line 300: What happens if the user enters an empty string?

jaypipes (commented):
@baijum cool stuff! couple things for you inline :)

apis/core/v1alpha1/field_export.go
baijum: Added docs. a974e46b904f7299f756fcd0a90d5a66522859fb

pkg/runtime/field_export_reconciler.go
baijum: I have changed the logic to consider empty strings. With an empty string value, the key will fall back to `<namespace>.<FieldExport-resource-name>`. a974e46b904f7299f756fcd0a90d5a66522859fb

pkg/runtime/field_export_reconciler.go
```diff
@@ -340,6 +343,9 @@ func (r *fieldExportReconciler) writeToSecret(
 ) error {
 	// Construct the data key
 	key := fmt.Sprintf("%s.%s", desired.Namespace, desired.Name)
+	if desired.Spec.To.Key != nil && strings.TrimSpace(*desired.Spec.To.Key) != "" {
```
A-Hilaly, line 346: Shall we handle cases where `Spec.To` is nil? to avoid nil panics?

A-Hilaly (commented):
Looks great! left one question

pkg/runtime/field_export_reconciler.go
baijum: Done. f6caeceb92749758b5b6200c83dd384091650f8c

A-Hilaly (approved):
Neat! 👍

baijum:
Once this PR is merged. I am planning to write a blog. Here is the early draft: https://gist.github.com/baijum/96158408eaf544f692acfad42f2b49df

RedbackThomson (approved):
This is an awesome addition. It's something I considered in the original design, but left it out just because I thought it wasn't necessary for an initial release. Thank you!

/lgtm

PR #103, opened 2022-10-04
Add force-s3-path-style option

redradrat:
Description of changes:

In sdk v1 the config to use path-style interactions for S3, is on the central session config object. This means that we need to expose this setting as a runtime config option as well, as the session construction is centralized across controllers. 

It's (at least to me) not nice to thus have this as config for all rendered controllers, but the definite central aws.Config is really the culprit here, rather than having an interface or sthg. Is this better in v2? not sure...

For sdk v1 this is probably the way forward? 

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -156,7 +156,7 @@ func (r *resourceReconciler) Reconcile(ctx context.Context, req ctrlrt.Request) 
 	roleARN := r.getRoleARN(acctID)
 	endpointURL := r.getEndpointURL(desired)
 	gvk := desired.RuntimeObject().GetObjectKind().GroupVersionKind()
-	sess, err := r.sc.NewSession(region, &endpointURL, roleARN, gvk)
+	sess, err := r.sc.NewSession(region, &endpointURL, r.cfg.ForceS3PathStyle, roleARN, gvk)
 	if err != nil {
 		return ctrlrt.Result{}, err
 	}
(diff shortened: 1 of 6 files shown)


jaypipes (commented):
Hi @redradrat! Can you explain why you need to force the path-style for S3 communication? Is there something about your buckets that makes the virtual-host style untenable?

redradrat:
In some environments it might be needed to proxy endpoints for example or utilize services to mock s3 offline. 

In those environments there might not be the DNS infrastructure in place. As the option is there within the SDK would be great to make use of it.

pkg/runtime/reconciler.go
```diff
@@ -156,7 +156,7 @@ func (r *resourceReconciler) Reconcile(ctx context.Context, req ctrlrt.Request)
 	roleARN := r.getRoleARN(acctID)
 	endpointURL := r.getEndpointURL(desired)
 	gvk := desired.RuntimeObject().GetObjectKind().GroupVersionKind()
-	sess, err := r.sc.NewSession(region, &endpointURL, roleARN, gvk)
+	sess, err := r.sc.NewSession(region, &endpointURL, r.cfg.ForceS3PathStyle, roleARN, gvk)
```
A-Hilaly, line 159: I don't see why we wouldn't add a similar flag if this is supported by aws-sdk-go...
/remove-lifecycle frozen

A-Hilaly:
I don't see why we wouldn't add a similar flag if this is supported by aws-sdk-go...
/remove-lifecycle frozen

PR #104, opened 2022-10-04
add identity endpoint & unsafe url options

redradrat:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1498

Description of changes:

https://github.com/aws-controllers-k8s/runtime/blob/ae946aa1e0dd309db0064c2c347b724074725c60/pkg/config/config.go#L146-L154

As the comment says so perfectly, a temporary session is set up to get the account info from STS.
This PR adds options to:
1. configure the endpoint of this identity call via `--aws-identity-endpoint-url`. 
2. configure whether both `--aws-identity-endpoint-url` & `--aws-endpoint-url` should be checked for unsafe URL schema (aka http)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -168,9 +168,6 @@ func (cfg *Config) SetAWSAccountID() error {
 	if cfg.IdentityEndpointURL != "" {
 		awsCfg.Endpoint = aws.String(cfg.IdentityEndpointURL)
 	}
-	if cfg.Region != "" {
-		awsCfg.Region = aws.String(cfg.Region)
-	}
 
 	// use sts to find AWS AccountId
 	session, err := session.NewSession(&awsCfg)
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
These changes look very reasonable to me, thank you for your PR @redradrat! :)

redradrat:
@jaypipes do you think the AWS_REGION is fine in the AccountId lookup?

I was thinking to only use the IdentityEndpoint and not AWS_REGION or have a dedicated REGION var... 

WDYT?

redradrat:
I removed using `--aws-region` from the identity call. It might not align with the region that a user wants to configure for the service API, and a user can just configure the identity endpoint with the fully qualified path. (e.g. `--aws-identity-endpoint-url sts.eu-south-1.amazonaws.com`)

jaypipes (approved):
Yeah, I still like this, especially after removing the region from the STS GetCallerIdentity lookup. ++

PR #105, opened 2022-12-02
Regenerate CRDs using controller-gen `v0.9.2`

embano1:
Signed-off-by: Michael Gasch <15986659+embano1@users.noreply.github.com>

Description of changes: When using the Kubernetes `e2e-framework` for testing a CRD setup function fails due to the newline in the two bases CRDs provided by this `runtime`.

The fix is to remove the newlines which should not cause any issues with ACK downstream dependencies, controllers, code-gen, etc.

### DIFF
--- a/config/crd/bases/services.k8s.aws_adoptedresources.yaml
+++ b/config/crd/bases/services.k8s.aws_adoptedresources.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
--- a/config/crd/bases/services.k8s.aws_adoptedresources.yaml
+++ b/config/crd/bases/services.k8s.aws_adoptedresources.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.9.2
   creationTimestamp: null
   name: adoptedresources.services.k8s.aws
 spec:
@@ -169,6 +169,7 @@ spec:
                           - name
                           - uid
                           type: object
+                          x-kubernetes-map-type: atomic
                         type: array
                     type: object
                 required:
@@ -223,9 +224,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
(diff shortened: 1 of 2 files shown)


config/crd/bases/services.k8s.aws_adoptedresources.yaml
```diff
@@ -1,4 +1,3 @@
-
```
A-Hilaly, line 1: The real real problem here is the controller-gen version we used to generate those files (`v0.7.0`) - Check line 7 in both files. This bug was fixed in [`v0.8.0`](https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.8.0) (PR https://github.com/kubernetes-sigs/controller-tools/pull/517).
Can please regenerate these file using `v0.9.2` ? it's the version we use for the generated controller now.

A-Hilaly (commented):
Good catch, thank you @embano1 !
Left one comment below

config/crd/bases/services.k8s.aws_adoptedresources.yaml
A-Hilaly: Script to regen the CRDs is here https://github.com/aws-controllers-k8s/runtime/blob/main/scripts/build-apis.sh

embano1: Sorry, not sure I fully understand: I'm using `aws-controllers-k8s/code-generator@20e82a5` with `aws-controllers-k8s/runtime@007e038` (i.e. `latest`) to generate the files. From the tags those commits are way ahead of `v0.[8|9].x`?

embano1: And the weird thing is that the other service-related CRDs are created without the newline.

A-Hilaly: Yep the service related CRD are generated using controller-gen v0.9.2 - the common CRDs were generated long time ago using v0.7.0.
The thing is that CRD files are generated using `controller-gen` not `ack-generate`. 

Script that build service related CRDs: https://github.com/aws-controllers-k8s/code-generator/blob/main/scripts/build-controller.sh#L196-L200
Script that build common CRDs: https://github.com/aws-controllers-k8s/runtime/blob/main/scripts/build-apis.sh#L20-L22

A-Hilaly:
Just pushed [25f7576](https://github.com/aws-controllers-k8s/runtime/pull/105/commits/25f75768b1aa1fc022b707b21f9331b9758760d8) that regenerates the CRD files with controller-gen v0.9.2
/approve

PR #106, opened 2022-12-22
Add support for resource-specific resync periods and default drift remediation period

A-Hilaly:
Part of: https://github.com/aws-controllers-k8s/community/issues/1367

This patch introduces the ability to specify resource-specific resync periods in
the drift remediation configuration, as well as a default drift remediation period
in the controller configuration. The resync period for each reconciler is
determined by trying to retrieve it from the following sources, in this order:

1. A resource-specific period specified in the drift remediation configuration.
2. A resource-specific requeue on success period specified by the resource manager
   factory.
3. The default drift remediation period specified in the controller configuration.
4. The default resync period defined in the ACK runtime package.

This allows users to customize the drift remediation behavior for different
resources as needed, while still providing a fallback option for resources that do
not have a specific period specified.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,6 +17,9 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strconv"
+	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -32,20 +35,22 @@ import (
 )
 
 const (
-	flagEnableLeaderElection   = "enable-leader-election"
-	flagMetricAddr             = "metrics-addr"
-	flagEnableDevLogging       = "enable-development-logging"
-	flagAWSRegion              = "aws-region"
-	flagAWSEndpointURL         = "aws-endpoint-url"
-	flagAWSIdentityEndpointURL = "aws-identity-endpoint-url"
-	flagUnsafeAWSEndpointURLs  = "allow-unsafe-aws-endpoint-urls"
-	flagLogLevel               = "log-level"
-	flagResourceTags           = "resource-tags"
-	flagWatchNamespace         = "watch-namespace"
-	flagEnableWebhookServer    = "enable-webhook-server"
-	flagWebhookServerAddr      = "webhook-server-addr"
-	flagDeletionPolicy         = "deletion-policy"
-	envVarAWSRegion            = "AWS_REGION"
+	flagEnableLeaderElection           = "enable-leader-election"
+	flagMetricAddr                     = "metrics-addr"
+	flagEnableDevLogging               = "enable-development-logging"
+	flagAWSRegion                      = "aws-region"
+	flagAWSEndpointURL                 = "aws-endpoint-url"
+	flagAWSIdentityEndpointURL         = "aws-identity-endpoint-url"
+	flagUnsafeAWSEndpointURLs          = "allow-unsafe-aws-endpoint-urls"
+	flagLogLevel                       = "log-level"
+	flagResourceTags                   = "resource-tags"
+	flagWatchNamespace                 = "watch-namespace"
+	flagEnableWebhookServer            = "enable-webhook-server"
+	flagWebhookServerAddr              = "webhook-server-addr"
+	flagDeletionPolicy                 = "deletion-policy"
+	flagReconcileDefaultResyncSeconds  = "reconcile-default-resync-seconds"
+	flagReconcileResourceResyncSeconds = "reconcile-resource-resync-seconds"
+	envVarAWSRegion                    = "AWS_REGION"
 )
 
 var (
@@ -63,20 +68,22 @@ var (
 
 // Config contains configuration options for ACK service controllers
 type Config struct {
-	MetricsAddr              string
-	EnableLeaderElection     bool
-	EnableDevelopmentLogging bool
-	AccountID                string
-	Region                   string
-	IdentityEndpointURL      string
-	EndpointURL              string
-	AllowUnsafeEndpointURL   bool
-	LogLevel                 string
-	ResourceTags             []string
-	WatchNamespace           string
-	EnableWebhookServer      bool
-	WebhookServerAddr        string
-	DeletionPolicy           ackv1alpha1.DeletionPolicy
+	MetricsAddr                    string
+	EnableLeaderElection           bool
+	EnableDevelopmentLogging       bool
+	AccountID                      string
+	Region                         string
+	IdentityEndpointURL            string
+	EndpointURL                    string
+	AllowUnsafeEndpointURL         bool
+	LogLevel                       string
+	ResourceTags                   []string
+	WatchNamespace                 string
+	EnableWebhookServer            bool
+	WebhookServerAddr              string
+	DeletionPolicy                 ackv1alpha1.DeletionPolicy
+	ReconcileDefaultResyncSeconds  int
+	ReconcileResourceResyncSeconds []string
 }
 
 // BindFlags defines CLI/runtime configuration options
@@ -151,6 +158,19 @@ func (cfg *Config) BindFlags() {
 	flag.Var(
 		&cfg.DeletionPolicy, flagDeletionPolicy,
 		"The default deletion policy for all resources managed by the controller",
+	)
+	flag.IntVar(
+		&cfg.ReconcileDefaultResyncSeconds, flagReconcileDefaultResyncSeconds,
+		60,
+		"The default duration, in seconds, to wait before resyncing desired state of custom resources. "+
+			"This value is used if no resource-specific override has been specified. Default is 60 seconds.",
+	)
+	flag.StringArrayVar(
+		&cfg.ReconcileResourceResyncSeconds, flagReconcileResourceResyncSeconds,
+		[]string{},
+		"A Key/Value list of strings representing the reconcile resync configuration for each resource. This"+
+			" configuration maps resource kinds to drift remediation periods in seconds. If provided, "+
+			" resource-specific resync periods take precedence over the default period.",
 	)
 }
 
@@ -233,6 +253,16 @@ func (cfg *Config) Validate() error {
 	if cfg.DeletionPolicy == "" {
 		cfg.DeletionPolicy = ackv1alpha1.DeletionPolicyDelete
 	}
+
+	if cfg.ReconcileDefaultResyncSeconds < 0 {
+		return fmt.Errorf("invalid value for flag '%s': resync seconds default must be greater than 0", flagReconcileDefaultResyncSeconds)
+	}
+
+	_, err := cfg.ParseReconcileResourceResyncSeconds()
+	if err != nil {
+		return fmt.Errorf("invalid value for flag '%s': %v", flagReconcileResourceResyncSeconds, err)
+	}
+
 	return nil
 }
 
@@ -243,4 +273,51 @@ func (cfg *Config) checkUnsafeEndpoint(endpoint *url.URL) error {
 		}
 	}
 	return nil
+}
+
+// ParseReconcileResourceResyncSeconds parses the values of the --reconcile-resource-resync-seconds
+// flag and returns a map that maps resource names to resync periods.
+// The flag arguments are expected to have the format "resource=seconds", where "resource" is the
+// name of the resource and "seconds" is the number of seconds that the reconciler should wait before
+// reconciling the resource again.
+func (cfg *Config) ParseReconcileResourceResyncSeconds() (map[string]time.Duration, error) {
+	resourceResyncPeriods := make(map[string]time.Duration, len(cfg.ReconcileResourceResyncSeconds))
+	for _, resourceResyncSecondsFlag := range cfg.ReconcileResourceResyncSeconds {
+		// Parse the resource name and resync period from the flag argument
+		resourceName, resyncSeconds, err := parseReconcileFlagArgument(resourceResyncSecondsFlag)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing flag argument '%v': %v. Expected format: resource=seconds", resourceResyncSecondsFlag, err)
+		}
+		resourceResyncPeriods[strings.ToLower(resourceName)] = time.Duration(resyncSeconds)
+	}
+	return resourceResyncPeriods, nil
+}
+
+// parseReconcileFlagArgument parses a flag argument of the form "key=value" into
+// its individual elements. The key must be a non-empty string and the value must be
+// a non-empty positive integer. If the flag argument is not in the expected format
+// or has invalid elements, an error is returned.
+//
+// The function returns the parsed key and value as separate elements.
+func parseReconcileFlagArgument(flagArgument string) (string, int, error) {
+	delimiter := "="
+	elements := strings.Split(flagArgument, delimiter)
+	if len(elements) != 2 {
+		return "", 0, fmt.Errorf("invalid flag argument format: expected key=value")
+	}
+	if elements[0] == "" {
+		return "", 0, fmt.Errorf("missing key in flag argument")
+	}
+	if elements[1] == "" {
+		return "", 0, fmt.Errorf("missing value in flag argument")
+	}
+
+	resyncSeconds, err := strconv.Atoi(elements[1])
+	if err != nil {
+		return "", 0, fmt.Errorf("invalid value in flag argument: %v", err)
+	}
+	if resyncSeconds < 0 {
+		return "", 0, fmt.Errorf("invalid value in flag argument: expected non-negative integer, got %d", resyncSeconds)
+	}
+	return elements[0], resyncSeconds, nil
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -161,7 +161,7 @@ func (cfg *Config) BindFlags() {
 	)
 	flag.IntVar(
 		&cfg.ReconcileDefaultResyncSeconds, flagReconcileDefaultResyncSeconds,
-		60,
+		0,
 		"The default duration, in seconds, to wait before resyncing desired state of custom resources. "+
 			"This value is used if no resource-specific override has been specified. Default is 60 seconds.",
 	)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -163,7 +163,7 @@ func (cfg *Config) BindFlags() {
 		&cfg.ReconcileDefaultResyncSeconds, flagReconcileDefaultResyncSeconds,
 		0,
 		"The default duration, in seconds, to wait before resyncing desired state of custom resources. "+
-			"This value is used if no resource-specific override has been specified. Default is 60 seconds.",
+			"This value is used if no resource-specific override has been specified. Default is 10 hours.",
 	)
 	flag.StringArrayVar(
 		&cfg.ReconcileResourceResyncSeconds, flagReconcileResourceResyncSeconds,
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,0 +1,61 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package config
+
+import "testing"
+
+func TestParseReconcileFlagArgument(t *testing.T) {
+	tests := []struct {
+		flagArgument   string
+		expectedKey    string
+		expectedVal    int
+		expectedErr    bool
+		expectedErrMsg string
+	}{
+		// Test valid flag arguments
+		{"key=1", "key", 1, false, ""},
+		{"key=123456", "key", 123456, false, ""},
+		{"key=600", "key", 600, false, ""},
+		{"k=1", "k", 1, false, ""},
+		{"ke_y=123456", "ke_y", 123456, false, ""},
+
+		// Test invalid flag arguments
+		{"key", "", 0, true, "invalid flag argument format: expected key=value"},
+		{"key=", "", 0, true, "missing value in flag argument"},
+		{"=value", "", 0, true, "missing key in flag argument"},
+		{"key=value1=value2", "", 0, true, "invalid flag argument format: expected key=value"},
+		{"key=a", "", 0, true, "invalid value in flag argument: strconv.Atoi: parsing \"a\": invalid syntax"},
+		{"key=-1", "", 0, true, "invalid value in flag argument: expected non-negative integer, got -1"},
+		{"key=-123456", "", 0, true, "invalid value in flag argument: expected non-negative integer, got -123456"},
+		{"key=1.1", "", 0, true, "invalid value in flag argument: strconv.Atoi: parsing \"1.1\": invalid syntax"},
+	}
+	for _, test := range tests {
+		key, val, err := parseReconcileFlagArgument(test.flagArgument)
+		if err != nil && !test.expectedErr {
+			t.Errorf("unexpected error for flag argument '%s': %v", test.flagArgument, err)
+		}
+		if err == nil && test.expectedErr {
+			t.Errorf("expected error for flag argument '%s', got nil", test.flagArgument)
+		}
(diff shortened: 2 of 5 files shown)


pkg/config/config.go
```diff
@@ -145,6 +150,20 @@ func (cfg *Config) BindFlags() {
 		"Specific namespace the service controller will watch for object creation from CRD. "+
 			" By default it will listen to all namespaces",
 	)
+	flag.IntVar(
+		&cfg.DefaultDriftRemediationPeriod, flagDefaultDriftRemediationPerdiod,
```
jaypipes, line 154: How about we call this flag `--reconcile-default-resync-seconds`?

pkg/config/config.go
```diff
@@ -145,6 +150,20 @@ func (cfg *Config) BindFlags() {
...
+		60,
+		"The default duration, in seconds, to wait before attempting to remediate drift in custom resources. "+
+			"This value is used if no resource-specific period or requeue on success period has been specified. "+
+			"Default is 60 seconds.",
```
jaypipes, line 158:
```suggestion
		"The default duration, in seconds, to wait before resyncing desired state of custom resources. "+
			"This value is used if no resource-specific override has been specified. "+
			"Default is 60 seconds.",
```

pkg/config/config.go
```diff
@@ -145,6 +150,20 @@ func (cfg *Config) BindFlags() {
...
+	)
+	flag.StringVar(
+		&cfg.DriftRemediationConfigurationJSON, flagDriftRemediationConfigurationJSON,
```
jaypipes, line 161: How about we call this flag `--reconcile-resource-resync-seconds` and make it a [`flag.StringSliceVar`](https://pkg.go.dev/github.com/spf13/pflag#StringSliceVar). Then split each string by the `=` sign to form an eventual `map[string]int`.

That way, the user could specify multiple overrides with multiple invocations of the flag, like so:

```bash
./controller --reconcile-resource-resync-seconds "Function=30" \
  --reconcile-resource-resync-seconds "Alias=120"
```

and there would be no annoying JSON serialization/deserialization to do in a command-line flag.

jaypipes (commented):
Just some naming suggestions inline... otherwise I like it!

pkg/config/config.go
```diff
@@ -31,19 +32,21 @@ import (
...
+	flagAWSRegion                         = "aws-region"
+	flagAWSEndpointURL                    = "aws-endpoint-url"
+	flagAWSIdentityEndpointURL            = "aws-identity-endpoint-url"
+	flagUnsafeAWSEndpointURLs             = "allow-unsafe-aws-endpoint-urls"
+	flagLogLevel                          = "log-level"
+	flagResourceTags                      = "resource-tags"
+	flagWatchNamespace                    = "watch-namespace"
+	flagEnableWebhookServer               = "enable-webhook-server"
+	flagWebhookServerAddr                 = "webhook-server-addr"
+	flagDefaultDriftRemediationPerdiod    = "default-drift-remediation-period"
```
jljaco, line 47: typo:  `s/Perdiod/Period/`

pkg/config/config.go
```diff
@@ -31,19 +32,21 @@ import (
...
+	flagDriftRemediationConfigurationJSON = "drift-remediation-configuration"
```
jljaco, line 48: nit:  drop the `JSON` suffix on the variable name

pkg/config/config.go
```diff
@@ -61,19 +64,21 @@ var (
...
+	IdentityEndpointURL               string
+	EndpointURL                       string
+	AllowUnsafeEndpointURL            bool
+	LogLevel                          string
+	ResourceTags                      []string
+	WatchNamespace                    string
+	EnableWebhookServer               bool
+	WebhookServerAddr                 string
+	DefaultDriftRemediationPeriod     int
+	DriftRemediationConfigurationJSON string
```
jljaco, line 81: ditto re: the the nit about dropping the `JSON` suffix

pkg/config/config.go
```diff
@@ -222,6 +241,29 @@ func (cfg *Config) Validate() error {
...
+	}
+
+	if cfg.DriftRemediationConfigurationJSON != "" {
+		// Validate that the drift remediation configuration is a valid JSON string
+		var drc map[string]int
+		if err := json.Unmarshal([]byte(cfg.DriftRemediationConfigurationJSON), &drc); err != nil {
+			return fmt.Errorf("invalid value for flag '%s': %v", flagDriftRemediationConfigurationJSON, err)
+		}
+
+		// Validate that all resource-specific periods are greater than or equal to 0
```
jljaco, line 256: Can we confirm here whether the `kind` refers to a valid resource?

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,50 @@ func (r *resourceReconciler) getEndpointURL(
 	return r.cfg.EndpointURL
 }
 
+// getResyncPeriod returns the duration a reconciler should use to drift remediate custom
```
jljaco, line 1045: Clearer to say "the period of the recurring reconciler process which remediates drift in custom resources" here.   Generally, prefer the term "period" over "duration" here.

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,50 @@ func (r *resourceReconciler) getEndpointURL(
...
+// resources. It attempts to retrieve the duration from the following sources, in this order:
+// 1. A resource-specific drift remediation period specified in the drift remediation
+//    configuration JSON.
+// 2. A resource-specific requeue on success period specified by the resource manager factory.
```
jljaco, line 1049: Perhaps indicate that the resource manager factory is controller-specific, and thus this period is to be specified by controller authors.

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,50 @@ func (r *resourceReconciler) getEndpointURL(
...
+// 3. The default drift remediation period specified in the configuration.
```
jljaco, line 1050: Be more specific here:  "the controller's generator.yaml configuration".

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,50 @@ func (r *resourceReconciler) getEndpointURL(
...
+// 4. The default resync period defined in the ACK runtime package.
```
jljaco, line 1051: Nit: point out that this setting is found in this same file.

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,50 @@ func (r *resourceReconciler) getEndpointURL(
...
+//
+// Each reconciler has a unique value to use. This function should only be called during the
+// instantiation of an AWSResourceReconciler and should not be called during the reconciliation
```
jljaco, line 1054: Be more specific here:  instead of "the reconciliation function", give a reference to the specific function.

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,50 @@ func (r *resourceReconciler) getEndpointURL(
...
+	// First, try to use a resource-specific resync period if provided in the drift
+	// remediation configuration.
+	if duration, ok := drc[rmf.ResourceDescriptor().GroupKind().Kind]; ok && duration > 0 {
+		return time.Duration(duration) * time.Second
+	}
+
+	// Second, try to use a resource-specific requeue on success period specified by the
+	// resource manager factory. This value is set during the code generation of the
+	// controller and takes precedence over the default drift remediation period because
+	// it allows existing controllers that relies on this value to maintain their intended
```
jljaco, line 1072: typo: `s/relies/rely/`

jljaco (commented):
This is great work!    A few comments, mostly on the contents of the inline documentation.

pkg/config/config.go
A-Hilaly: Good suggestion! Unfortunately we cannot verify this at the configuration level. I can try to add a panic during controller startup (if a user provides a configuration for a non existing resource) WDYT ?

pkg/config/config.go
A-Hilaly: I like it!

pkg/config/config.go
A-Hilaly: Done

pkg/runtime/reconciler.go
A-Hilaly: This (point 3) is provided using a flag

pkg/config/config.go
A-Hilaly: Having second thoughts here, adding this might also require few changes in code-generator.. Reconcilers are only aware of one resources (the one they manage)

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,55 @@ func (r *resourceReconciler) getEndpointURL(
...
+	// Third, try to use the default resync period resync period specified during controller
+	// start-up.
+	if cfg.ReconcileDefaultResyncSeconds > 0 {
+		return time.Duration(cfg.ReconcileDefaultResyncSeconds) * time.Second
+	}
+
+	// If none of the above values are present or valid, use the default resync period
+	// defined in the ACK runtime package. Defined in `defaultResyncPeriod` within the
+	// same file
+	return defaultResyncPeriod
```
azpaulp, line 1091: Would it be good for this method to log the return value?  Since there are four (if I'm not mistaken) places the value can come from, it may be helpful for diagnostics.

pkg/config/config_test.go
```diff
@@ -0,0 +1,61 @@
...
+		{"key=", "", 0, true, "missing value in flag argument: expected non-empty value"},
+		{"=value", "", 0, true, "missing key in flag argument: expected non-empty key"},
+		{"key=value1=value2", "", 0, true, "invalid flag argument format: expected key=value"},
+		{"key=a", "", 0, true, "invalid value in flag argument: strconv.Atoi: parsing \"a\": invalid syntax"},
+		{"key=-1", "", 0, true, "invalid value in flag argument: expected non-negative integer, got -1"},
+		{"key=-123456", "", 0, true, "invalid value in flag argument: expected non-negative integer, got -123456"},
+		{"key=1.1", "", 0, true, "invalid value in flag argument: strconv.Atoi: parsing \"1.1\": invalid syntax"},
+	}
+	for _, test := range tests {
+		key, val, err := parseReconcileFlagArgument(test.flagArgument)
```
azpaulp, line 44: If we're going to test a private function by having our test share the same package as the unit under test, then how about we also test _runtime.reconcilier.getResyncPeriod_ via the same technique?

pkg/config/config.go
```diff
@@ -233,3 +263,50 @@ func (cfg *Config) checkUnsafeEndpoint(endpoint *url.URL) error {
...
+//
+// The function returns the parsed key and value as separate elements.
+func parseReconcileFlagArgument(flagArgument string) (string, int, error) {
+	delimiter := "="
+	elements := strings.Split(flagArgument, delimiter)
+	if len(elements) != 2 {
+		return "", 0, fmt.Errorf("invalid flag argument format: expected key=value")
+	}
+	if elements[0] == "" {
+		return "", 0, fmt.Errorf("missing key in flag argument: expected non-empty key")
```
RedbackThomson, line 298: nit: redundant error message
```suggestion
return "", 0, fmt.Errorf("missing key in flag argument")
```

pkg/config/config.go
```diff
@@ -233,3 +263,50 @@ func (cfg *Config) checkUnsafeEndpoint(endpoint *url.URL) error {
...
+	}
+	if elements[1] == "" {
+		return "", 0, fmt.Errorf("missing value in flag argument: expected non-empty value")
```
RedbackThomson, line 301: nit: redundant
```suggestion
return "", 0, fmt.Errorf("missing value in flag argument")
```

pkg/config/config.go
```diff
@@ -233,3 +263,50 @@ func (cfg *Config) checkUnsafeEndpoint(endpoint *url.URL) error {
...
+	}
+
+	resyncSeconds, err := strconv.Atoi(elements[1])
+	if err != nil {
+		return "", 0, fmt.Errorf("invalid value in flag argument: %v", err)
+	}
+	if resyncSeconds < 0 {
```
RedbackThomson, line 308: I think `resyncSeconds <= 0` should be the check. If there are 0 seconds of reconciliation, it'll be constantly reconciled with no exponential backoff. Maybe we can suggest at least 1 second of wait?

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,55 @@ func (r *resourceReconciler) getEndpointURL(
...
+func getResyncPeriod(rmf acktypes.AWSResourceManagerFactory, cfg ackcfg.Config) time.Duration {
+	// The reconciliation resync period configuration has already been validated as
+	// a clean map. Therefore, we can safely ignore any errors that may occur while
+	// parsing it and avoid changing the signature of NewReconcilerWithClient.
+	drc, _ := cfg.ParseReconcileResourceResyncSeconds()
+
+	// First, try to use a resource-specific resync period if provided in the resource
+	// resync period configuration.
+	resourceKind := rmf.ResourceDescriptor().GroupKind().Kind
+	if duration, ok := drc[resourceKind]; ok && duration > 0 {
```
RedbackThomson, line 1069: Should we parse all keys to lowercase? Also, here you have `duration > 0`, whereas in the validation you have a check that `duration < 0`. That means `duration == 0` is not accounted for. See my earlier comment for this.

pkg/runtime/reconciler.go
```diff
@@ -1064,6 +1042,55 @@ func (r *resourceReconciler) getEndpointURL(
...
+	}
+
+	// Second, try to use a resource-specific requeue on success period specified by the
+	// resource manager factory. This value is set during the code generation of the
+	// controller and takes precedence over the default resync period period because
+	// it allows existing controllers that rely on this value to maintain their intended
+	// behavior.
+	if duration := rmf.RequeueOnSuccessSeconds(); duration > 0 {
+		return time.Duration(duration) * time.Second
+	}
```
RedbackThomson, line 1080: ++ great forethought!

pkg/runtime/reconciler.go
RedbackThomson: +1 a simple log when the controller starts would be useful for users to debug any misconfigurations in their systems, rather than having to wait for the controller to requeue to see the requeue time.

RedbackThomson (commented):
Please see my comments regarding the `duration = 0` edge case. Otherwise nits

pkg/runtime/reconciler.go
jljaco: OK, then consider changing the language to "specified by the command-line flag"

pkg/config/config.go
jljaco: My concern is that users might mis-spell the resource name on the command line or in the config and as things stand right now, there would be simply a silent failure to enforce the intended resync period.  Is there another way to mitigate that possible problem?  If not, then let's be sure to emphatically document that there isn't any error checking so users should be careful to get the names right.

jaypipes (changes requested):

pkg/config/config_test.go
A-Hilaly: `config_test.go` is part of `config` package, however `runtime_test.go` is part of `runtime_test` package making it impossible to test private functions. I would prefer dealing with this issue in a separate issue/pr, as it requires a lot refactoring

pkg/config/config.go
```diff
@@ -233,3 +263,50 @@ func (cfg *Config) checkUnsafeEndpoint(endpoint *url.URL) error {
...
+	resourceResyncPeriods := make(map[string]time.Duration, len(cfg.ReconcileResourceResyncSeconds))
+	for _, resourceResyncSecondsFlag := range cfg.ReconcileResourceResyncSeconds {
+		// Parse the resource name and resync period from the flag argument
+		resourceName, resyncSeconds, err := parseReconcileFlagArgument(resourceResyncSecondsFlag)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing flag argument '%v': %v. Expected format: resource=seconds", resourceResyncSecondsFlag, err)
+		}
+		resourceResyncPeriods[resourceName] = time.Duration(resyncSeconds)
+	}
+	return resourceResyncPeriods, nil
```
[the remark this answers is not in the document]
A-Hilaly: +++ Agree with you folks. Since this requires changes to the `code-generator` too, I prefer to handle it in a separate PR and keep this one simple for now.

pkg/runtime/reconciler.go
A-Hilaly: Yes in this case i will also apply strings.ToLower for the arguments provided by the users

pkg/config/config.go
A-Hilaly: I just checked here, if a used provides 0 seconds, the controller will use the package default value which is 10hours

pkg/config/config.go
A-Hilaly: related: https://github.com/aws-controllers-k8s/runtime/pull/106#discussion_r1086940308

pkg/config/config.go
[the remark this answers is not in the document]
jljaco: Can you please create a GH issue to track the tech debt?

pkg/config/config.go
jljaco: https://github.com/aws-controllers-k8s/runtime/pull/106#discussion_r1088307180

pkg/config/config.go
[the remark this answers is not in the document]
A-Hilaly: Voila https://github.com/aws-controllers-k8s/community/issues/1647

pkg/runtime/reconciler.go
A-Hilaly: Added logging

pkg/config/config_test.go
jljaco: Could you please create a new issue w/ appropriate labels and link to it here?

A-Hilaly: Voila! https://github.com/aws-controllers-k8s/community/issues/1648

pkg/runtime/reconciler.go
```diff
@@ -1126,16 +1154,23 @@ func NewReconcilerWithClient(
 	metrics *ackmetrics.Metrics,
 	cache ackrtcache.Caches,
 ) acktypes.AWSResourceReconciler {
+	rtLog := log.WithName("ackrt")
+	resyncPeriod := getResyncPeriod(rmf, cfg)
+	rtLog.V(1).Info("Initiating reconciler",
```
jljaco, line 1159: It makes more sense to me to move this log message inside of `getResyncPeriod()` (and just pass the `rtLog` into that method as an optional parameter)

A-Hilaly: I disagree a little bit here. The thing is that moving this log statement to `getResyncPeriod` have few flows:
- Rarely (maybe never) seen logger as a parameter to functions, we usually use context, which we don't have here, and there is no need to.
- We will have redundant log statements (at least 4 times) and maybe with parameter declaration changes we can use a `defer` but that is just over-engineering a simple function.
- It makes more sense to log that "Initiating controller with an X resync value" than saying "Chose X resync value)
- The goal is to help users know what's their resource resync value. Yes, it might be valuable to explain to users why we chose a certain value, but in that case, that should apply to every corner case in the runtime library which is just a killer IMO.

jljaco: OK, since `getResyncPeriod()` is only ever called here, then this makes sense.

jljaco (approved):
Great job, thanks for the thorough work here!

pkg/runtime/reconciler.go
```diff
@@ -1103,6 +1082,55 @@ func (r *resourceReconciler) getEndpointURL(
 	return r.cfg.EndpointURL
 }
 
+// getResyncPeriod returns the period of the recurring reconciler process which ensures the desired
+// state of custom resources is maintained.
+// It attempts to retrieve the duration from the following sources, in this order:
+// 1. A resource-specific reconciliation resync period specified in the reconciliation resync
```
surajkota, line 1088: why resync configuration map?

can we use annotation on the resource like other existing features e.g.
- https://aws-controllers-k8s.github.io/community/docs/user-docs/multi-region-resource-management/
- https://aws-controllers-k8s.github.io/community/docs/user-docs/deletion-policy/

surajkota (changes requested):
this is a great feature

pkg/runtime/reconciler.go
A-Hilaly: I'm not a super fan of resource-level drift remediation control - but happy to hear what other folks think about it.

surajkota: Ack on the controller flag, my question is related to highest in order to precedence.

Annotation or configmap, both are giving control to the user to configure the resync period. Since we don't have a common configmap or a CRD to define all controller configurations, my preference is to keep the experience consistent and not introduce another place

pkg/config/config.go
```diff
@@ -152,6 +159,19 @@ func (cfg *Config) BindFlags() {
 		&cfg.DeletionPolicy, flagDeletionPolicy,
 		"The default deletion policy for all resources managed by the controller",
 	)
+	flag.IntVar(
+		&cfg.ReconcileDefaultResyncSeconds, flagReconcileDefaultResyncSeconds,
+		60,
```
RedbackThomson, line 164: Woah woah wait we don't want every resource to try and reconcile itself every 60 seconds. That's way too often. I was thinking on the order of every 6 or even 10 hours.

surajkota: Agree for following reasons:
1. We risk causing throttling error as number of resources managed increase
2. There can be side affect of No 1 on inter service communication, e.g. services which use Autoscaling, SageMaker hosting service makes call to autoscaling service and if throttling happens, the SageMaker service will get throttled and impact scale in/out of fleet serving customer traffic
3. Not all resources need this, e.g. SageMaker training jobs are one time jobs, SageMaker models do not support updates so the only way to introduce drift is by deleting and recreating the entire resource outside of ACK but still not resolvable by reconciling.

My vote is to drop this flag completely and let the service teams use `requeue_on_success` to configure or let the customer override based on their preference. I strongly suggest starting conservative here

pkg/runtime/reconciler.go
surajkota: Hi folks, Nick helped me with clarification that resync configuration map != `ConfigMap`, so we can resolve this comment with one suggestion, rename the comment to say `resource resync CLI flag` or something along those lines

also see my comment on No 3 here. We should drop it

surajkota (changes requested):

pkg/config/config.go
RedbackThomson: I think we can still support this default resync period and also have a setting that ignores this for one-time style resources like `TrainingJob`. Call that a future feature request. Managing resync periods for every resource individually would require users to update their configuration every time we add a new resource - this is essentially just a shortcut for all of them. 

I actually believe I don't want service teams to set `requeue_on_success` at all. It's ultimately more important that a customer can set their expectations for requeuing resources, since they know their drift conditions better than any ACK contributor will. Sure some resources should be shorter than 10 hours, but whether it's every 5 hours or every 10 minutes, it'll depend on everyone's specific deployment context.

pkg/config/config.go
```diff
@@ -152,6 +159,19 @@ func (cfg *Config) BindFlags() {
 		&cfg.DeletionPolicy, flagDeletionPolicy,
 		"The default deletion policy for all resources managed by the controller",
 	)
+	flag.IntVar(
+		&cfg.ReconcileDefaultResyncSeconds, flagReconcileDefaultResyncSeconds,
+		0,
+		"The default duration, in seconds, to wait before resyncing desired state of custom resources. "+
+			"This value is used if no resource-specific override has been specified. Default is 60 seconds.",
```
RedbackThomson, line 166: Please update this description for the new default

RedbackThomson (changes requested):
One last change then I'm happy to ship this

pkg/config/config.go
```diff
@@ -163,7 +163,7 @@ func (cfg *Config) BindFlags() {
 		&cfg.ReconcileDefaultResyncSeconds, flagReconcileDefaultResyncSeconds,
 		0,
 		"The default duration, in seconds, to wait before resyncing desired state of custom resources. "+
-			"This value is used if no resource-specific override has been specified. Default is 60 seconds.",
+			"This value is used if no resource-specific override has been specified. Default is 0 seconds.",
```
RedbackThomson, line 166: Well... the default is the 10 hour fallback. 0 seconds makes it sound like it's constantly reconciling without any pause

surajkota (approved):
Lgtm with this comment addressed

https://github.com/aws-controllers-k8s/runtime/pull/106#discussion_r1096216315

jaypipes (approved):
Let's get this shipped, please. :)

RedbackThomson:
Thanks for the last minute changes @A-Hilaly 

/lgtm

PR #107, opened 2023-01-06
Add deletion policy CLI arguments and annotations

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/82

Description of changes:
As per the proposal: https://github.com/aws-controllers-k8s/community/pull/1148

Adds support for a new optional CLI argument for the controller binary `--deletion-policy`, supporting `delete` and `retain` as available options. Setting this argument to `delete` leaves the controller as the current default behaviour, whereby it deletes the AWS resource under management before it is removed from the K8s API server. Setting this argument to `retain` modifies the default behaviour, leaving the AWS resources intact (taking no action on them) before removing it from the K8s API server.

Adds support for a new annotation on `Namespace` objects: `{service}.services.k8s.aws/deletion-policy: delete/retain` (where `{service}` is the service alias of the controller eg. `s3`). This annotation overrides the deletion policy behaviour configured for the controller for all resources deployed under the namespace.

Adds support for a new annotation on any ACK resource object: `services.k8s.aws/deletion-policy: delete/retain`. This annotation overrides only the deletion policy behaviour configured for the resource.

### DIFF
--- a/apis/core/v1alpha1/deletion_policy.go
+++ b/apis/core/v1alpha1/deletion_policy.go
@@ -1,0 +1,45 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package v1alpha1
+
+import "errors"
+
+// DeletionPolicy represents how the ACK reconciler will handle the deletion of
+// a resource. A DeletionPolicy of "delete" will delete the underlying AWS
+// resource, whereas a DeletionPolicy of "retain" will only delete the K8s
+// object leaving the AWS resource intact.
+type DeletionPolicy string
+
+const (
+	DeletionPolicyDelete DeletionPolicy = "delete"
+	DeletionPolicyRetain DeletionPolicy = "retain"
+)
+
+func (e *DeletionPolicy) String() string {
+	return string(*e)
+}
+
+func (e *DeletionPolicy) Set(v string) error {
+	switch v {
+	case string(DeletionPolicyDelete), string(DeletionPolicyRetain):
+		*e = DeletionPolicy(v)
+		return nil
+	default:
+		return errors.New(`invalid DeletionPolicy value`)
+	}
+}
+
+func (e *DeletionPolicy) Type() string {
+	return "DeletionPolicy"
+}
--- a/apis/core/v1alpha1/deletion_policy.go
+++ b/apis/core/v1alpha1/deletion_policy.go
@@ -13,7 +13,9 @@
 
 package v1alpha1
 
-import "errors"
+import (
+	"fmt"
+)
 
 // DeletionPolicy represents how the ACK reconciler will handle the deletion of
 // a resource. A DeletionPolicy of "delete" will delete the underlying AWS
@@ -36,7 +38,7 @@ func (e *DeletionPolicy) Set(v string) error {
 		*e = DeletionPolicy(v)
 		return nil
 	default:
-		return errors.New(`invalid DeletionPolicy value`)
+		return fmt.Errorf("invalid DeletionPolicy value: %s", v)
 	}
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -27,6 +27,7 @@ import (
 	ctrlrt "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	acktags "github.com/aws-controllers-k8s/runtime/pkg/tags"
 )
 
@@ -43,6 +44,7 @@ const (
 	flagWatchNamespace         = "watch-namespace"
 	flagEnableWebhookServer    = "enable-webhook-server"
 	flagWebhookServerAddr      = "webhook-server-addr"
+	flagDeletionPolicy         = "deletion-policy"
 	envVarAWSRegion            = "AWS_REGION"
 )
 
@@ -74,6 +76,7 @@ type Config struct {
 	WatchNamespace           string
 	EnableWebhookServer      bool
 	WebhookServerAddr        string
+	DeletionPolicy           ackv1alpha1.DeletionPolicy
 }
 
 // BindFlags defines CLI/runtime configuration options
@@ -144,6 +147,10 @@ func (cfg *Config) BindFlags() {
 		"",
 		"Specific namespace the service controller will watch for object creation from CRD. "+
 			" By default it will listen to all namespaces",
+	)
+	flag.Var(
+		&cfg.DeletionPolicy, flagDeletionPolicy,
+		"The default deletion policy for all resources managed by the controller",
 	)
 }
 
@@ -221,6 +228,10 @@ func (cfg *Config) Validate() error {
 
 	if cfg.EnableWebhookServer && cfg.WebhookServerAddr == "" {
 		return errors.New("empty webhook server address")
+	}
+
+	if cfg.DeletionPolicy == "" {
+		cfg.DeletionPolicy = ackv1alpha1.DeletionPolicyDelete
 	}
 	return nil
 }
--- a/pkg/runtime/cache/namespace.go
+++ b/pkg/runtime/cache/namespace.go
@@ -14,6 +14,7 @@
 package cache
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/go-logr/logr"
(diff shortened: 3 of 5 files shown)


jljaco (commented):
This looks very solid.  I have a couple of very minor comments, both about logging.

How do you propose to functionally test this change?  If, for example, in an e2e test of a controller, can you please link to the relevant PR or test here?

RedbackThomson:
> How do you propose to functionally test this change? If, for example, in an e2e test of a controller, can you please link to the relevant PR or test here?

I am writing an additional test for S3 that creates a series of different namespaces and resources with varying annotations. The only thing that we cannot test with a standard controller e2e test is the CLI argument (we can't update the controller CLI args inside a Pytest). For that the best I can do for now is manual testing, sadly

pkg/runtime/reconciler.go
```diff
@@ -194,10 +194,18 @@ func (r *resourceReconciler) reconcile(
...
-		return r.deleteResource(ctx, rm, res)
+		// Determine whether we should retain or delete the resource
+		if r.getDeletionPolicy(res) == ackv1alpha1.DeletionPolicyDelete {
+			// Resolve references before deleting the resource.
+			// Ignore any errors while resolving the references
+			res, _ = rm.ResolveReferences(ctx, r.apiReader, res)
+			return r.deleteResource(ctx, rm, res)
+		}
+
+		if err := r.setResourceUnmanaged(ctx, res); err != nil {
```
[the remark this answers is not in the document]
RedbackThomson: Added an `INFO` log line before setting resource unmanaged:
```
2023-01-10T00:31:24.228Z        INFO    ackrt   AWS resource will not be deleted - deletion policy set to retain  {"account": "394120634394", "role": "", "region": "us-west-2", "kind": "Bucket", "namespace": "default", "name": "test-bucket", "generation": 2}
```

pkg/runtime/reconciler.go
```diff
@@ -202,6 +202,8 @@ func (r *resourceReconciler) reconcile(
 			return r.deleteResource(ctx, rm, res)
 		}
 
+		rlog := ackrtlog.FromContext(ctx)
+		rlog.Info("AWS resource will not be deleted - deletion policy set to retain")
```
jljaco, line 206: Is it possible to include a mention of which resource is being retained?  Is there a convenient resource name available in `res` that you can copy a bit above before deletion?

RedbackThomson: This is automatically added into the logs when using the `rlog` object. I posted a comment underneath your last comment asking for logs with an example log message

RedbackThomson:
> > How do you propose to functionally test this change? If, for example, in an e2e test of a controller, can you please link to the relevant PR or test here?
> 
> I am writing an additional test for S3 that creates a series of different namespaces and resources with varying annotations. The only thing that we cannot test with a standard controller e2e test is the CLI argument (we can't update the controller CLI args inside a Pytest). For that the best I can do for now is manual testing, sadly

These tests can be found in this pull request https://github.com/aws-controllers-k8s/s3-controller/pull/91

pkg/runtime/reconciler.go
RedbackThomson: https://github.com/aws-controllers-k8s/runtime/pull/107/files#r1065230998

pkg/runtime/reconciler.go
```diff
@@ -1045,6 +1055,35 @@ func (r *resourceReconciler) getRegion(
 	return ackv1alpha1.AWSRegion(r.cfg.Region)
 }
 
+// getRegion returns the resource's deletion policy based on the default
```
RedbackThomson, line 1058: Just noticed a typo here. Fixing now

pkg/config/config.go
```diff
@@ -222,6 +229,10 @@ func (cfg *Config) Validate() error {
 	if cfg.EnableWebhookServer && cfg.WebhookServerAddr == "" {
 		return errors.New("empty webhook server address")
 	}
+
+	if cfg.DeletionPolicy == "" {
+		cfg.DeletionPolicy = ackv1alpha1.DeletionPolicyDelete
+	}
```
A-Hilaly, line 235: This is not the best place to do defaulting, maybe use `flag.StringVar` with a default value?

A-Hilaly: I understand that maybe you want to avoid doing type casting during reconciliation.. maybe we can make DeletionPolicy part of reconciler struct?

pkg/runtime/cache/namespace.go
```diff
@@ -177,6 +201,17 @@ func (c *NamespaceCache) setNamespaceInfoFromK8sObject(ns *corev1.Namespace) {
 	if ok {
 		nsInfo.endpointURL = EndpointURL
 	}
+
+	nsInfo.deletionPolicies = map[string]string{}
+	nsDeletionPolicySuffix := "." + ackv1alpha1.AnnotationDeletionPolicy
```
A-Hilaly, line 206: nit:  Very tiny optimization: we can declare this as a constant outside of this function to avoid making unnecessary allocations everytime the cache syncs

pkg/runtime/cache/namespace.go
```diff
@@ -177,6 +201,17 @@ func (c *NamespaceCache) setNamespaceInfoFromK8sObject(ns *corev1.Namespace) {
...
+	for key, elem := range nsa {
+		if !strings.HasSuffix(key, nsDeletionPolicySuffix) {
+			continue
+		}
+		service := strings.TrimSuffix(key, nsDeletionPolicySuffix)
+		nsInfo.deletionPolicies[service] = elem
```
A-Hilaly, line 212: I think we can apply a `string.ToLower` to `service` variable here. Thinking users can provide an `eCR` or `Eks` ...

A-Hilaly: I also see that you're using `strings.ToLower(service)` to query the deletion policy

pkg/runtime/reconciler.go
```diff
@@ -194,10 +194,20 @@ func (r *resourceReconciler) reconcile(
 	res acktypes.AWSResource,
 ) (acktypes.AWSResource, error) {
 	if res.IsBeingDeleted() {
-		// Resolve references before deleting the resource.
-		// Ignore any errors while resolving the references
-		res, _ = rm.ResolveReferences(ctx, r.apiReader, res)
-		return r.deleteResource(ctx, rm, res)
+		// Determine whether we should retain or delete the resource
+		if r.getDeletionPolicy(res) == ackv1alpha1.DeletionPolicyDelete {
```
A-Hilaly, line 198: Didn't know this was possible, I always thought it's not possible to use `==` for "similar" type aliases

pkg/runtime/reconciler.go
```diff
@@ -194,10 +194,20 @@ func (r *resourceReconciler) reconcile(
...
+			// Resolve references before deleting the resource.
+			// Ignore any errors while resolving the references
+			res, _ = rm.ResolveReferences(ctx, r.apiReader, res)
```
A-Hilaly, line 201: QQ: Any reason why we need to resolve the references here?

pkg/runtime/reconciler.go
```diff
@@ -1045,6 +1055,35 @@ func (r *resourceReconciler) getRegion(
...
+// precedence:
+//   - The resource's `services.k8s.aws/deletion-policy` annotation, if present
+//   - The resource's Namespace's `{service}.services.k8s.aws/deletion-policy` annotation, if present
+//   - The controller's `--deletion-policy` CLI flag
+func (r *resourceReconciler) getDeletionPolicy(
+	res acktypes.AWSResource,
+) ackv1alpha1.DeletionPolicy {
+	// look for deletion policy in CR metadata annotations
+	resAnnotations := res.MetaObject().GetAnnotations()
+	deletionPolicy, ok := resAnnotations[ackv1alpha1.AnnotationDeletionPolicy]
```
A-Hilaly, line 1071: I also didn't know it is possible to use type aliases for map access..

apis/core/v1alpha1/deletion_policy.go
```diff
@@ -0,0 +1,47 @@
...
+		*e = DeletionPolicy(v)
+		return nil
+	default:
+		return fmt.Errorf("invalid DeletionPolicy value: %s", v)
+	}
+}
+
+func (e *DeletionPolicy) Type() string {
+	return "DeletionPolicy"
+}
```
A-Hilaly, line 47: Are these methods used somewhere? i understand that `String()` implements the Stringer interface but i don't see what `Set` and `Type` are needed for
nit: since these are exported functions, can we have some comments for GoDoc?

A-Hilaly (commented):
Great one! Leaving few comments and questions!

pkg/runtime/reconciler.go
RedbackThomson: `DeletionPolicyDelete` is a string enum, so this is just a string comparison operation

pkg/runtime/reconciler.go
RedbackThomson: I didn't modify this code, here, I just added a conditional around it. But I believe we need to resolve references so that we have all the necessary fields to call `sdkDelete` if they require those fields.

pkg/runtime/reconciler.go
RedbackThomson: `AnnotationDeletionPolicy` is a string - nothing weird going on here.

apis/core/v1alpha1/deletion_policy.go
RedbackThomson: These are required to set an enum as a flag using the cobra library. It must implement the [`Value`](https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#Value) interface.

PR #111, opened 2023-02-09
Add `inline` directive to the secret reference

arjunrn:
The `SecretKeyReference` uses the `SecretReference` as an embedded stuct. When serializing through the apimachinery's [DefaultUnstructuredConverter](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#pkg-variables) the json `inline` directive needs to be presented so that the fields are inline.

Issue #, if available:

Description of changes:

### DIFF
--- a/apis/core/v1alpha1/secret.go
+++ b/apis/core/v1alpha1/secret.go
@@ -20,8 +20,9 @@ import (
 // SecretKeyReference combines a k8s corev1.SecretReference with a
 // specific key within the referred-to Secret
 type SecretKeyReference struct {
-	// Empty JSON tag is required to solve encountered struct field "" without JSON tag  error.
-	k8scorev1.SecretReference `json:""`
+	// Empty JSON tag with "inline" attribute is required to properly inline the
+	// field in JSON output due to k8s apimachinery constraints
+	k8scorev1.SecretReference `json:",inline"`
 	// Key is the key within the secret
 	Key string `json:"key"`
 }


apis/core/v1alpha1/secret.go
```diff
@@ -20,8 +20,7 @@ import (
 // SecretKeyReference combines a k8s corev1.SecretReference with a
 // specific key within the referred-to Secret
 type SecretKeyReference struct {
-	// Empty JSON tag is required to solve encountered struct field "" without JSON tag  error.
-	k8scorev1.SecretReference `json:""`
+	k8scorev1.SecretReference `json:",inline"`
```
A-Hilaly, line 23: I tried to marshal the `SecretKeyReference` using `,inline` and empty JSON tag and both returned the same result:
```go
package main

import (
	"fmt"

	"gopkg.in/yaml.v3"
	v1 "k8s.io/api/core/v1"
	
	"github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
)

func main() {
	s := &v1alpha1.SecretKeyReference{
		SecretReference: v1.SecretReference{},
		Key:             "key",
	}
	b, _ := yaml.Marshal(s)
	fmt.Println(string(b))
}
```
What am I missing here?

arjunrn: I checked again. There seems to be no difference in how the `Marshall` function from `encoding/json`, `gopkg.in/yaml.v3` and `sigs.k8s.io/yaml` serializes a field which is _inline_. I traced down our issue to the use of [runtime.Unstructured](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured). Sorry for the bother. Will followup there.

Julian-Chu: @A-Hilaly  
I have same issue with `FieldExport` when using runtime.DefaultUnstructuredConverter.
It seems runtime converter need inline to handle embedded struct properly,  [see inline test case in apimachinery](https://github.com/kubernetes/apimachinery/blob/7687996c715ee7d5c8cf1e3215e607eb065a4221/pkg/runtime/converter_test.go#L115)

could you try the following code:

```gopackage main

import (
	"fmt"

	corev1apha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
	"github.com/aws/aws-sdk-go/aws"
	"k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/util/json"
)

func main() {
	field := &corev1apha1.FieldExport{
		TypeMeta: metav1.TypeMeta{
			Kind:       "FieldExport",
			APIVersion: corev1apha1.GroupVersion.String(),
		},
		ObjectMeta: metav1.ObjectMeta{
			Name:      "test",
			Namespace: "test-ns",
		},
		Spec: corev1apha1.FieldExportSpec{
			From: &corev1apha1.ResourceFieldSelector{
				Resource: corev1apha1.NamespacedResource{
					GroupKind: metav1.GroupKind{
						Group: "group",
						Kind:  "kind",
					},
					Name: aws.String("name"),
				},
				Path: aws.String("path"),
			},
			To: &corev1apha1.FieldExportTarget{
				Name:      aws.String("name"),
				Kind:      corev1apha1.FieldExportOutputType("kind"),
				Namespace: aws.String("test-ns"),
				Key:       aws.String("key"),
			},
		},
	}

	bs, _ := json.Marshal(field)
	var a map[string]interface{}
	fmt.Println("---FieldExport---")
	_ = json.Unmarshal(bs, &a)
	fmt.Println(a)
	b, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(field)
	fmt.Println(b)

	fmt.Println("---SecretKeyReference---")
	ref := &corev1apha1.SecretKeyReference{
		SecretReference: v1.SecretReference{
			Name:      "test",
			Namespace: "test-ns",
		},
		Key: "test-key",
	}
	bs, _ = json.Marshal(ref)
	var c map[string]interface{}
	_ = json.Unmarshal(bs, &c)
	fmt.Println(c)
	d, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(ref)
	fmt.Println(d)
}
```

here is my console output
```shell
---FieldExport---
map[apiVersion:services.k8s.aws/v1alpha1 kind:FieldExport metadata:map[creationTimestamp:<nil> name:test namespace:test-ns] spec:map[from:map[path:path resource:map[group:group kind:kind name:name]] to:map[key:key kind:kind name:name namespace:test-ns]] status:map[conditions:<nil>]]
map[apiVersion:services.k8s.aws/v1alpha1 kind:FieldExport metadata:map[creationTimestamp:<nil> name:test namespace:test-ns] spec:map[from:map[path:path resource:map[groupKind:map[group:group kind:kind] name:name]] to:map[key:key kind:kind name:name namespace:test-ns]] status:map[conditions:<nil>]]
---SecretKeyReference---
map[key:test-key name:test namespace:test-ns]
map[key:test-key secretReference:map[name:test namespace:test-ns]]

```
if it's fine with you, I will also submit a PR to fix FieldExport

arjunrn: @A-Hilaly So I looked into this some more and it is indeed an issue in apimachinery. Having the inline directive affects how the `DefaultUnstructuredConverter` converts a resource to unstructured. I tracked it down to this section of code 

https://github.com/kubernetes/apimachinery/blob/6eedab24c4fc1453a5557599a6dcbaf2f9cd80c3/pkg/runtime/converter.go#L378-L394

Basically when the tag is `json:""` the logic defaults to using the field name. If there is anything else in the tag then it treats it as a special case and sets the name to empty which means it will inline the struct. I guess this is a bug in apimachinery. However it is convention to add `inline` to all inlined fields. Would you be open to merging this PR in that case?

For a demonstrative example:

```golang
package main

import (
	"fmt"

	"k8s.io/apimachinery/pkg/runtime"
	"sigs.k8s.io/yaml"
)

type SecretReference struct {
	Name      string `json:"name,omitempty"`
	Namespace string `json:"namespace,omitempty"`
}

type SecretKeyReferenceOld struct {
	SecretReference `json:""`
	// Key is the key within the secret
	Key string `json:"key"`
}

type SecretKeyReferenceInline struct {
	SecretReference `json:",inline"`
	// Key is the key within the secret
	Key string `json:"key"`
}

func main() {
	old := SecretKeyReferenceOld{
		SecretReference: SecretReference{Name: "test", Namespace: "testns"},
		Key:             "test-key",
	}
	output, _ := yaml.Marshal(old)
	fmt.Println(string(output))

	inline := SecretKeyReferenceInline{
		SecretReference: SecretReference{Name: "test", Namespace: "testns"},
		Key:             "test-key",
	}
	output, _ = yaml.Marshal(inline)
	fmt.Println(string(output))
	oldUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&old)
	output, _ = yaml.Marshal(oldUnstructured)
	fmt.Println(string(output))
	inlineUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&inline)
	output, _ = yaml.Marshal(inlineUnstructured)
	fmt.Println(string(output))
}
```

apis/core/v1alpha1/secret.go
```diff
@@ -20,8 +20,7 @@ import (
 // SecretKeyReference combines a k8s corev1.SecretReference with a
 // specific key within the referred-to Secret
 type SecretKeyReference struct {
-	// Empty JSON tag is required to solve encountered struct field "" without JSON tag  error.
-	k8scorev1.SecretReference `json:""`
```
jaypipes, line 24: @arjunrn Please restore and update the docstring comment :) Maybe something like:

```go
// Empty JSON tag with "inline" attribute is required to properly inline the
// field in JSON output due to k8s apimachinery constraints
```
```

jaypipes (commented):
No harm in this PR. Just please restore the docstring :)

apis/core/v1alpha1/secret.go
arjunrn: Done.

apis/core/v1alpha1/secret.go
A-Hilaly: Thank you @arjunrn and @Julian-Chu ! TIL :)

jaypipes (approved):
:+1: danke!

A-Hilaly (approved):
++++ Wunderbar!
/lgtm

PR #113, opened 2023-02-10
Bump `mockery` version to v2.19.0

A-Hilaly:
THis patch upgrades the version of Mockery used in the project from
v2.13 to v2.19. The upgrade was necessary due to compatibility issues
between Mockery and Go 1.20. Specifically, mockery v2.13 was causing an
`internal error: package "context" without types was imported from
"github.com/aws-controllers-k8s/runtime/pkg/types"` error.

The root cause of this issue was that `golang.org/x/tools@v0.5.0` was
needed in order to make the combination of Go 1.20 and mockery work
properly.

Upgrading to v2.19 resolved this issue and ensures that the project
continues to function as expected with Go 1.20.


jaypipes (approved):
:+1: @A-Hilaly if you could update the PR summary and/or commit message to explain more the mockery error that this is correcting, that would be great.

PR #117, opened 2023-03-03
Validate resource names for drift remediation flags

A-Hilaly:
[fixes https://github.com/aws-controllers-k8s/community/issues/1647]

As discussed in https://github.com/aws-controllers-k8s/runtime/pull/106

this patch brings resource name validation to the arguments passed to
`--reconcile-resource-resync-seconds`. It also slightly changes the
previously implemented `ParseReconcileResourceResyncSeconds` to avoid
uncessary validation ops.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -27,11 +27,13 @@ import (
 	"github.com/jaypipes/envutil"
 	flag "github.com/spf13/pflag"
 	"go.uber.org/zap/zapcore"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrlrt "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	acktags "github.com/aws-controllers-k8s/runtime/pkg/tags"
+	ackutil "github.com/aws-controllers-k8s/runtime/pkg/util"
 )
 
 const (
@@ -211,7 +213,15 @@ func (cfg *Config) SetAWSAccountID() error {
 }
 
 // Validate ensures the options are valid
-func (cfg *Config) Validate() error {
+func (cfg *Config) Validate(options ...Option) error {
+	merged := mergeOptions(options)
+	if len(merged.gvks) > 0 {
+		err := cfg.validateReconcileConfigResources(merged.gvks)
+		if err != nil {
+			return fmt.Errorf("invalid value for flag '%s': %v", flagReconcileResourceResyncSeconds, err)
+		}
+	}
+
 	if cfg.Region == "" {
 		return errors.New("unable to start service controller as AWS region is missing. Please pass --aws-region flag or set AWS_REGION environment variable")
 	}
@@ -257,12 +267,6 @@ func (cfg *Config) Validate() error {
 	if cfg.ReconcileDefaultResyncSeconds < 0 {
 		return fmt.Errorf("invalid value for flag '%s': resync seconds default must be greater than 0", flagReconcileDefaultResyncSeconds)
 	}
-
-	_, err := cfg.ParseReconcileResourceResyncSeconds()
-	if err != nil {
-		return fmt.Errorf("invalid value for flag '%s': %v", flagReconcileResourceResyncSeconds, err)
-	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -32,6 +32,7 @@ import (
 
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	acktags "github.com/aws-controllers-k8s/runtime/pkg/tags"
+	ackrtutil "github.com/aws-controllers-k8s/runtime/pkg/util"
```
jaypipes, line 35: you can just name this alias `ackutil` since we're already in the runtime package.

jaypipes (commented):
I'm good with this. In the future, it may be good to adopt an "option strategy" for the Config validation (or initialization). This "option strategy" would allow for introducing changes to the runtime that would not break binary/compile-time compatibility in the code generator (like has happened on the related PR there that introduces the resource manager factories getter).

The "option strategy" would look something like this:

```go
// pkg/config/option.go
type Option struct {
    kinds []string
    // other fields that can be used in validation/initialization...
}

// WithKinds instructs the configuration to validate against a set of
// supplied resource kinds
func WithKinds(kinds []string) Option {
    return Option{kinds: kinds}
}

// mergeOptions merges all Option structs into a single Option
// and sets any defaults to missing values
func mergeOptions(opts []Option) Option {
    merged := Option{}
    for _, opt := range opts {
        if opt.kinds != nil {
            merged.kinds = opt.kinds
        }
        // repeat for each field in Option ...
    }
    // set any defaults on merged if still nil...
    return merged
}
```

```go
// pkg/config/config.go
...
// Note this could also be a New() function instead of Config.Validate()...
func (cfg *Config) Validate(
    opts ...Option,
) error {
    merged := mergeOptions(opts)
    if merged.kinds != nil {
        // Validate the kinds...
    }
}
```

This way as we add additional validations or configuration options, we don't need to change the function signature for `Config.Validate()` -- or `pkg/config.New()`

A-Hilaly:
>I'm good with this. In the future, it may be good to adopt an "option strategy" for the Config validation (or initialization). This "option strategy" would allow for introducing changes to the runtime that would not break binary/compile-time compatibility in the code generator (like has happened on the related PR there that introduces the resource manager factories getter).

I love this idea. Let me get it in this PR ...

jaypipes (approved):
muy bueno, thanks @A-Hilaly!

pkg/config/config.go
```diff
@@ -275,6 +278,24 @@ func (cfg *Config) checkUnsafeEndpoint(endpoint *url.URL) error {
 	return nil
 }
 
+// validateReconcileConfigResources validates the --reconcile-resource-resync-seconds flag
+// by checking the resource names and their corresponding duration.
+func (cfg *Config) validateReconcileConfigResources(valideResourceNames []string) error {
```
RedbackThomson, line 283:
```suggestion
func (cfg *Config) validateReconcileConfigResources(validResourceNames []string) error {
```

pkg/config/config.go
```diff
@@ -275,6 +278,24 @@ func (cfg *Config) checkUnsafeEndpoint(endpoint *url.URL) error {
...
+	for _, resourceResyncSecondsFlag := range cfg.ReconcileResourceResyncSeconds {
+		resourceName, _, err := parseReconcileFlagArgument(resourceResyncSecondsFlag)
+		if err != nil {
+			return fmt.Errorf("error parsing flag argument '%v': %v. Expected format: resource=seconds", resourceResyncSecondsFlag, err)
+		}
+		if !ackutil.InStrings(resourceName, valideResourceNames) {
```
RedbackThomson, line 289: FYI `lo.Contains` - https://github.com/samber/lo#contains. But I think we want to do case insensitive contains

pkg/config/config.go
```diff
@@ -211,7 +212,15 @@ func (cfg *Config) SetAWSAccountID() error {
 }
 
 // Validate ensures the options are valid
-func (cfg *Config) Validate() error {
+func (cfg *Config) Validate(options ...Option) error {
+	merged := mergeOptions(options)
+	if len(merged.kinds) > 0 {
```
RedbackThomson, line 217: I don't really understand the value in the `options` struct? I think it's an early optimisation that isn't adding much - especially since we aren't providing defaults.

RedbackThomson (changes requested):

pkg/config/config.go
A-Hilaly: I'd like to do a special PR where we go and bring Contains everywhere. Probably after dealing with https://github.com/aws-controllers-k8s/pkg/pulls

pkg/config/config.go
A-Hilaly: Fren-glish moves

pkg/config/config.go
A-Hilaly: It's a know paradigm in Go - we could definitely do a follow up to bring default values this way. I agree that it's not bringing that much, but it allows us to add good stuff if needed :)

A-Hilaly: Thinking about https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/config/config.go#L253-L255

pkg/config/config.go
RedbackThomson: Works for me

RedbackThomson (approved):
/lgtm

PR #120, opened 2023-03-21
Use resource descriptor GVK for `NewSession`

RedbackThomson:
Description of changes:
Some version of K8s do not reliably return `TypeMeta` information when you call `apiReader.Get()` (see https://github.com/kubernetes/kubernetes/issues/3030 and https://github.com/kubernetes/kubernetes/issues/80609). This is a [known bug](https://github.com/kubernetes-sigs/controller-runtime/issues/1517) in `controller-runtime` that they don't plan on fixing. 

Parts of the code, namely around setting up the user agent, currently rely on these fields - and they are currently being given an empty struct (with empty strings for all values). To work around this bug, this PR has introduced a new `GroupVersionKind()` getter in the `ResourceDescriptor` (replacing the existing `GroupKind`), which returns a static description of the GVK. Then, rather than referencing the `RuntimeObject` `TypeMeta` properties, we can reference this new GVK getter anywhere in the runtime. It also replaces any existing use of `GroupKind` to now use `GroupVersionKind`.


A-Hilaly (approved):
Neat :+1:
Small breaking change but impact is huge.
/lgtm

PR #128, opened 2023-12-07
Add `WriteToSecret` method to `Reconciler` interface

ndbhat:
Description of changes:
WriteToSecret writes a string value to a Secret given the namespace, name, and key of the Secret

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	ctrlrt "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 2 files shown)


pkg/runtime/reconciler.go
```diff
@@ -152,6 +153,43 @@ func (r *reconciler) SecretValueFromReference(
...
+	nsn := types.NamespacedName{
+		Name: name,
+	}
+	nsn.Namespace = namespace
+
+	secret := &corev1.Secret{}
+	err := r.apiReader.Get(ctx, nsn, secret)
+	if err != nil {
+		return ackerr.SecretNotFound
+	}
```
a-hilaly, line 176: I think we don't need to do the `Get` here. Patch will fail anyways if the secret doesn't exist. Can you test this function with no `Get` calls?

ndbhat: `Get` retrieves the secret from the given namespace and name so I think we need to call it before Patch. I followed the field export WriteToSecret method [https://github.com/aws-controllers-k8s/runtime/blob/e5668082edc3bf40d7d2d6bdcf43e0189426d528/pkg/runtime/field_export_reconciler.go#L336C1-L382C1](https://github.com/aws-controllers-k8s/runtime/blob/e5668082edc3bf40d7d2d6bdcf43e0189426d528/pkg/runtime/field_export_reconciler.go#L336C1-L382C1)

go.mod
a-hilaly, line 1: Let's not include this change

pkg/runtime/reconciler.go
a-hilaly: Makes sense, thank you!

a-hilaly: One last ask, can we reuse this function in fieldExportReconciler?

a-hilaly: Discussed with Neha off-line we will address this in a seperate PR before releasing runtime 0.34.0

a-hilaly (approved):
👍 
/lgtm

PR #133, opened 2024-01-18
Introduce scope configuration for internal caches

a-hilaly:
This patch enhances the runtime internal caches by introducing a
configuration mechanism to tailor their behavior. The caches can now be
configured to watch a specific set of namespaces (instead of the
previous "all or nothing" style).

This is intended to be complete the multi-namspace-watch mode feature.
Now when a user configures a controller to watch a set of namespaces the
ACK runtime caches will also respect that scope.

In addition this commit adds `kube-node-lease` to the set of namespace
ignored by default (This namespace is dedicated to node Leases objects).

### DIFF
--- a/pkg/runtime/service_controller.go
+++ b/pkg/runtime/service_controller.go
@@ -14,6 +14,7 @@
 package runtime
 
 import (
+	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


a-hilaly:
cc @acornett21

pkg/runtime/service_controller.go
```diff
@@ -187,13 +189,39 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
...
+	}
+
+	cache := ackrtcache.New(c.log, ackrtcache.Config{
+		WatchScope: namespaces,
+		// Default to ignoring the ack-system kube-system, kube-public, and
+		// kube-node-lease namespaces.
+		// NOTE: Maybe we should make this configurable? It's not clear that
+		// we'd ever want to watch these namespaces.
+		Ignored: []string{
+			"ack-system",
```
acornett21, line 204: Do we really want to go down the path of hiding what is ignored? It gets into a slippery slope. Maybe for some `kube` namespaces, but I don't think `ack-system` should be ignored. There are probably users out there that are creating CR's in this namespace since it's the default namespace that the controller runs in.

a-hilaly: Since we added cross-account resource management to ACK, the caches ignored to the `ack-system` namespace (or the namespace where the controller is running). Still, the controller will handle resources in that namespace using the controller role. This means that only CARM will not work on `ack-system`. 

Now, looking at the bigger picture, I agree it might be a bit odd, but we might just allow it since there don't seem to be any issues.. and it won't have any side effects

a-hilaly: I still don't think it's a good idea to expose this level of configuration at the helm/flag level of controllers. However i'm down to remove `ack-system` from the ignored list and only leave `kube-system`, `kube-public` and `kube-node-lease`.

acornett21: > I still don't think it's a good idea to expose this level of configuration at the helm/flag level of controllers. However i'm down to remove `ack-system` from the ignored list and only leave `kube-system`, `kube-public` and `kube-node-lease`.

I also do not thing adding a configuration for this is needed, that would add more work/confusion and we'd need to change some documentation.

jonathan-innis (approved):
/lgtm

PR #140, opened 2024-02-29
Properly handle CARM errors and requeues

a-hilaly:
Prior to this patch we introduced new mechanisms to handle Role ARN
retrievals from the CARM cache. Which improved ACK runtime scaling
capabilities and addressed possible race condition scenarios. However in
the process we missed two things:
- Setting an `ACK.ResourceSynced` condition stating that the resource
  isn't synced, yet.
- Returning the **correct** runtime error that will cause the reconciller to
  requeue every 15seconds.

Both of the problems stemmed from the fact that we're not "yet" in the
reconcile function (`rm.Sync`) that is wrapped by a proper error handler
(that triggers requeues, and resets/sets resource conditions). In this
special case, we need to manually inject the condition and return a
controller-runtime error that will trigger a requeue after 15seconds.

While this is a "fair" fix, we're planning on refactoring a lot of the
runtime logic to make easier to read, maintain and more importantly
expose reusable component that will help avoid falling into such traps.

### DIFF
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -124,6 +124,7 @@ func (r *adoptionReconciler) reconcile(ctx context.Context, req ctrlrt.Request) 
 		// is not available.
 		roleARN, err = r.getRoleARN(acctID)
 		if err != nil {
+			ackrtlog.InfoAdoptedResource(r.log, res, fmt.Sprintf("Unable to start adoption reconcilliation %s: %v", acctID, err))
 			// r.getRoleARN errors are not terminal, we should requeue.
 			return requeue.NeededAfter(err, roleARNNotAvailableRequeueDelay)
 		}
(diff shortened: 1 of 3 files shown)


pkg/runtime/adoption_reconciler.go
```diff
@@ -124,6 +124,7 @@ func (r *adoptionReconciler) reconcile(ctx context.Context, req ctrlrt.Request)
 		// is not available.
 		roleARN, err = r.getRoleARN(acctID)
 		if err != nil {
+			ackrtlog.InfoAdoptedResource(r.log, res, fmt.Sprintf("Unable to start adoption reconcilliation %s: %v", acctID, err))
 			// r.getRoleARN errors are not terminal, we should requeue.
 			return requeue.NeededAfter(err, roleARNNotAvailableRequeueDelay)
```
a-hilaly, line 129: This error works fine, since `r.reconcile` is wrapped by `handleReconcileErrors`

PR #141, opened 2024-03-02
Introduce flags for fine-tuning maximum concurrent reconciles per resource

a-hilaly:
This commit aims to improve the performance and scalability of the ACK
service controllers by introducing support for configuring the maximum
number of concurrent reconciles for individual resources. The primary
motivation behind this change is to address varying workload demands and
resource requirements in Kubernetes environements.

This patch introduces two new flags (soon exposed to in the helm chart
values):
- `--reconcile-default-max-concurrent-syncs`: allow users to specify the
  default maximum concurrency level for all the resources.
- `--reconcile-resource-max-concurrent-syncs`: enable users to define
  resource-specific maximum concurrency settings, overriding the default
  value.

A use case example would be the scenario of an admin wanting to manage
EKS resources using the ACK `eks-controller`. It is normal for each
cluster to have multiple nodegroups/PIAs/AccessEntries/FargateProfiles,
which can indicate the proportion of needed max concurrencies for each
resource.

For instance, you can configure the EKS Controller with a default
maximum of 2 concurrent reconciles for all the resources, and override
the maximum concurrency to 10 for `AccessEntries` and
`PodIdentityAssociations`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -39,24 +39,26 @@ import (
 )
 
 const (
-	flagEnableLeaderElection           = "enable-leader-election"
-	flagLeaderElectionNamespace        = "leader-election-namespace"
-	flagMetricAddr                     = "metrics-addr"
-	flagHealthzAddr                    = "healthz-addr"
-	flagEnableDevLogging               = "enable-development-logging"
-	flagAWSRegion                      = "aws-region"
-	flagAWSEndpointURL                 = "aws-endpoint-url"
-	flagAWSIdentityEndpointURL         = "aws-identity-endpoint-url"
-	flagUnsafeAWSEndpointURLs          = "allow-unsafe-aws-endpoint-urls"
-	flagLogLevel                       = "log-level"
-	flagResourceTags                   = "resource-tags"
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -202,6 +206,19 @@ func (cfg *Config) BindFlags() {
 			" configuration maps resource kinds to drift remediation periods in seconds. If provided, "+
 			" resource-specific resync periods take precedence over the default period.",
 	)
+	flag.IntVar(
+		&cfg.ReconcileDefaultMaxConcurrency, flagReconcileDefaultMaxConcurrency,
+		1,
```
a-hilaly, line 211: We will maintain the default maximum reconciles at 1

pkg/config/config.go
```diff
@@ -202,6 +206,19 @@ func (cfg *Config) BindFlags() {
...
+		"The default maximum number of concurrent reconciles for a resource reconciler. This value is used if no "+
+			"resource-specific override has been specified. Default is 1.",
+	)
```
a-hilaly, line 214: We will maintain the default maximum reconciles at 1

PR #143, opened 2024-03-18
Create a seperated CARM map from accounts

zicongmei:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/2031

Description of changes:

Add a new version the CARM map in parallel of existing `ack-role-account-map` 

Usage:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: ack-carm-map
  namespace: $ACK_SYSTEM_NAMESPACE
data:
  owner-account-id/111111111111: arn:aws:iam::111111111111:role/s3FullAccess
```

If the account ID in `ack-carm-map` exists, we will use this value. Otherwise, it will use the value in the older version  `ack-role-account-map` .

This will allow the expension to the key of `team-id/team-a`.


zicongmei:
Offline discussion:
1. rename the configmap to `ack-setting-map`
2. hide it under a feature flag

PR #144, opened 2024-03-25
Add a new flag for enabling/disabling AdoptedResource reconciler

a-hilaly:
This commit adds a new feature flag `enable-adopted-resource-reconciler` to
allow enabling or disabling the `AdoptedResource` reconciler. The reconciler
is now only started if this flag is set to true.

This lays the groundwork for a new way to adopt resources, via an adopted
resource annotation instead of using the `AdoptedResource` custom resource.

This change provides more flexibility in choosing whether users want to
use the AdoptedResource CRD or the adoption annotation to adopt
resources.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -39,6 +39,7 @@ import (
 )
 
 const (
+	flagEnableAdoptedResourceReconciler = "enable-adopted-resource-reconciler"
 	flagEnableLeaderElection            = "enable-leader-election"
 	flagLeaderElectionNamespace         = "leader-election-namespace"
 	flagMetricAddr                      = "metrics-addr"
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -126,6 +128,11 @@ func (cfg *Config) BindFlags() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.",
 	)
+	flag.BoolVar(
+		&cfg.EnableAdoptedResourceReconciler, flagEnableAdoptedResourceReconciler,
+		true,
+		"Enable the AdoptedResource reconciler.",
+	)
```
a-hilaly, line 135: Keeping the default to true for now. We should start thinking about deprecation and removal dates, once the adoption via annotation is support.

PR #145, opened 2024-03-26
Add read-only support for ACK resources via annotation

a-hilaly:
This commit adds support for making ACK resources read-only via an
annotation. A new `services.k8s.aws/read-only` annotation can be
added to a resource to instruct the ACK service controller to not
create/patch/delete the underlying AWS resource.

If the annotation is set to true, the controller will:
- return an error on create if the resource doesn't exist
- only patch metadata/spec for updates but not make cloud API calls
- skip deletion of the underlying resource when the k8s resource
  is deleted

This allows adopting externally managed resources into ACK for read-only
purposes like referencing values, without ACK attempting to manage the
resource lifecycle.

### DIFF
--- a/pkg/runtime/reconciler_test.go
+++ b/pkg/runtime/reconciler_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zapcore"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


pkg/runtime/reconciler_test.go
a-hilaly, line 1: More reasons to switch to envtest.. https://github.com/aws-controllers-k8s/community/issues/1321

michaelhtm (commented):
🚀 🚀 
/lgtm

knottnt:
Link to the feature request issue:
https://github.com/aws-controllers-k8s/community/issues/2008

PR #146, opened 2024-03-26
Improve handling of missing namespace in `SecretKeyReference`

a-hilaly:
Fixes https://github.com/aws-controllers-k8s/community/issues/1699

When reconciling a resource that references a Kubernetes Secret via a
`SecretKeyReference`, if the namespace field is not specified, the code
previously defaulted to the "default" namespace.

This commit changes that behavior to instead use the namespace of the
resource being reconciled, if available in the reconcile context.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -126,8 +126,18 @@ func (r *reconciler) SecretValueFromReference(
 	}
 
 	namespace := ref.Namespace
-	if namespace == "" {
-		namespace = "default"
+	// During the reconcile process, the resourceNamespace is stored in the context
+	// and can be used to fetch the secret if the namespace is not provided in the
+	// SecretKeyReference.
+	//
+	// NOTE(a-hilaly): When refactoring the runtime, we might want to consider passing
+	// the ObjectMeta in the context.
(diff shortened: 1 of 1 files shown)


jlbutler:
Do we have any sanity check on if any use cases have a dependency on the secret being in `default`? I'm guessing no, just wanted to confirm you considered it.

a-hilaly:
@jlbutler Good call - all the tests we have around secret references, explicitly set a namespace. e.g RDS DBCluster https://github.com/aws-controllers-k8s/rds-controller/blob/main/test/e2e/tests/test_db_cluster.py#L52-L76

Ideally we want to test the defaulting behavior in the runtime package. Should we deal with it during the refactoring, as we will have easier testing (probably using envtest)? rather than just investing on more mocking.

jlbutler:
Yes I think that's ok to look at this for refactoring. I think this is the correct default (look to the namespace in context by default, not the general default).

I'm just invoking Hyrum's Law here and ensuring we consider this as a potentially breaking change, even if depending on this would be a bit of an anti-pattern.

Let's go with it, and keep an eye on issues for the corner case, and we can roll back if our instinct is wrong.

PR #154, opened 2024-08-01
Wait for the `Namespace` and `Account` informers to sync before controller startup

a-hilaly:
Add a new functionality to help waiting for the `Namespace` and `Account` cache
informers to sync before proceeding with the creation of the reconcilers.

Key changes:
- Publish `informer.HasSynced` to the top level structs for namespace and
  accout caches
- Use these `informer.HasSynced` to wait for the caches to sync using
  `"k8s.io/client-go/tools/cache"`.
- We call the wait mechanism right after the function that spins the informers
  a.k.a `caches.Run`.

Last to note, we are using a `context.TODO()` context, as a temporary workaround
until we figure out a better mechanism for context cancellation.

### DIFF
--- a/pkg/runtime/service_controller.go
+++ b/pkg/runtime/service_controller.go
@@ -14,6 +14,7 @@
 package runtime
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


pkg/runtime/service_controller.go
```diff
@@ -233,6 +234,10 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 		// Run the caches. This will not block as the caches are run in
 		// separate goroutines.
 		cache.Run(clientSet)
+		// Wait for the caches to sync
+		ctx := context.TODO()
```
michaelhtm, line 238: can we do 1 minute for this instead

a-hilaly: I think indeed this will solve the failing unit tests, but imposes risk of race condition

a-hilaly: maybe we can configure that specific failing unit test to not use CARM caches? it is only supposed to tests that the controller build happen correctly

a-hilaly: what's happening here is that: the unit tests attempt to initialize a cache, but `WaitForCachesToSync` requires communication with the api server, which is not running during unit tests. so... the function waits indefinitely an eventually leads to timeouts..

PR #160, opened 2024-09-03
Handling controller error when there is no default account ID

michaelhtm:
Issue [#2111](https://github.com/aws-controllers-k8s/community/issues/2111)

Description of changes:

Previously the controller was exiting with an error 
when there was no credentials provided by user, through
environment variables or the config files.

By not exiting the controller immediately, we allow the 
user to provide account IDs through the CARM features.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -257,6 +257,8 @@ func (r *resourceReconciler) Reconcile(ctx context.Context, req ctrlrt.Request) 
 		if err != nil {
 			return r.handleCacheError(ctx, err, desired)
 		}
+	} else if !needCARMLookup && acctID == "" {
+		return ctrlrt.Result{}, fmt.Errorf("unable to locate AWS credentials to manage resource: both controller credentials and a proper CARM configuration are missing")
 	}
 
 	region := r.getRegion(desired)
@@ -1100,7 +1102,7 @@ func (r *resourceReconciler) getOwnerAccountID(
 		return ackv1alpha1.AWSAccountID(accID), true
 	}
 
-	controllerAccountID := ackv1alpha1.AWSAccountID(r.cfg.AccountID)
+	controllerAccountID := ackv1alpha1.AWSAccountID(r.cfg.DefaultAccountID)
 	// look for owner account id in the resource status
 	acctID := res.Identifiers().OwnerAccountID()
 	if acctID != nil {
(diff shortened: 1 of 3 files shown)


pkg/runtime/reconciler.go
```diff
@@ -257,6 +257,8 @@ func (r *resourceReconciler) Reconcile(ctx context.Context, req ctrlrt.Request)
 		if err != nil {
 			return r.handleCacheError(ctx, err, desired)
 		}
+	} else if !needCARMLookup && acctID == "" {
+		return ctrlrt.Result{}, fmt.Errorf("no default account id nor account id in namespace found")
 	}
```
a-hilaly, line 262: Inexistant accountID is a symptom is a consequence of credentials not being provided. tiny suggestion:
```suggestion
	} else if !needCARMLookup && acctID == "" {
		return ctrlrt.Result{}, fmt.Errorf("unable to locate AWS credentials to manage resource: both controller credentials and a proper CARM configuration are missing")
	}
```

pkg/runtime/reconciler.go
```diff
@@ -1100,7 +1102,7 @@ func (r *resourceReconciler) getOwnerAccountID(
 		return ackv1alpha1.AWSAccountID(accID), true
 	}
 
-	controllerAccountID := ackv1alpha1.AWSAccountID(r.cfg.AccountID)
+	controllerAccountID := ackv1alpha1.AWSAccountID(r.cfg.DefailtAccountID)
 	// look for owner account id in the resource status
 	acctID := res.Identifiers().OwnerAccountID()
 	if acctID != nil {
```
a-hilaly, line 1108: This block is a little worrying:
```go
acctID := res.Identifiers().OwnerAccountID()
	if acctID != nil {
		return *acctID, *acctID != controllerAccountID
	}
```
I'If `controllerAccountID` is empty, we will return true, which doesn't match the logic in the main reconcile function.

a-hilaly (commented):
Thanks @michaelhtm ! I have two comments bellow.

@TiberiuGC this is interfering a little with CARM systems, can you please take a look?

pkg/runtime/reconciler.go
michaelhtm: I'm not sure what you mean here

a-hilaly: You have a change that depends on the boolean (*acctID != controllerAccountID):
```go
	} else if !needCARMLookup && acctID == "" {
		return ctrlrt.Result{}, fmt.Errorf("unable to locate AWS credentials to manage resource: both controller credentials and a proper CARM configuration are missing")
	}
```

if `*controllerAccountID = ""` then `*acctID != controllerAccountID` will be true which impacts the entry `!needCARMLookup && acctID == ""`

michaelhtm: what could the value of `accID` be in that case?

PR #164, opened 2024-11-14
Add addoption by annotation feature

michaelhtm:
Issue #, if available:

Description of changes:
These changes introduce a new feature gate called `ResourceAdoption` 
which allows users to provide the `Read` required fields in the annotation and 
an empty spec, and the controller would populate the resource and adopt 
from AWS

Currently we are considering two values for the `adoption-policy` annotation,
one being `adopt`, which will just try to adopt a resource, and if not found,
keep trying until a resource to adopt exists, or `adopt-or-create` which will
be able to create a resource if adoption fails because resource doesn't exist.

Before we support `adopt-or-create` we need a solution in code-gen
where we change how we handle the `PopulateResourceFromAnnotation`
which currently only allows the population of fields that are
required for a readOne operation, and they happen to be all scalar
fields (besides ARN, but we have a way of handling that), but the
required fields for create would need to be sometimes structs,
and this would require users to provide values in form of maps
eg.
Creating an EKS cluster requires a ResourceVPCConfig, which is a
struct that contains subnetIDs etc.
We can have a couple of ways to address this.
1. Accept these values in the spec, and return terminal error when
   we attempt a create and the create required fields are not provided
2. Accept these values in the `adoption-fields` annotation. This would
   need a code-gen change to allow reading from structs and assigning
   fields. but it would also make the annotation easy to make mistakes
   with when using yaml

Here's an example for cluster:
```yaml
apiVersion: eks.services.k8s.aws/v1alpha1
kind: Cluster
metadata:
  name: my-cluster
  annotations:
    services.k8s.aws/adoption-policy: "adopt"
    services.k8s.aws/adoption-fields: | 
        {
          "name": "my-cluster"
        }
```

Here's another one to adopt a nodegroup

```yaml
apiVersion: eks.services.k8s.aws/v1alpha1
kind: NodeGroup
metadata:
  name: my-ng
  annotations:
    services.k8s.aws/adoption-policy: "adopt"
    services.k8s.aws/adoption-fields: | 
        {
          "name": "ng-12324",
          "clusterName": "my-cluster"
        }
```

### DIFF
--- a/pkg/featuregate/features.go
+++ b/pkg/featuregate/features.go
@@ -19,9 +19,9 @@ package featuregate
 import "fmt"
 
 const (
-	// AdoptResources is a feature gate for enabling forced adoption of resources
+	// ResourceAdoption is a feature gate for enabling forced adoption of resources
 	// by annotation
-	AdoptResources = "AdoptResources"
+	ResourceAdoption = "ResourceAdoption"
 
 	// ReadOnlyResources is a feature gate for enabling ReadOnly resources annotation.
 	ReadOnlyResources = "ReadOnlyResources"
@@ -36,7 +36,7 @@ const (
 // defaultACKFeatureGates is a map of feature names to Feature structs
 // representing the default feature gates for ACK controllers.
 var defaultACKFeatureGates = FeatureGates{
-	AdoptResources:    {Stage: Alpha, Enabled: false},
+	ResourceAdoption:  {Stage: Alpha, Enabled: false},
 	ReadOnlyResources: {Stage: Alpha, Enabled: false},
 	TeamLevelCARM:     {Stage: Alpha, Enabled: false},
 	ServiceLevelCARM:  {Stage: Alpha, Enabled: false},
--- a/pkg/featuregate/features.go
+++ b/pkg/featuregate/features.go
@@ -19,9 +19,9 @@ package featuregate
 import "fmt"
 
 const (
-	// ForcedAdoptResources is a feature gate for enabling forced adoption of resources
+	// AdoptResources is a feature gate for enabling forced adoption of resources
 	// by annotation
-	ForcedAdoptResources = "ForcedAdoptResources"
+	AdoptResources = "AdoptResources"
 
 	// ReadOnlyResources is a feature gate for enabling ReadOnly resources annotation.
 	ReadOnlyResources = "ReadOnlyResources"
@@ -36,10 +36,10 @@ const (
 // defaultACKFeatureGates is a map of feature names to Feature structs
 // representing the default feature gates for ACK controllers.
 var defaultACKFeatureGates = FeatureGates{
-	ForcedAdoptResources: {Stage: Alpha, Enabled: false},
-	ReadOnlyResources:    {Stage: Alpha, Enabled: false},
-	TeamLevelCARM:        {Stage: Alpha, Enabled: false},
-	ServiceLevelCARM:     {Stage: Alpha, Enabled: false},
+	AdoptResources:    {Stage: Alpha, Enabled: false},
+	ReadOnlyResources: {Stage: Alpha, Enabled: false},
+	TeamLevelCARM:     {Stage: Alpha, Enabled: false},
+	ServiceLevelCARM:  {Stage: Alpha, Enabled: false},
 }
 
 // FeatureStage represents the development stage of a feature.
--- a/pkg/featuregate/features.go
+++ b/pkg/featuregate/features.go
@@ -19,6 +19,10 @@ package featuregate
 import "fmt"
 
 const (
+	// ForcedAdoptResources is a feature gate for enabling forced adoption of resources
+	// by annotation
+	ForcedAdoptResources = "ForcedAdoptResources"
+
 	// ReadOnlyResources is a feature gate for enabling ReadOnly resources annotation.
 	ReadOnlyResources = "ReadOnlyResources"
 
@@ -32,9 +36,10 @@ const (
 // defaultACKFeatureGates is a map of feature names to Feature structs
 // representing the default feature gates for ACK controllers.
 var defaultACKFeatureGates = FeatureGates{
-	ReadOnlyResources: {Stage: Alpha, Enabled: false},
-	TeamLevelCARM:     {Stage: Alpha, Enabled: false},
-	ServiceLevelCARM:  {Stage: Alpha, Enabled: false},
+	ForcedAdoptResources: {Stage: Alpha, Enabled: false},
+	ReadOnlyResources:    {Stage: Alpha, Enabled: false},
+	TeamLevelCARM:        {Stage: Alpha, Enabled: false},
+	ServiceLevelCARM:     {Stage: Alpha, Enabled: false},
 }
 
 // FeatureStage represents the development stage of a feature.
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -360,6 +360,81 @@ func (r *resourceReconciler) Sync(
 	isAdopted := IsAdopted(desired)
 	rlog.WithValues("is_adopted", isAdopted)
 
+	if r.cfg.FeatureGates.IsEnabled(featuregate.ForcedAdoptResources) {
+		
+		// If the resource is being adopted by force, we need to access
+		// the required field passed by annotation and attempt a read.
+		if NeedAdoption(desired) {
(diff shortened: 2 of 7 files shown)


pkg/featuregate/features.go
```diff
@@ -19,6 +19,10 @@ package featuregate
 import "fmt"
 
 const (
+	// ForcedAdoptResources is a feature gate for enabling forced adoption of resources
+	// by annotation
+	ForcedAdoptResources = "ForcedAdoptResources"
```
a-hilaly, line 24: It feels like "Forced" is a bit of a strong word, how about just `AdoptedResources`? the only thing is that it might bring confusion with the AdoptedResource CR

pkg/runtime/util.go
```diff
@@ -68,3 +69,50 @@ func IsReadOnly(res acktypes.AWSResource) bool {
 	}
 	return false
 }
+
+// IsForcedAdoption returns true if the supplied AWSResource has an annotation
+// indicating that it should be adopted
+func isForcedAdoption(res acktypes.AWSResource) bool {
```
a-hilaly, line 75: nit: `hasAdoptionAnnotations`

pkg/runtime/util.go
```diff
@@ -68,3 +69,50 @@ func IsReadOnly(res acktypes.AWSResource) bool {
...
+		}
+	}
+	return false
+}
+
+func NeedAdoption(res acktypes.AWSResource) bool {
+	return isForcedAdoption(res) && !IsAdopted(res)
+}
+
+func ExtractAdoptionFields(res acktypes.AWSResource) (map[string]string, error) {
```
a-hilaly, line 93: Do we need these two functions to be public?

pkg/runtime/util.go
```diff
@@ -68,3 +69,50 @@ func IsReadOnly(res acktypes.AWSResource) bool {
...
+	extractedFields := &map[string]string{}
+	err := json.Unmarshal([]byte(fields), extractedFields)
+	if err != nil {
+		return nil, err
+	}
+
+	return *extractedFields, nil
+}
+
+func extractFieldsFromAnnotation(res acktypes.AWSResource) string {
```
a-hilaly, line 105: This sounds like it's only getting the value of annotation. Shall we rename it to `getAdoptionFields` ?

pkg/runtime/reconciler.go
```diff
@@ -360,13 +360,90 @@ func (r *resourceReconciler) Sync(
...
+
 	if r.cfg.FeatureGates.IsEnabled(featuregate.ReadOnlyResources) {
 		isReadOnly := IsReadOnly(desired)
 		rlog.WithValues("is_read_only", isReadOnly)
 
 		// NOTE(a-hilaly): When the time comes to support adopting resources
 		// using annotations, we will need to think a little bit more about
 		// the case where a user, wants to adopt a resource as read-only.
+		//
+		// NOTE(michaelhtm): Done, tnx :)
```
a-hilaly, line 446: 🚀

pkg/runtime/reconciler.go
```diff
@@ -360,13 +360,90 @@ func (r *resourceReconciler) Sync(
 	isAdopted := IsAdopted(desired)
 	rlog.WithValues("is_adopted", isAdopted)
 
+	if r.cfg.FeatureGates.IsEnabled(featuregate.ForcedAdoptResources) {
+		
+		// If the resource is being adopted by force, we need to access
+		// the required field passed by annotation and attempt a read.
+		if NeedAdoption(desired) {
+			rlog.WithValues("is_forced_adoption", "true")
```
a-hilaly, line 368:
```suggestion
			rlog.WithValues("is_adopted", "true")
```

pkg/runtime/reconciler.go
```diff
@@ -360,13 +360,90 @@ func (r *resourceReconciler) Sync(
...
+			rlog.Info("Adopting Resource")
+			extractedFields, err := ExtractAdoptionFields(desired)
```
a-hilaly, line 370: To me this sounds like a user input issue, shall we return a terminal error?

pkg/runtime/util.go
```diff
@@ -68,3 +69,50 @@ func IsReadOnly(res acktypes.AWSResource) bool {
...
+}
+
+func NeedAdoption(res acktypes.AWSResource) bool {
+	return isForcedAdoption(res) && !IsAdopted(res)
+}
+
+func ExtractAdoptionFields(res acktypes.AWSResource) (map[string]string, error) {
+	fields := extractFieldsFromAnnotation(res)
+
+	extractedFields := &map[string]string{}
```
a-hilaly, line 96: Mmmmm how about nested fields? does it make sense to use `map[string]interface{}{}` instead? maybe in a follow up patch?

pkg/runtime/reconciler.go
```diff
@@ -360,13 +360,90 @@ func (r *resourceReconciler) Sync(
...
+			if err != nil {
+				return desired, err
+			}
+			if extractedFields == nil {
+				// TODO(michaelhtm) figure out error here
+				return nil, fmt.Errorf("Failed extracting fields from annotation")
+			}
```
a-hilaly, line 377: is there a case where users don't have to provide this annotation? should we instead check for `len(adoptionFields)` ?

pkg/runtime/reconciler.go
```diff
@@ -360,13 +360,90 @@ func (r *resourceReconciler) Sync(
...
+				rlog.Enter("rm.ResolveReferences")
+				resolved, hasReferences, err := rm.ResolveReferences(ctx, r.apiReader, res)
+				rlog.Exit("rm.ResolveReferences", err)
+				if err != nil {
+					return ackcondition.WithReferencesResolvedCondition(res, err), err
+				}
+				if hasReferences {
+					resolved = ackcondition.WithReferencesResolvedCondition(resolved, err)
+				}
+			}
```
a-hilaly, line 394: I don't think we should support this, for now... sounds like too early to solve this problem.

a-hilaly (commented):
Great work on this Mr Michael! i left a few comments below

pkg/types/aws_resource.go
```diff
@@ -48,4 +48,7 @@ type AWSResource interface {
 	SetStatus(AWSResource)
 	// DeepCopy will return a copy of the resource
 	DeepCopy() AWSResource
+	// PopulateResourceFromAnnotation will set the Spec or Status field that user
+	// provided from annotations
+	PopulateResourceFromAnnotation(fields map[string]string) error
```
adriananeci, line 53: Adding a new method to the `AWSResource` interface will invalidate all controllers `resource` structs that are implementing this interface. But I guess you were expecting this. 
I assume there are 2 ways for addressing it:
- PR can be merged even if PR checks associated with `iam`, `ecr`, etc controllers integration tests are failing. ACK controllers can be adjusted as a followup step when runtime version is bumped inside each controller.
- Add the `PopulateResourceFromAnnotation` method beforehand on all controllers and for all `resource` structs that are implementing the `AWSResource` interface

michaelhtm: Hey @adriananeci, thanks for the comment. We do have a PR pending for code-gen as well [#558](https://github.com/aws-controllers-k8s/code-generator/pull/558). We will attempt generating the controllers 
with those changes and test again.

pkg/runtime/reconciler.go
```diff
@@ -298,6 +302,70 @@ func (r *resourceReconciler) handleCacheError(
...
+	// Ensure tags again after adding the finalizer and patching the
+	// resource. Patching desired resource omits the controller tags
+	// because they are not persisted in etcd. So we again ensure
+	// that tags are present before performing the create operation.
+	rlog.Enter("rm.EnsureTags")
+	err = rm.EnsureTags(ctx, latest, r.sc.GetMetadata())
+	rlog.Exit("rm.EnsureTags", err)
+	if err != nil {
+		return latest, err
+	}
```
a-hilaly, line 357: Does this mean that the first reconciliation omits tags? then we eventually add them?

a-hilaly: Do we need to call `EnsureTags` at L333?

a-hilaly (approved):
Great work on this Michael! merging to see how tests are doing on code-gen
/lgtm

jlbutler:
closes https://github.com/aws-controllers-k8s/community/issues/1965

PR #167, opened 2025-01-08
Migrate to `aws-sdk-go-v2`

michaelhtm:
Description of changes:
* Creating Config instead of session
* Pass Config to controllers using rm/ManagerFor
* Regenerate mock tests with new Interfaces


pkg/runtime/config.go
```diff
@@ -0,0 +1,95 @@
...
+
+	awsCfg.AppID = val
+
+	if *endpointURL != "" {
+		awsCfg.BaseEndpoint = endpointURL
+	}
+
+	if roleARN != "" {
+		client := sts.NewFromConfig(awsCfg)
+		awsCfg.Credentials = stscreds.NewAssumeRoleProvider(client, string(roleARN))
```
cheeseandcereal, line 74: You will want to make sure to wrap this with a credentials cache, otherwise this will end up making an STS assume-role API call every time another API is called.

See example in sdk docs: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials/stscreds#hdr-Assume_Role

pkg/runtime/config.go
```diff
@@ -0,0 +1,95 @@
...
+
+	awsCfg, err := config.LoadDefaultConfig(
+		ctx,
+		config.WithRegion(string(region)),
+		config.WithAPIOptions([]func(*middleware.Stack) error{
+			userAgentAppender, 
+		}),
+	)
+
+	awsCfg.AppID = val
```
cheeseandcereal, line 66: This further changes the user agent header, but you're already fully setting a custom user agent with the middleware above, so this shouldn't be necessary

pkg/runtime/config.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return awsCfg, err
+	}
+
+	return awsCfg, nil
+}
+
+func (c *serviceController) getHandlerValue(groupVersionKind schema.GroupVersionKind) string {
+
+	val := fmt.Sprintf("%s/%s (%s) (%s) (%s) (%s)",
```
cheeseandcereal, line 86: I don't believe this is generating the same user-agent string as before. See the logic to generate this string that was being used before: https://github.com/aws/aws-sdk-go/blob/7112c0a0c2d01713a9db2d57f0e5722225baf5b5/aws/request/handlers.go#L307-L320

cheeseandcereal (changes requested):

pkg/runtime/config.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/aws/smithy-go/middleware"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+)
+
+const appName = "aws-controllers-k8s"
+
+func (c *serviceController) NewConfig(
```
a-hilaly, line 34: Two nits here:
- `runtime.NewSession` was unique in this package, now `runtime.NewConfig` can mean a lot of different things... can we name it `NewAWSConfig` or `NewSDKConfig` instead?
- need godoc

pkg/runtime/config.go
```diff
@@ -0,0 +1,94 @@
...
+	ctx context.Context,
+	region ackv1alpha1.AWSRegion,
+	endpointURL *string,
+	roleARN ackv1alpha1.AWSResourceName,
+	groupVersionKind schema.GroupVersionKind,
+) (aws.Config, error) {
```
a-hilaly, line 40: is there a reason we're not returning a ptr instead? most of the consumers require use it with a ptr in https://github.com/aws/aws-sdk-go-v2/blob/config/v1.29.1/aws/config.go

pkg/runtime/config.go
```diff
@@ -0,0 +1,94 @@
...
+
+	val := handlerValue(
+		appName,
+		groupVersionKind.Group+"-"+c.VersionInfo.GitVersion,
+		"BuildDate/" + c.VersionInfo.BuildDate,
```
a-hilaly, line 45: we're missing
```
"GitCommit/"+c.VersionInfo.GitCommit,
```
from the previous session.go code

pkg/runtime/config.go
```diff
@@ -0,0 +1,94 @@
...
+	return awsCfg, nil
+}
+
+func handlerValue(name, version string, extra ...string) string {
+	ua := fmt.Sprintf("%s/%s", name, version)
+	if len(extra) > 0 {
+		ua += fmt.Sprintf(" (%s)", strings.Join(extra, "; "))
+	}
+	return ua
+}
```
a-hilaly, line 94: nit: shall we get some unit tests for this function?

pkg/runtime/config.go
```diff
@@ -0,0 +1,94 @@
...
+	val := handlerValue(
+		appName,
+		groupVersionKind.Group+"-"+c.VersionInfo.GitVersion,
+		"BuildDate/" + c.VersionInfo.BuildDate,
+		"CRDKind/" + groupVersionKind.Kind,
+		"CRDVersion/" + groupVersionKind.Version,
+	)
+
+	userAgentAppender := func(stack *middleware.Stack) error {
+		return stack.Build.Add(middleware.BuildMiddlewareFunc(fmt.Sprintf("%s/user-agent", appName), func(
```
a-hilaly, line 51: qq: I see `appName` being used here AND in `handlerValue`, what's the difference? do we need both?

pkg/runtime/config.go
a-hilaly: nit: 
```suggestion
func formatUsentAgent(name, version string, extra ...string) string {
	ua := fmt.Sprintf("%s/%s", name, version)
	if len(extra) > 0 {
		ua += fmt.Sprintf(" (%s)", strings.Join(extra, "; "))
	}
	return ua
}
```

pkg/runtime/reconciler_test.go
```diff
@@ -127,7 +127,7 @@ func managerFactoryMocks(
 	rd.On("GroupVersionKind").Return(
 		schema.GroupVersionKind{
 			Group: "bookstore.services.k8s.aws",
-			Kind:  "fakeBook",
+			Kind:  "fakeBookx",
```
a-hilaly, line 130:
```suggestion
			Kind:  "fakeBook",
```

pkg/runtime/reconciler.go
```diff
@@ -441,7 +442,7 @@ func (r *resourceReconciler) Sync(
 				// TODO(michaelhtm): Change the handling of
 				// the error to allow Adopt or Create here
 				// when supported
-				return latest, err
+				return desired, err
```
a-hilaly, line 445: Is this change needed? Do we need to do in this PR?

pkg/runtime/reconciler.go
```diff
@@ -338,7 +339,7 @@ func (r *resourceReconciler) handleAdoption(
 	rlog.Enter("rm.ReadOne")
 	latest, err := rm.ReadOne(ctx, resolved)
 	if err != nil {
-		return latest, err
+		return nil, err
```
a-hilaly, line 342: ditto

a-hilaly (commented):
Good stuff @michaelhtm ! 👍 
left a few comments below

pkg/runtime/config.go
michaelhtm: yes, but when creating a client it accepts a non pointer.

pkg/runtime/config.go
michaelhtm: not sure..I mostly tried to make it look like the following from session.go:
```golang
	handlers.Build.PushFrontNamed(request.NamedHandler{
		Name: fmt.Sprintf("%s/user-agent", appName),
		Fn: request.MakeAddToUserAgentHandler(
			appName,
			groupVersionKind.Group+"-"+c.VersionInfo.GitVersion,
			"GitCommit/"+c.VersionInfo.GitCommit,
			"BuildDate/"+c.VersionInfo.BuildDate,
			"CRDKind/"+groupVersionKind.Kind,
			"CRDVersion/"+groupVersionKind.Version),
	})
```

pkg/runtime/reconciler.go
michaelhtm: we do not..reverting

a-hilaly (approved):
👍 
/lgtm

a-hilaly (commented):
/lgtm cancel

PR #170, opened 2025-02-10
feat(tags): filter AWS tags and ACK tags during reconciliation

michaelhtm:
Description of changes:
When we adopt ACK resources, if those resources are injected by AWS,
we are trying to avoid from patching them into the resource.
For this reason we are adding two new methods for the AWSResourceManager 
interface.

MirrowAWSTags:
This method ensures the AWS tags returned by the API after a read
operation are mirrored in the desired resource. Mirroring the tags in the desired
makes them undetected by the controller.

New FilterSystemTags Function:
Removes System tags injected by AWS (cloudformation aws:) 
or the controller (services.k8s.aws). This method is called during
adoption, to ensure these tags are not patched into the controller.

_____________________________
Also removing EnsureTag operations during adoption to avoid ACK from patching controller tags
to the resource.

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -13,10 +13,6 @@
 
 package tags
 
-import (
-	"strings"
-)
-
 // Tags represents the AWS tags which will be added to the AWS resource.
 // Inside aws-sdk-go, Tags are represented using multiple types, Ex: map of
 // string, list of structs etc...
@@ -50,31 +46,4 @@ func Merge(a Tags, b Tags) Tags {
 		}
 	}
 	return result
-}
-
-// SyncAWSTags ensures AWS-managed tags (prefixed with "aws:") from the latest resource state
-// are preserved in the desired state. This prevents the controller from attempting to
-// modify AWS-managed tags, which would result in an error.
-//
-// AWS-managed tags are automatically added by AWS services (e.g., CloudFormation, Service Catalog)
-// and cannot be modified or deleted through normal tag operations. Common examples include:
-// - aws:cloudformation:stack-name
-// - aws:servicecatalog:productArn
-//
-// Parameters:
-//   - a: The target Tags map to be updated (typically desired state)
-//   - b: The source Tags map containing AWS-managed tags (typically latest state)
-//
-// Example:
-//
-//	latest := Tags{"aws:cloudformation:stack-name": "my-stack", "environment": "prod"}
-//	desired := Tags{"environment": "dev"}
-//	SyncAWSTags(desired, latest)
-//	desired now contains {"aws:cloudformation:stack-name": "my-stack", "environment": "dev"}
-func SyncAWSTags(a Tags, b Tags) {
-	for k := range b {
-		if strings.HasPrefix(k, "aws:") {
-			a[k] = b[k]
-		}
-	}
 }
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -13,6 +13,10 @@
 
 package tags
 
+import (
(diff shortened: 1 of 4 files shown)


pkg/types/aws_resource_manager.go
```diff
@@ -85,6 +85,8 @@ type AWSResourceManager interface {
 	// If the AWSResource does not support tags, only then the controller tags
 	// will not be added to the AWSResource.
 	EnsureTags(context.Context, AWSResource, ServiceControllerMetadata) error
+
+	FilterAWSTags(AWSResource)
```
a-hilaly, line 89: Please document the interface methods, it's good for our godocs :)

pkg/tags/tags.go
```diff
@@ -47,3 +51,29 @@ func Merge(a Tags, b Tags) Tags {
...
+// removed by an ACK controller
+func IgnoreAWSTags(tags Tags) {
+	for k := range tags {
+		if strings.HasPrefix(k, "aws:") {
+			delete(tags, k)
+		}
+	}
+}
+
+// IgnoreAWSTags ignores tags that have keys that start with "aws:"
```
a-hilaly, line 68:
```suggestion
// SyncAWSTags ignores tags that have keys that start with "aws:"
```

a-hilaly: Is this a sync? or a copy?

a-hilaly: This code can panic if `a` is not pre-allocated

a-hilaly (commented):
Thank you @michaelhtm ! left a few comments in-line

pkg/tags/tags.go
michaelhtm: IT's sync..we make sure both have AWS: tags...updating the comments now.
the preallocation happens in the controller https://github.com/aws-controllers-k8s/s3-controller/pull/149/files#diff-25ef9137262ccb16752e6f45e181294b0897f972fa85387b635e62a80dbc6ec9R333-R353

pkg/types/aws_resource_manager.go
```diff
@@ -85,6 +85,25 @@ type AWSResourceManager interface {
...
+	// aws tags are not present in a diff. The logic of the controller will
+	// ensure these tags aren't patched to the resource in the cluster, and
+	// will only be present to make sure we don't try to remove these tags.
+	//
+	// Although there are a lot of similarities between this function and
+	// EnsureTags, they are very much different.
+	// While EnsureTags tries to make sure the resource contains the controller
+	// tags, mirrowAWSTags tries to make sure tags injected by AWS are mirrored
+	// from the latest resoruce to the desired resource.
+	MirrorAWSTags(resA AWSResource, resB AWSResource)
```
a-hilaly, line 99: nit: for clarity
```suggestion
	MirrorAWSTags(desired AWSResource, latest AWSResource)
```

pkg/types/aws_resource_manager.go
```diff
@@ -85,6 +85,25 @@ type AWSResourceManager interface {
...
+	// FilterAWSTags ignores tags that have keys that start with "aws:"
+	// is needed to ensure the controller does not attempt to remove
+	// tags set by AWS. This function needs to be called after each Read
+	// operation.
+	// Eg. resources created with cloudformation have tags that cannot be
+	//removed by an ACK controller
+	FilterAWSTags(resA AWSResource)
```
a-hilaly, line 106: is this function intended to be called for desired & latest or only one of them? should we document that?

a-hilaly (approved):
👍

PR #175, opened 2025-03-27
feat: Add label-based object filtering functionality

a-hilaly:
Introduces label selectors to filter which objects controllers watch, reducing resource usage
and improving performance in large clusters by allowing controllers to focus only on objects
with specific labels instead of processing all resources.


rushmash91 (commented):
thank you @itaiatu and @a-hilaly !!
/lgtm

PR #177, opened 2025-04-03
update: Update the `Resource already exists` condition reason

michaelhtm:
Description of changes:
Currently the condition reason for the `Resource Already Exists` 
condition message instructs users to adopt the existing resource 
using the `AdoptResource` CRD, which we no longer implement.

This change updates the message to instruct users to instead use 
the adoption by annotation feature to import an existing resource.


a-hilaly (approved):
Thanks @michaelhtm !
/lgtm

a-hilaly:
@michaelhtm these test look super flaky, are they related to the previous ec2 change?

michaelhtm:
@a-hilaly the failing tests are mostly due to eventual consistency..I noticed the more tests we added the worse it got..maybe we can restructure it better

PR #179, opened 2025-04-04
feat: Adopt-or-Create

michaelhtm:
Issue [#2406](https://github.com/aws-controllers-k8s/community/issues/2406)

Description of changes:
The changes handle the addition of a new `adoption-policy` called `adopt-or-create`.
The adopt-or-create policy requires users to define the Spec fields in the resource Spec,
and the Status fields in the `adoption-fields` annotation. Requiring the Spec fields be in Spec
for adopt-or-create ensures that the users provide fields that may also be required for a create
operation. 

Here is the controller lifecycle of adopt-or-create policy:
```
Controller looks for resource in AWS
If found
   check if there are any differences between user resource definition (desired) and 
    resource definition in AWS (latest). Update if there are any differences
If not Found
  create the resource
```
Unlike the `adopt` adoption policy, if a resource has `adopt-or-create` policy and `read-only`, and it is not found 
in AWS, the controller will mark that resource with a terminal error.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -303,54 +303,29 @@ func (r *resourceReconciler) handleCacheError(
 	return r.HandleReconcileError(ctx, desired, latest, requeue.NeededAfter(err, roleARNNotAvailableRequeueDelay))
 }
 
-func (r *resourceReconciler) handleAdoption(
-	ctx context.Context,
-	rm acktypes.AWSResourceManager,
-	desired acktypes.AWSResource,
+func (r *resourceReconciler) handlePopulation(
 	rlog acktypes.Logger,
+	desired acktypes.AWSResource,
 ) (acktypes.AWSResource, error) {
 	// If the resource is being adopted by force, we need to access
 	// the required field passed by annotation and attempt a read.
 
 	rlog.Info("Adopting Resource")
-	extractedFields, err := ExtractAdoptionFields(desired)
+	adoptionFields, err := ExtractAdoptionFields(desired)
 	if err != nil {
 		return desired, ackerr.NewTerminalError(err)
 	}
-	if len(extractedFields) == 0 {
-		// TODO(michaelhtm) Here we need to figure out if we want to have an
-		// error or not. should we consider accepting values from Spec?
-		// And then we can let the ReadOne figure out if we have missing
-		// required fields for a Read
-		return nil, fmt.Errorf("failed extracting fields from annotation")
-	}
-	resolved := desired.DeepCopy()
-	err = resolved.PopulateResourceFromAnnotation(extractedFields)
-	if err != nil {
+
+	populated := desired.DeepCopy()
+	err = populated.PopulateResourceFromAnnotation(adoptionFields)
+	// maybe don't return errors when it's adopt-or-create?
+	// TODO (michaelhtm) change PopulateResourceFromAnnotation to understand
+	// adopt-or-create, and validate Spec fields are not nil...
+	if err != nil && GetAdoptionPolicy(desired) != "adopt-or-create" {
 		return nil, err
 	}
 
-	rlog.Enter("rm.ReadOne")
-	latest, err := rm.ReadOne(ctx, resolved)
-	rlog.Exit("rm.ReadOne", err)
-	rm.FilterSystemTags(latest)
-	if err != nil {
-		return latest, err
-	}
-
-	if err = r.setResourceManaged(ctx, rm, latest); err != nil {
-		return latest, err
-	}
-
-	r.rd.MarkAdopted(latest)
-	rlog.WithValues("is_adopted", "true")
-	latest, err = r.patchResourceMetadataAndSpec(ctx, rm, desired, latest)
-	if err != nil {
-		return latest, err
-	}
-
-	rlog.Info("Resource Adopted")
-	return latest, nil
+	return populated, nil
 }
 
 // reconcile either cleans up a deleted resource or ensures that the supplied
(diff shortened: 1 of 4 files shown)


gfaraj:
Hi @michaelhtm! Quick question -- will this overwrite the entire Kubernetes manifest if it's able to adopt an existing resource? From the code I'm seeing, it seems like that's the case.

In my case, I would like to avoid that if possible. Overwriting the manifest from the adopted resource means I have to hit "sync" again for my argo app. Really, I just want to upsert my resource to match my ACK spec.

While I think overwriting the whole thing does seem like a valid use case, maybe it's possible to support both cases? Might be looking at a separate adoption policy of "upsert" or something like that?

michaelhtm:
@gfaraj Hello, The entire manifest would be replaced with the spec defined in AWS if the `adoption-policy` is `adopt`. With the current implementation, if it's `adopt-or-create`, the controller will do an update after the adoption, if not it will create the resource. Maybe we need a better name than `adopt-or-create`..

gfaraj:
@michaelhtm oh nice, thanks for confirming that, I may have misread part of the changes. Yeah then this seems more like an "upsert" -- maybe that should be the policy name instead?

pkg/runtime/reconciler.go
```diff
@@ -303,54 +303,33 @@ func (r *resourceReconciler) handleCacheError(
 	return r.HandleReconcileError(ctx, desired, latest, requeue.NeededAfter(err, roleARNNotAvailableRequeueDelay))
 }
 
-func (r *resourceReconciler) handleAdoption(
-	ctx context.Context,
-	rm acktypes.AWSResourceManager,
-	desired acktypes.AWSResource,
+func (r *resourceReconciler) handlePopulation(
```
a-hilaly, line 306: interesting rename, can you please add a comment explaining what is the purpose (and non-purpose) of this function?

pkg/runtime/reconciler.go
```diff
@@ -303,54 +303,33 @@ func (r *resourceReconciler) handleCacheError(
...
 	rlog acktypes.Logger,
```
a-hilaly, line 307: can we get the logger from the context instead? (following the existing pattern)

pkg/runtime/reconciler.go
```diff
@@ -303,54 +303,33 @@ func (r *resourceReconciler) handleCacheError(
...
+	desired acktypes.AWSResource,
+	adoptionPolicy AdoptionPolicy,
```
a-hilaly, line 309: can you get the adoption policy from the desired object?

pkg/runtime/reconciler.go
```diff
@@ -303,54 +303,33 @@ func (r *resourceReconciler) handleCacheError(
...
 ) (acktypes.AWSResource, error) {
 	// If the resource is being adopted by force, we need to access
 	// the required field passed by annotation and attempt a read.
 
 	rlog.Info("Adopting Resource")
-	extractedFields, err := ExtractAdoptionFields(desired)
+	adoptionFields, err := ExtractAdoptionFields(desired)
```
a-hilaly, line 315: does `ExtractAdoptionFields` need to be public?

pkg/runtime/reconciler.go
```diff
@@ -303,54 +303,33 @@ func (r *resourceReconciler) handleCacheError(
...
-	r.rd.MarkAdopted(latest)
-	rlog.WithValues("is_adopted", "true")
-	latest, err = r.patchResourceMetadataAndSpec(ctx, rm, desired, latest)
-	if err != nil {
-		return latest, err
+	populated := desired.DeepCopy()
+	err = populated.PopulateResourceFromAnnotation(adoptionFields)
+	// maybe don't return errors when it's adopt-or-create?
+	// TODO (michaelhtm) change PopulateResourceFromAnnotation to understand
+	// adopt-or-create, and validate Spec fields are not nil...
```
a-hilaly, line 327: yes 💯

a-hilaly: or maybe `PopulateResourceFromAnnotation` shouldn't return an error? how about a boolean, stating whether a resource got populated or not.

pkg/runtime/reconciler.go
```diff
@@ -419,92 +398,53 @@ func (r *resourceReconciler) Sync(
...
-				// If we get an error, we want to return here
-				// TODO(michaelhtm): Change the handling of
-				// the error to allow Adopt or Create here
-				// when supported
-				return latest, err
-			}
-			return latest, nil
-		}
+	// find out if we need to adopt early on
+	needAdoption := NeedAdoption(desired) && !r.rd.IsManaged(desired) && r.cfg.FeatureGates.IsEnabled(featuregate.ResourceAdoption)
```
a-hilaly, line 402:
```suggestion
    // If AdoptionAnnotation feature gate is enabled, and the resource has an adoption
    // annotation and doesn't hold an ACK finalizer, trigger the adoption path.
	needAdoption := r.cfg.FeatureGates.IsEnabled(featuregate.ResourceAdoption) && NeedAdoption(desired) && !r.rd.IsManaged(desired)
```

pkg/runtime/util.go
```diff
@@ -24,6 +25,20 @@ import (
 )
 
 // TODO(jaypipes): Place this code somewhere separate
+//     (michaelhtm)             ^ +1
+
+// AdoptionPolicy stores adoptionPolicy values we expect users to
+// provide in the resources `adoption-policy` annotation
+// TODO(michaelhtm) Maybe we need a different place for this...
+// next refactor maybe? 🤷‍♂️
```
a-hilaly, line 33:
```suggestion
// AdoptionPolicy stores adoptionPolicy values we expect users to
// provide in the resources `adoption-policy` annotation
//
// TODO(michaelhtm) Maybe we need a different place for this...
// next refactor maybe? 🤷‍♂️
```

a-hilaly: An yes, i think you can move a lot of this logic to an `adoption.go` file

pkg/runtime/util.go
```diff
@@ -24,6 +25,20 @@ import (
...
+type AdoptionPolicy string
+
+const (
+	AdoptPolicy AdoptionPolicy = "adopt"
+	// calling this policy constant `AdoptOrCreatePolicy`
+	//  and see if we need to rename it!!!
+	AdoptOrCreatePolicy AdoptionPolicy = "adopt-or-create"
+)
```
a-hilaly, line 41:
```suggestion
const (
    // AdoptPolicy is ...
	AdoptPolicy AdoptionPolicy = "adopt"
	// AdoptOrCreatePolicy is ...
	AdoptOrCreatePolicy AdoptionPolicy = "adopt-or-create"
)
```

pkg/runtime/reconciler_test.go
```diff
@@ -262,6 +262,205 @@ func TestReconcilerReadOnlyResource(t *testing.T) {
...
+
+	desired, _, metaObj := resourceMocks()
+	desired.On("ReplaceConditions", []*ackv1alpha1.Condition{}).Return()
+	metaObj.SetAnnotations(map[string]string{
+		ackv1alpha1.AnnotationAdoptionPolicy: "adopt-or-create",
+		ackv1alpha1.AnnotationAdoptionFields: "{\"arn\": \"my-adopt-book-arn\"}",
+	})
+
+	ids := &ackmocks.AWSResourceIdentifiers{}
+	// ids.On("ARN").Return(&arn)
```
a-hilaly, line 407:
```suggestion
```

pkg/runtime/reconciler_test.go
```diff
@@ -262,6 +262,205 @@ func TestReconcilerReadOnlyResource(t *testing.T) {
...
```
a-hilaly, line 337:
```suggestion
```

pkg/runtime/reconciler_test.go
```diff
@@ -262,6 +262,205 @@ func TestReconcilerReadOnlyResource(t *testing.T) {
...
+	rm.AssertNotCalled(t, "Create", 0)
+	rm.AssertNotCalled(t, "Update", 0)
+	rm.AssertNotCalled(t, "Delta", 0)
+}
+
+func TestReconcilerAdoptOrCreateResource_Create(t *testing.T) {
+	require := require.New(t)
+
+	ctx := context.TODO()
+	// arn := ackv1alpha1.AWSResourceName("my-adopt-book-arn")
```
a-hilaly, line 327:
```suggestion
```

pkg/runtime/reconciler.go
```diff
@@ -514,13 +454,30 @@ func (r *resourceReconciler) Sync(
 		if err != ackerr.NotFound {
 			return latest, err
 		}
-		if isAdopted {
+		if adoptionPolicy == AdoptPolicy || isAdopted {
```
a-hilaly, line 457: do you also want to check for `needAdoption`

pkg/runtime/reconciler.go
```diff
@@ -514,13 +454,30 @@ func (r *resourceReconciler) Sync(
...
 			return nil, ackerr.AdoptedResourceNotFound
 		}
+		if isReadOnly {
+			return nil, ackerr.ReadOnlyResourceNotFound
+		}
 		if latest, err = r.createResource(ctx, rm, resolved); err != nil {
 			return latest, err
 		}
-	} else {
+	} else if adoptionPolicy == AdoptPolicy {
```
a-hilaly, line 466: ditto

a-hilaly (commented):
Thanks @michaelhtm ! Looks good overall, I left a few comments below

pkg/runtime/reconciler.go
a-hilaly, line 1: nit: maybe it's about time to factor the Sync function in multiple small functions? maybe in a seperate PR? looks like it's growing exponentially, and most of it's logic is really the core of ACK.

pkg/runtime/reconciler.go
michaelhtm: not really. privatizing...

pkg/runtime/reconciler.go
michaelhtm: i check that early on..`adoptionPolicy = ""` if `!needAdoption`

pkg/runtime/reconciler.go
michaelhtm: ➕1

pkg/runtime/reconciler.go
michaelhtm: Actually there's more functions in `util.go` that are public without needing to be..we can change their signature during the next refactor..

pkg/runtime/reconciler.go
michaelhtm: For this, i was mostly thinking we can pass the `adoptionPolicy` to `PopulateResourceFromAnnotation`. Doing so, we can decide not to populate Spec fields if the policy is Adopt-Or-Create...

michaelhtm: Or..we can divide this up into two..PopulateSpecFromAnnotations, and PopulateStatusFromAnnotation..(we need a better name)

rushmash91 (commented):
Nice @michaelhtm !! would be great to have more details in the PR description.

pkg/runtime/reconciler.go
michaelhtm: ditto

PR #181, opened 2025-05-06
fix: ensure Status fields are populated!

michaelhtm:
Issue [#2459](https://github.com/aws-controllers-k8s/community/issues/2459)

Description of changes:
When we have adopt-or-create, the populated status was not 
getting patched to k8s. So, during the next reconcile, the controller 
attempts to recreate the resource.

With this fix, we ensure that the controller does patch the resource with 
the adoption field!


a-hilaly (approved):
👍

PR #182, opened 2025-05-07
fix: mark resource as adopted

michaelhtm:
Issue #, if available:

Description of changes:
Ensure resource is marked as adopted after successful adoption


pkg/runtime/reconciler.go
```diff
@@ -475,6 +475,9 @@ func (r *resourceReconciler) Sync(
 		if isReadOnly {
 			return nil, ackerr.ReadOnlyResourceNotFound
 		}
+		if adoptionPolicy == AdoptionPolicy_AdoptOrCreate {
+			r.rd.MarkAdopted(latest)
+		}
```
a-hilaly, line 480: this is a question not a suggestion, what do you think the code below? any edge cases where it wouldn't work?
```
defer func() {
    if adoptionpolicy == adopt || adoptOrCreate {
        r.rd.MarkAdopt
    }
}
```

michaelhtm: Yes, mostly errors. If there's an error for nay reason, the controller won't attempt to adopt during the next reconciliation.

michaelhtm: Also, for adopt-or-create, i'm thinking of removing this portion, and not mark the resource as adopted if we're going to create it..

michaelhtm:
/test ec2-controller-test
--

PR #185, opened 2025-05-13
fix: Remove Finalizer from resource if create returns error

michaelhtm:
Issue [#2406#issuecomment-2860631351](https://github.com/aws-controllers-k8s/community/issues/2406#issuecomment-2860631351)

Description of changes:
The reason marking a resource as managed (putting the finalizer)
before attempting a create is a general practice in kubernetes. The main reason we 
do it is to protect against deletion protection.
If we don't put the finalizer, there is no deletion protection against the resource.

The current adoption logic expects the resource to not be managed 
(no finalizer) to trigger an adoption.
If the initial creation attempt of a resource fails due to any AWS 
error, any subsequent reconciliations attempting to adopt an existing 
resource will not succeed.

These changes set the resource as unmanaged if for any reason there is
an error during the create call, which will allow the adoption logic to 
run in subsequent reconciliations.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -636,6 +636,16 @@ func (r *resourceReconciler) createResource(
 	latest, err = rm.Create(ctx, desired)
 	rlog.Exit("rm.Create", err)
 	if err != nil {
+		// Here we're deciding to set a resource as unmanaged
+		// if the error is an AWS API Error. This will ensure
+		// that we're only managing (put finalizer) the resources
+		// that actually exist in AWS.
+		if _, ok := ackerr.AWSError(err); ok {
+			mErr := r.setResourceUnmanaged(ctx, rm, desired)
+			if mErr != nil {
+				return latest, err
+			}
+		}
 		return latest, err
 	}
 
--- a/pkg/runtime/reconciler_test.go
+++ b/pkg/runtime/reconciler_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/smithy-go"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


pkg/runtime/reconciler.go
```diff
@@ -638,6 +638,9 @@ func (r *resourceReconciler) createResource(
 	latest, err = rm.Create(ctx, desired)
 	rlog.Exit("rm.Create", err)
 	if err != nil {
+		if _, ok := ackerr.AWSError(err); ok {
+			r.setResourceUnmanaged(ctx, rm, desired)
```
a-hilaly, line 642: do you want to handle the error here?

michaelhtm: You're right! addressed

knottnt (commented):
It would be good to have a new test case for this new behavior in reconciler_test.go.

michaelhtm:
Thanks @knottnt 
Waiting on https://github.com/aws-controllers-k8s/runtime/pull/187 to be merged so i can add tests!

michaelhtm:
@knottnt 
Just added a test

pkg/runtime/reconciler_test.go
```diff
@@ -224,6 +225,63 @@ func TestReconcilerCreate_BackoffRetries(t *testing.T) {
...
+	rd.On("ResourceFromRuntimeObject", desiredRTObj).Return(desired)
+	rd.On("Delta", desired, desired).Return(ackcompare.NewDelta())
+
+	r, kc, scmd := reconcilerMocks(rmf)
+	rm.On("EnsureTags", ctx, desired, scmd).Return(nil)
+	// Use specific matcher for WithoutCancel context instead of mock.Anything
+	kc.On("Patch", withoutCancelContextMatcher, latestRTObj, mock.AnythingOfType("*client.mergeFromPatch")).Return(nil)
+	_, err := r.Sync(ctx, rm, desired)
+	require.NotNil(err)
+	rm.AssertNumberOfCalls(t, "ReadOne", 1)
```
knottnt, line 282: Should we also assert that `setResourceUnmanaged` was called?

PR #186, opened 2025-05-14
fix: ensure we mark resource as adopted adopt-or-create

michaelhtm:
Issue [#2459#issuecomment-2879691276](https://github.com/aws-controllers-k8s/community/issues/2459#issuecomment-2879691276)

Description of changes:
Currently the `adopted` annotation was getting removed when the resource
goes through update path during adopt-or-create adoption. These changes
ensure the annotation is persisted in k8s.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -732,8 +732,7 @@ func (r *resourceReconciler) updateResource(
 	rm acktypes.AWSResourceManager,
 	desired acktypes.AWSResource,
 	latest acktypes.AWSResource,
-) (acktypes.AWSResource, error) {
-	var err error
+) (updated acktypes.AWSResource, err error) {
 	rlog := ackrtlog.FromContext(ctx)
 	exit := rlog.Trace("r.updateResource")
 	defer func() {
@@ -754,21 +753,24 @@ func (r *resourceReconciler) updateResource(
 			"diff", delta.Differences,
 		)
 		rlog.Enter("rm.Update")
-		latest, err = rm.Update(ctx, desired, latest, delta)
+		updated, err = rm.Update(ctx, desired, latest, delta)
 		rlog.Exit("rm.Update", err, "latest", latest)
 		if err != nil {
-			return latest, err
+			return updated, err
 		}
 		// Ensure that we are patching any changes to the annotations/metadata and
 		// the Spec that may have been set by the resource manager's successful
 		// Update call above.
-		latest, err = r.patchResourceMetadataAndSpec(ctx, rm, desired, latest)
+		if IsAdopted(latest) {
+			r.rd.MarkAdopted(updated)
+		}
+		updated, err = r.patchResourceMetadataAndSpec(ctx, rm, desired, updated)
 		if err != nil {
-			return latest, err
+			return updated, err
 		}
 		rlog.Info("updated resource")
 	}
-	return latest, nil
+	return updated, nil
 }
 
 // lateInitializeResource calls AWSResourceManager.LateInitialize() method and


pkg/runtime/reconciler.go
```diff
@@ -501,6 +503,7 @@ func (r *resourceReconciler) Sync(
 			return latest, err
 		}
 	}
+	fmt.Println(latest)
```
a-hilaly, line 506: was this really helpful

pkg/runtime/reconciler.go
```diff
@@ -730,13 +730,13 @@ func (r *resourceReconciler) updateResource(
...
+) (updated acktypes.AWSResource, err error) {
 	rlog := ackrtlog.FromContext(ctx)
 	exit := rlog.Trace("r.updateResource")
 	defer func() {
 		exit(err)
 	}()
+	updated = latest
 
 	// Ensure the resource is managed
 	if err = r.failOnResourceUnmanaged(ctx, latest); err != nil {
```
knottnt, line 742: Should this if block use `updated` instead of `latest` as well?

pkg/runtime/reconciler.go
```diff
@@ -752,21 +752,24 @@ func (r *resourceReconciler) updateResource(
...
+			return updated, err
 		}
 		// Ensure that we are patching any changes to the annotations/metadata and
 		// the Spec that may have been set by the resource manager's successful
 		// Update call above.
-		latest, err = r.patchResourceMetadataAndSpec(ctx, rm, desired, latest)
+		if IsAdopted(latest) {
+			r.rd.MarkAdopted(updated)
+		}
+		updated, err = r.patchResourceMetadataAndSpec(ctx, rm, desired, updated)
```
knottnt, line 766: It would be good to have a test verifying that this works and that future changes don't cause a regression.

pkg/runtime/reconciler.go
knottnt: I was thinking that we'd changed latest -> updated in most of this function. But looking closer I can see that this is only the case after `rm.Update` is called. I think you're right it should be left as-is.

pkg/runtime/reconciler_test.go
```diff
@@ -454,36 +481,50 @@ func TestReconcilerAdoptOrCreateResource_Adopt(t *testing.T) {
...
 	kc.On("Patch", withoutCancelContextMatcher, latestRTObj, mock.AnythingOfType("*client.mergeFromPatch")).Return(nil)
-	statusWriter.On("Patch", withoutCancelContextMatcher, latestRTObj, mock.AnythingOfType("*client.mergeFromPatch")).Return(nil)
+	kc.On("Patch", withoutCancelContextMatcher, updatedRTObj, mock.AnythingOfType("*client.mergeFromPatch")).Return(nil)
+	statusWriter.On("Patch", withoutCancelContextMatcher, updatedRTObj, mock.AnythingOfType("*client.mergeFromPatch")).Return(nil)
 	_, err := r.Sync(ctx, rm, desired)
 	require.Nil(err)
 	rm.AssertNumberOfCalls(t, "ReadOne", 1)
 	rm.AssertCalled(t, "Update", ctx, desired, latest, delta)
 	rd.AssertCalled(t, "Delta", desired, latest)
+	rd.AssertNumberOfCalls(t, "MarkAdopted", 2)
```
knottnt, line 527: Can we assert that the AwsResource has the expected adoption fields as well when it is patched? Checking that the function call happened twice is a good indication, but we could be more definitive.

michaelhtm: When MarkAdopted is called (an interface method) we're not really doing anything in these unit tests. We can only trust the actual implementation is handling the MarkAdopted is correct. The best way to actually test this is using e2e tests

knottnt: Ahh I see. That's due to the implementation being generated per service controller?

knottnt:
/lgtm
/hold

Just holding in case we want a second reviewer.

pkg/runtime/reconciler_test.go
michaelhtm: Yes

PR #188, opened 2025-05-22
feat: wait for resource to sync before updating

michaelhtm:
Description of changes:
Currently, We've been setting a resource as NotSynced in ReadOne, if the
resource status is not a certain way (Eg. waiting for EKS Cluster to be
Active), and also checking the same resource Status in sdkUpdate.

This is the first attempt at addressing that issue. We are going to need
to ensure every controller/resource returns the proper NotSynced reason
and the requeue error inside the condition. This will allow us to set a
custom requeue right from ReadOne, that will be returned in preUpdate.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -745,6 +745,13 @@ func (r *resourceReconciler) updateResource(
 		return latest, err
 	}
 
+	// We don't want to attempt updating the resource until it is Synced!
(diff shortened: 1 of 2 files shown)


knottnt (commented):
It would be nice if we included some integration and unit tests.

michaelhtm:
This change is no longer needed (it's being addressed using Updatable in code-gen)

PR #191, opened 2025-06-02
Fix controller shutdown status patching with a fresh context for patches

rushmash91:
Fixes: https://github.com/aws-controllers-k8s/community/issues/2497, https://github.com/aws-controllers-k8s/community/issues/2103

**Description of changes:**

During controller shutdown (SIGTERM), status updates were failing because `patchResourceStatus` was attempting to patch using the original cancelled context. This resulted in resources not having their final status properly updated in the resource manifest, leaving them in an inconsistent state.

**FIX**: Always use `context.WithoutCancel(ctx)` for all patch operations to prevent cancellation propagation while preserving context values. The 30s SIGTERM grace period serves as the effective timeout, eliminating the need for additional timeout logic that could interfere with normal Kubernetes client timeout/retry strategy.

**Changes:**

Created util functions:
```
- patchWithoutCancel()
- patchStatusWithoutCancel() 
```

**Operations updated to Use` context.WithoutCancel(ctx)`**
Main Reconciler:
```
- patchResourceMetadataAndSpec()
- patchResourceStatus()

```
Adoption Reconciler:
```
- patchMetadataAndSpec()
- patchStatus()

```
Field Export Reconciler:
```
- patchMetadataAndSpec()
- patchStatus()
- writeToConfigMap()
- writeToSecret()
```

eg logs: 
```
{"level":"debug","ts":"2025-06-05T09:52:16.723-0700","logger":"ackrt","msg":"< r.Sync","kind":"Policy","namespace":"default","name":"iiobalic90","account":"xx","role":"","region":"us-west-2","is_adopted":false,"generation":1,"error":"operation error IAM: GetPolicy, context canceled"}
{"level":"info","ts":"2025-06-05T09:52:16.723-0700","logger":"ackrt","msg":"created patch context with timeout","kind":"Policy","namespace":"default","name":"iiobalic90","account":"xx","role":"","region":"us-west-2","is_adopted":false,"generation":1,"timeout_seconds":10}
{"level":"debug","ts":"2025-06-05T09:52:16.723-0700","logger":"ackrt","msg":"> r.patchResourceStatus","kind":"Policy","namespace":"default","name":"iiobalic90","account":"xx","role":"","region":"us-west-2","is_adopted":false,"generation":1}
```

### DIFF
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -576,17 +576,7 @@ func (r *adoptionReconciler) patchMetadataAndSpec(
 	// Keep a copy of status field to reset the status of 'res' after patch call
 	resStatusCopy := res.DeepCopy().Status
 
-	// Always use WithoutCancel to prevent patch operations from being
-	// cancelled while preserving context values. The 30s SIGTERM grace period acts as
-	// the effective timeout - no additional timeout needed to avoid interfering with
-	// normal Kubernetes client timeout/retry strategy.
-	patchCtx := context.WithoutCancel(ctx)
-
-	err := r.kc.Patch(
-		patchCtx,
-		res,
-		client.MergeFrom(base),
-	)
+	err := patchWithoutCancel(ctx, r.kc, res, client.MergeFrom(base))
 	res.Status = resStatusCopy
 	return err
 }
@@ -600,17 +590,7 @@ func (r *adoptionReconciler) patchStatus(
 	res *ackv1alpha1.AdoptedResource,
 	base *ackv1alpha1.AdoptedResource,
 ) error {
-	// Always use WithoutCancel to prevent patch operations from being
-	// cancelled while preserving context values. The 30s SIGTERM grace period acts as
-	// the effective timeout - no additional timeout needed to avoid interfering with
-	// normal Kubernetes client timeout/retry strategy.
-	patchCtx := context.WithoutCancel(ctx)
-
-	return r.kc.Status().Patch(
-		patchCtx,
-		res,
-		client.MergeFrom(base),
-	)
+	return patchStatusWithoutCancel(ctx, r.kc, res, client.MergeFrom(base))
 }
 
 // NewAdoptionReconciler returns a new adoptionReconciler object
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -575,8 +575,15 @@ func (r *adoptionReconciler) patchMetadataAndSpec(
 	// content returned from apiserver.
 	// Keep a copy of status field to reset the status of 'res' after patch call
 	resStatusCopy := res.DeepCopy().Status
+
+	// Always use WithoutCancel to prevent patch operations from being
+	// cancelled while preserving context values. The 30s SIGTERM grace period acts as
+	// the effective timeout - no additional timeout needed to avoid interfering with
+	// normal Kubernetes client timeout/retry strategy.
+	patchCtx := context.WithoutCancel(ctx)
+
 	err := r.kc.Patch(
-		ctx,
+		patchCtx,
 		res,
 		client.MergeFrom(base),
 	)
@@ -593,8 +600,14 @@ func (r *adoptionReconciler) patchStatus(
 	res *ackv1alpha1.AdoptedResource,
 	base *ackv1alpha1.AdoptedResource,
 ) error {
+	// Always use WithoutCancel to prevent patch operations from being
+	// cancelled while preserving context values. The 30s SIGTERM grace period acts as
+	// the effective timeout - no additional timeout needed to avoid interfering with
+	// normal Kubernetes client timeout/retry strategy.
+	patchCtx := context.WithoutCancel(ctx)
+
 	return r.kc.Status().Patch(
-		ctx,
+		patchCtx,
 		res,
 		client.MergeFrom(base),
 	)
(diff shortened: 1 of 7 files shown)


pkg/runtime/reconciler.go
```diff
@@ -57,6 +57,9 @@ const (
 	// resource if the CARM cache is not synced yet, or if the roleARN is not
 	// available.
 	roleARNNotAvailableRequeueDelay = 15 * time.Second
+	// gracefulShutdownTimeout is the default duration to allow reconciliation
+	// to complete during controller shutdown
+	gracefulShutdownTimeout = 10 * time.Second
```
knottnt, line 62: Q: The PR description says this should be a 30 second window, but it looks like we're using 10 seconds here. What was the reason for shortening the window?

knottnt (commented):
Can we add a test for this behavior?

pkg/runtime/reconciler.go
rushmash91: added 10s timeout

pkg/runtime/reconciler.go
```diff
@@ -859,6 +863,21 @@ func (r *resourceReconciler) patchResourceMetadataAndSpec(
...
+	patchCtx, cancel := context.WithTimeout(context.Background(), patchContextTimeout)
+	defer cancel()
+
+	// Transfer important values from the original context to the patch context
+	patchCtx = context.WithValue(patchCtx, ackrtlog.ContextKey, rlog)
+	if ns := ctx.Value("resourceNamespace"); ns != nil {
+		patchCtx = context.WithValue(patchCtx, "resourceNamespace", ns)
+	}
+
+	rlog.Info("created patch context with timeout", "timeout_seconds", patchContextTimeout.Seconds())
```
a-hilaly, line 879: tiny performance comment, we create this context right before the patch call. in case 890 or 984 fail, we don't leave any unused variables in memory (yes, they will be gc-ed later on, but they need to exist at all)

pkg/runtime/reconciler.go
```diff
@@ -859,6 +863,21 @@ func (r *resourceReconciler) patchResourceMetadataAndSpec(
 ) (acktypes.AWSResource, error) {
 	var err error
 	rlog := ackrtlog.FromContext(ctx)
+
```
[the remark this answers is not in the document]
knottnt: Q: Is the reasoning behind using this that the KILL signal sent after the initial Pod termination grace period acts as an effective timeout in the event that the patch operation(s) take too long?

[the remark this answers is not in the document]
rushmash91: Changed it. The `SIGKILL 30s` acts as the timeout.

pkg/runtime/reconciler.go
```diff
@@ -57,10 +57,6 @@ const (
 	// resource if the CARM cache is not synced yet, or if the roleARN is not
 	// available.
 	roleARNNotAvailableRequeueDelay = 15 * time.Second
-	// adoptOrCreate is an annotation field that decides whether to create the
-	// resource if it doesn't exist, or adopt the resource if it exists.
-	// value comes from getAdoptionPolicy
-	// adoptOrCreate = "adopt-or-create"
```
michaelhtm, line 63: 👍

pkg/runtime/reconciler.go
```diff
@@ -880,7 +876,14 @@ func (r *resourceReconciler) patchResourceMetadataAndSpec(
...
 	lorig := latestCleaned.DeepCopy()
 	patch := client.MergeFrom(desiredCleaned.RuntimeObject())
-	err = r.kc.Patch(ctx, latestCleaned.RuntimeObject(), patch)
+
+	// NOTE (rushmash91): Always use WithoutCancel to prevent patch operations from being
+	// cancelled while preserving context values. The 30s SIGTERM grace period acts as
+	// the effective timeout - no additional timeout needed to avoid interfering with
+	// normal Kubernetes client timeout/retry strategy.
+	patchCtx := context.WithoutCancel(ctx)
+	err = r.kc.Patch(patchCtx, latestCleaned.RuntimeObject(), patch)
```
michaelhtm, line 885: what happens if the parent context gets canceled?

rushmash91: patch is not cancelled. Pretty cool. Plus the values are passed on.

pkg/runtime/adoption_reconciler.go
```diff
@@ -575,8 +575,15 @@ func (r *adoptionReconciler) patchMetadataAndSpec(
...
+
+	// Always use WithoutCancel to prevent patch operations from being
+	// cancelled while preserving context values. The 30s SIGTERM grace period acts as
+	// the effective timeout - no additional timeout needed to avoid interfering with
+	// normal Kubernetes client timeout/retry strategy.
+	patchCtx := context.WithoutCancel(ctx)
+
 	err := r.kc.Patch(
-		ctx,
+		patchCtx,
```
knottnt, line 586: We're using almost the exact same logic several files now. It might make sense to create a new `patchWithGracePeriod()` function that wraps this logic. The comment could also be moved to the function description instead of being copied in multiple places.

rushmash91: That's a good point.

rushmash91: Done!

pkg/runtime/reconciler_test.go
```diff
@@ -46,6 +48,22 @@ import (
...
+	}
+
+	// Check the type name for WithoutCancel context
+	typeName := fmt.Sprintf("%T", ctxVal)
+	return strings.Contains(typeName, "withoutCancelCtx")
+}
+
+// withoutCancelContextMatcher returns a matcher for WithoutCancel contexts
+var withoutCancelContextMatcher = mock.MatchedBy(isWithoutCancelContext)
+
```
rushmash91, line 66: Here.

a-hilaly (approved):
/lgtm

PR #193, opened 2025-07-17
chore: Use client instead of APIReader + Add observedGeneration to status Conditions

michaelhtm:
Description of changes:
Currently in ACK, at the beginning of the Reconciler for each event, we 
get the resource from the k8s APIServer. This is an unecessary call, as 
the latest object that triggered the reconciler is already stored in the 
informers cache. 

With this change, we will be retrieving the resource from the cache 
instead of the APIServer.


pkg/runtime/adoption_reconciler.go
```diff
@@ -625,7 +626,7 @@ func NewAdoptionReconcilerWithClient(
 			metrics:   metrics,
 			cache:     cache,
 			kc:        kc,
-			apiReader: apiReader,
+			ackResourceCache: ackResourceCache,
```
a-hilaly, line 629: i think this can be gofmt-ed

pkg/runtime/reconciler_test.go
```diff
@@ -505,7 +505,7 @@ func TestReconcilerAdoptOrCreateResource_Adopt(t *testing.T) {
 	latest, latestRTObj, latestMetaObj := resourceMocks()
 	latest.On("Identifiers").Return(ids)
 	latest.On("Conditions").Return([]*ackv1alpha1.Condition{})
-		latest.On(
+	latest.On(
```
a-hilaly, line 508: 👍

a-hilaly (commented):
Dankeshoen!

pkg/runtime/field_export_reconciler.go
```diff
@@ -197,15 +198,15 @@ func (r *fieldExportReconciler) getFieldExport(
...
 	ro := &ackv1alpha1.FieldExport{}
-	// Here we use k8s APIReader to read the k8s object by making the
+	// Here we use k8s ackResourceCache to read the k8s object by making the
 	// direct call to k8s apiserver instead of using k8sClient.
 	// The reason is that k8sClient uses a cache and sometimes k8sClient can
 	// return stale copy of object.
 	// It is okay to make direct call to k8s apiserver because we are only
 	// making single read call for complete reconciler loop.
 	// See following issue for more details:
 	// https://github.com/aws-controllers-k8s/community/issues/894
```
knottnt, line 208: Is the problem described in this issue no longer relevant?

michaelhtm: It's not related to a problem..it's mainly just an optimization.
I'll expand more on it in a community PR

PR #196, opened 2025-07-30
feat: retrieve aws partition from callerIdentity

michaelhtm:
Description of changes:
Retrieve Partition from sts Caller identity. This will help with the generation of 
ACK resource status, adding support for all partitions

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/jaypipes/envutil"
@@ -94,6 +95,7 @@ type Config struct {
 	LeaderElectionNamespace         string
 	EnableDevelopmentLogging        bool
 	AccountID                       string
+	Partition                       string
 	Region                          string
 	IdentityEndpointURL             string
 	EndpointURL                     string
@@ -296,9 +298,10 @@ func (cfg *Config) SetupLogger() {
 	klog.SetLogger(logger)
 }
 
-// SetAWSAccountID uses sts GetCallerIdentity API to find AWS AccountId and set
-// in Config
-func (cfg *Config) SetAWSAccountID(ctx context.Context) error {
+// SetAWSAccountInfo uses sts GetCallerIdentity API to find AWS AccountId and partition
+//
+//	to set in Config
+func (cfg *Config) SetAWSAccountInfo(ctx context.Context) error {
 	awsCfg, err := config.LoadDefaultConfig(
 		ctx,
 		config.WithRegion(cfg.Region),
@@ -316,6 +319,14 @@ func (cfg *Config) SetAWSAccountID(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("unable to get caller identity: %v", err)
 	}
+	if res.Arn == nil {
+		return fmt.Errorf("caller identity response missing ARN")
+	}
+	parsedARN, err := arn.Parse(*res.Arn)
+	if err != nil {
+		return fmt.Errorf("error parsing arn: %v", err)
+	}
+	cfg.Partition = parsedARN.Partition
 	cfg.AccountID = *res.Account
 	return nil
 }
@@ -369,8 +380,8 @@ func (cfg *Config) Validate(ctx context.Context, options ...Option) error {
 		}
 	}
 
-	if err := cfg.SetAWSAccountID(ctx); err != nil {
-		return fmt.Errorf("unable to determine account ID: %v", err)
+	if err := cfg.SetAWSAccountInfo(ctx); err != nil {
+		return fmt.Errorf("unable to determine account info: %v", err)
 	}
 
 	if cfg.EnableWebhookServer && cfg.WebhookServerAddr == "" {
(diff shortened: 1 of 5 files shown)


michaelhtm:
tested in https://github.com/aws-controllers-k8s/s3-controller/pull/173

pkg/config/config.go
```diff
@@ -369,7 +380,7 @@ func (cfg *Config) Validate(ctx context.Context, options ...Option) error {
 		}
 	}
 
-	if err := cfg.SetAWSAccountID(ctx); err != nil {
+	if err := cfg.SetAWSAccountInfo(ctx); err != nil {
 		return fmt.Errorf("unable to determine account ID: %v", err)
```
knottnt: nit: this error message is a bit out of date.
```suggestion
		return fmt.Errorf("unable to determine account info: %v", err)
```

pkg/runtime/config.go
```diff
@@ -109,3 +110,18 @@ func formatUserAgent(name, version string, extra ...string) string {
...
 	return ua
 }
+
+// getPartition gets the partition of the caller identity
+func (c *serviceController) getPartition(ctx context.Context, client *sts.Client) (string, error) {
+	identity, err := client.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+	// what od we do if ARN is nil?
+	if err != nil || identity.Arn == nil {
+		return "", err
+	}
```
knottnt: p: It seems like this would be a fairly unexpected situation so throwing an error here makes sense. That said we should make sure err is actually populated in this case.
```suggestion
	// what od we do if ARN is nil?
	if err != nil {
		return "", err
	}
	if identity == nil || identity.Arn == nil {
		return "", errors.New("GetCallerIdentity provided response with no ARN.")
	}
```

pkg/runtime/reconciler.go
```diff
@@ -1486,6 +1488,20 @@ func (r *resourceReconciler) getRegion(
...
+// getPartition attempts getting the partition from the resource status
+// if it exists
+func (r *resourceReconciler) getPartition(
+	res acktypes.AWSResource,
+) string {
+	// first try to get the region from the status.resourceMetadata
+	metadataRegion := res.Identifiers().Partition()
+	if metadataRegion != nil {
+		return string(*metadataRegion)
+	}
```
knottnt: nit: variable naming seems a bit off.
```suggestion
	metadataPartition := res.Identifiers().Partition()
	if metadataPartition != nil {
		return string(*metadataPartition)
	}
```

pkg/types/aws_resource_manager.go
```diff
@@ -114,6 +114,13 @@ type AWSResourceManagerFactory interface {
 		ackv1alpha1.AWSRegion,
 		ackv1alpha1.AWSResourceName,
 	) (AWSResourceManager, error)
+	// GetCachedManager returns an AWSResourceManager if it has previously been created
+	// and cahced, or returns nil if not
```
knottnt: nit
```suggestion
	// GetCachedManager returns an AWSResourceManager if it has previously been created
	// and cached, or returns nil if not
```

pkg/runtime/reconciler.go
```diff
@@ -333,24 +333,26 @@ func (r *resourceReconciler) Reconcile(ctx context.Context, req ctrlrt.Request)
...
 	}
 
-	// The config pivot to the roleARN will happen if it is not empty.
-	// in the NewResourceManager
-	clientConfig, err := r.sc.NewAWSConfig(ctx, region, &endpointURL, roleARN, gvk, desired.MetaObject().GetLabels())
-	if err != nil {
-		return ctrlrt.Result{}, err
-	}
-
 	rlog.WithValues(
```
knottnt, line 343: Q: Should we be including the partition in the log values?

pkg/runtime/reconciler.go
```diff
@@ -333,24 +333,26 @@ func (r *resourceReconciler) Reconcile(ctx context.Context, req ctrlrt.Request)
...
 		"role", roleARN,
 		"region", region,
 	)
 
-	rm, err := r.rmf.ManagerFor(
-		r.cfg, clientConfig, r.log, r.metrics, r, acctID, region, roleARN,
-	)
-	if err != nil {
-		return ctrlrt.Result{}, err
+	rm := r.rmf.GetCachedManager(acctID, region, roleARN)
```
knottnt: Q: ManagerFor already used a cache before creating a new resource manager. What's the reason for pulling out the cache check here?

pkg/runtime/config.go
```diff
@@ -109,3 +110,18 @@ func formatUserAgent(name, version string, extra ...string) string {
 	}
 	return ua
 }
+
+// getPartition gets the partition of the caller identity
+func (c *serviceController) getPartition(ctx context.Context, client *sts.Client) (string, error) {
```
knottnt: Q: Where is this called?

pkg/runtime/reconciler.go
michaelhtm: mm..not sure if it is necessary.

pkg/runtime/reconciler.go
michaelhtm: We're trying to avoid creating the config unless it's necessary

pkg/runtime/config.go
michaelhtm: it's no longer used..removing

pkg/runtime/reconciler.go
michaelhtm: Offline discussion
Decided to remove cache changes from this PR and address it in a separate PR

knottnt (approved):
/lgtm

knottnt:
e2e tests failing as corresponding code-gen changes are required. 

https://github.com/aws-controllers-k8s/code-generator/pull/615

PR #199, opened 2025-09-13
feat: adds EnableCARM flag to prevent watching namespaces in ns scope

vijaydandu:
fixes https://github.com/aws-controllers-k8s/community/issues/2625

Description of changes:
- adds explicit EnableCARM flag to enable/disable Cross Account Resource Management

### DIFF
--- a/pkg/runtime/service_controller.go
+++ b/pkg/runtime/service_controller.go
@@ -220,13 +220,8 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 		}},
 		cfg.FeatureGates,
 	)
-	// We want to run the caches if the length of the namespaces slice is
-	// either 0 (watching all namespaces) or greater than 1 (watching multiple
-	// namespaces).
-	//
-	// The caches are only used for cross account resource management. If the
-	// controller is not configured to watch multiple namespaces, then we don't
-	// need to run the caches.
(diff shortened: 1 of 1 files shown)


pkg/runtime/service_controller.go
```diff
@@ -227,7 +227,7 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 	// The caches are only used for cross account resource management. If the
 	// controller is not configured to watch multiple namespaces, then we don't
 	// need to run the caches.
-	if len(namespaces) == 0 || len(namespaces) >= 2 {
+	if cfg.EnableCARM {
```
a-hilaly, line 230: i'm not opposed to this, but if you want to allow users to enable CARM even they're on single namespace mode, should we update the comment above?

a-hilaly (commented):
Neat, thank you @vijaydandu ! i have one comment below

a-hilaly (approved):
Thank you @vijaydandu !

/cc @rushmash91 @michaelhtm @knottnt

tomreece:
Would it be worthwhile to name the new flag without an acronym to avoid confusion? Or does everyone know what CARM is?

`enable-cross-account-resource-management`

There are already some other long flag names, like `reconcile-resource-max-concurrent-syncs` so even though my suggested name is long, at least it's explicit.

What do you think @vijaydandu ?

PR #204, opened 2025-09-24
feat: add `IAMRoleSelector` CRD/feature

a-hilaly:
Implements https://github.com/aws-controllers-k8s/community/pull/2628 (mostly)

Introduces a new IAMRoleSelector CRD that enables dynamic IAM role
assignment based on namespace and resource type selectors. This feature
provides an alternative to CARM for role selection and cannot be used
simultaneously with CARM (enforced by validation).

Key components:
- New IAMRoleSelector CRD with namespace and resource type selectors
- Selector matching logic with AND between selector types, OR within arrays
- Dynamic informer-based cache for IAMRoleSelector resources
- Integration into the reconciler to override CARM role selection
- Alpha feature gate (IAMRoleSelector) defaulting to disabled

Note: ResourceTypeSelector uses schema.GroupVersionKind in the API,
which differs from the separate fields approach in the original types.
This may need adjustment based on CRD generation requirements.


pkg/runtime/iamroleselector/matcher.go
```diff
@@ -0,0 +1,174 @@
...
+		if seen[name] {
+			return fmt.Errorf("duplicate namespace name: %s", name)
+		}
+		seen[name] = true
+	}
+
+	// Validate label selector
+	if len(nsSelector.LabelSelector.MatchLabels) > 0 {
+		for key := range nsSelector.LabelSelector.MatchLabels {
+			if key == "" {
```
michaelhtm: nit: should we move all validations to a new file? called `validator`...they don't seem to be related to `matcher`

PR #205, opened 2025-09-24
feat: log delta differences for read-only resources when spec changes

a-hilaly:
Add informational logging when a read-only resource's desired state 
differs from its current state. This helps operators understand what 
changes would have been applied if the resource wasn't in read-only mode.


knottnt (approved):
👍

PR #207, opened 2025-09-26
Add retry mechanism for patch operations to handle resource version conflicts

rushmash91:
fixes: https://github.com/aws-controllers-k8s/community/issues/2267

ACK controllers were experiencing race conditions when multiple controllers attempted to patch the same Kubernetes resource simultaneously. This occurred because Kubernetes uses optimistic concurrency control through resource versions - each resource has a version that must match when performing updates. When multiple controllers read a resource and then attempt to patch it, only the first patch succeeds while subsequent patches fail with HTTP 409 conflict errors due to stale resource versions.

Description of changes:

A retry mechanism using Kubernetes client-go's standard `retry.RetryOnConflict` function with `retry.DefaultBackoff` configuration. When a patch operation encounters a resource version conflict, the mechanism automatically refreshes the resource version by fetching the latest state from the API server, then retries the patch operation with the updated version. This process repeats up to 5 times with exponential backoff (starting at 100ms, doubling each time, capped at 1 second), ensuring that temporary conflicts are resolved automatically while avoiding excessive API server load.

Ref:
- https://github.com/kubernetes-sigs/kro/blob/c1bc05c5384245d3ef2a5104198459732552b148/pkg/controller/resourcegraphdefinition/controller_status.go#L34-L75
- https://github.com/prometheus-operator/prometheus-operator/blob/3ff38ebe6216c28da344abd6e4f698831309b959/pkg/k8sutil/k8sutil.go#L239-L376
- https://github.com/prometheus-operator/prometheus-operator/blob/3ff38ebe6216c28da344abd6e4f698831309b959/pkg/k8sutil/k8sutil.go#L513-L522

### DIFF
--- a/pkg/runtime/adoption_reconciler.go
+++ b/pkg/runtime/adoption_reconciler.go
@@ -576,8 +576,7 @@ func (r *adoptionReconciler) patchResourceMetadataAndSpec(
 	// Keep a copy of status field to reset the status of 'res' after patch call
 	resStatusCopy := res.DeepCopy().Status
 
-	rlog := ackrtlog.FromContext(ctx)
-	err := patchMetadataAndSpec(ctx, r.kc, r.apiReader, res, client.MergeFrom(base), rlog)
+	err := patchMetadataAndSpec(ctx, r.kc, res, client.MergeFrom(base))
 	res.Status = resStatusCopy
 	return err
 }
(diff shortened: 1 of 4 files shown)


a-hilaly (approved):
👍

rushmash91:
@michaelhtm removed the retry for spec patches, I found a few examples for it but I agree it not required here and may create issue for conflicts for manual changes.

PR #228, opened 2026-02-27
fix: use controller's base region for CARM STS AssumeRole calls

michaelhtm:
Description of changes:
When a resource targets a different region than the controller's home
region, the STS AssumeRole call was incorrectly using the target
region's STS endpoint. This fails when the target is an opt-in region
whose STS endpoint cannot process AssumeRole for the base region's
credentials. With this change, AssumeRole calls are made using the
base region's STS endpoint, and the returned config is then set to the
desired target region for resource management


pkg/runtime/config.go
```diff
@@ -43,6 +43,7 @@ func (c *clientWithUserAgent) Do(r *http.Request) (*http.Response, error) {
 
 func (c *serviceController) NewAWSConfig(
 	ctx context.Context,
+	baseRegion ackv1alpha1.AWSRegion,
 	region ackv1alpha1.AWSRegion,
```
knottnt, line 47: nit: Without additional knowledge it isn't totally clear what the `baseRegion` and `region` parameters actually mean. Could you add a description for the function that defines these parameters? Also renaming the parameters to something like `defaultControllerRegion` and `targetRegion/resourceRegion` could help.

knottnt: An explanation of why both values are needed would also be helpful.

knottnt (commented):
@michaelhtm Just a minor nit on the actual code change, but I think we'll also need to add a test validate this fix.

PR #229, opened 2026-03-03
Verify resource deletion before removing finalizer

a-hilaly:
Some AWS APIs return success from Delete immediately while the resource
transitions through a deleting state. This change ensures the reconciler
verifies the resource is fully deleted (ReadOne returns NotFound) before
removing the finalizer and allowing the Kubernetes CR to be deleted.

When a resource still exists after Delete, the reconciler sets
ACK.ResourceSynced to False with reason "Deleting" and requeues after
15 seconds. This prevents orphaned AWS resources when the CR is deleted
before AWS finishes cleaning up.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -1129,13 +1129,36 @@ func (r *resourceReconciler) deleteResource(
 		latest, _ = r.patchResourceMetadataAndSpec(ctx, rm, current, latest)
 	}
 	if err != nil {
-		// NOTE: Delete() implementations that have asynchronously-completing
-		// deletions should return a RequeueNeededAfter.
 		return latest, err
 	}
 
-	// Now that external AWS service resources have been appropriately cleaned
-	// up, we remove the finalizer representing the CR is managed by ACK,
+	// Verify the AWS resource is actually deleted before removing the
+	// finalizer. Some AWS APIs return success immediately while the resource
+	// transitions through a "deleting" state. We need to ensure the resource
+	// is fully gone (ReadOne returns NotFound) before allowing the CR to be
+	// deleted from Kubernetes.
+	rlog.Enter("rm.ReadOne (verify deletion)")
+	_, verifyErr := rm.ReadOne(ctx, current)
+	rlog.Exit("rm.ReadOne (verify deletion)", verifyErr)
+	if verifyErr == nil {
+		// Resource still exists, requeue to wait for deletion to complete
+		resForCondition := latest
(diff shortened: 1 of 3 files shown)


pkg/runtime/reconciler.go
```diff
@@ -1129,13 +1129,36 @@ func (r *resourceReconciler) deleteResource(
...
 		return latest, err
 	}
 
-	// Now that external AWS service resources have been appropriately cleaned
-	// up, we remove the finalizer representing the CR is managed by ACK,
+	// Verify the AWS resource is actually deleted before removing the
+	// finalizer. Some AWS APIs return success immediately while the resource
+	// transitions through a "deleting" state. We need to ensure the resource
+	// is fully gone (ReadOne returns NotFound) before allowing the CR to be
+	// deleted from Kubernetes.
```
knottnt, line 1139: This should be good enough for most resources, but there could be cases where the AWS resource is still returned by ReadOne, but with a state indicating that deletion was successful. One example that comes to mind is JobRuns from emr-containers. We likely need to added a check for a "deleted" state similar to our `sycned:when` check.

knottnt: We'll also need to consider how this is rolled out since we'll need to have those checks in place before applying this to avoid these kinds of resources becoming stuck during deletion.

pkg/runtime/reconciler.go
```diff
@@ -1129,13 +1129,36 @@ func (r *resourceReconciler) deleteResource(
...
+	rlog.Enter("rm.ReadOne (verify deletion)")
+	_, verifyErr := rm.ReadOne(ctx, current)
+	rlog.Exit("rm.ReadOne (verify deletion)", verifyErr)
+	if verifyErr == nil {
+		// Resource still exists, requeue to wait for deletion to complete
```
knottnt, line 1144: On the second reconcile we may need to protect against sdkDelete returning an error if the AWS API returns a validation error when called on a deleting resource.

pkg/runtime/reconciler.go
a-hilaly: I'm aware of those edge cases and willing to make a code-gen patch to support that. I just wanted to get a feel on this change, and alignment that this is something we want to support in ACK.

pkg/runtime/reconciler.go
a-hilaly: Yeah we'll probably need a both a `IsDeleted` and `IsDeleting`

PR #234, opened 2026-04-08
feat(runtime): Add pre-delete sync step to deleteResource

sapphirew:
Add `AWSResourceDescriptorWithPreDeleteDelta` optional interface and preDeleteSync helper method to reconcile spec differences (including `compare.is_ignored` fields) before calling rm.Delete. This ensures fields like `DeletionProtectionEnabled` are synced to AWS before deletion.

The pre-delete sync:
- Uses `DeltaForPreDelete` if the descriptor implements the optional interface, otherwise falls back to standard Delta
- Calls `rm.Update` when spec differences are detected
- Proceeds with deletion even if the update fails
- Is a no-op when no spec differences exist

Issue #, if available:

Description of changes:


pkg/runtime/reconciler.go
```diff
@@ -1089,6 +1089,52 @@ func (r *resourceReconciler) patchResourceStatus(
...
+	// DeltaForPreDelete interface. This interface produces a delta that
+	// includes fields marked with compare.is_ignored (e.g.
+	// DeletionProtectionEnabled). Without it, the standard Delta is
+	// already handled by the normal Sync flow, so there is nothing
+	// extra to reconcile before deletion.
+	pdrd, ok := r.rd.(acktypes.AWSResourceDescriptorWithPreDeleteDelta)
+	if !ok {
+		return observed
+	}
+	delta := pdrd.DeltaForPreDelete(desired, observed)
```
sapphirew: proposal: make a deep copy of observed, apply delta, return as a new var like newDesired to pass through
```
rm.Update(ctx, newDesired, observed, delta)
```
otherwise the Update call would use desired including fields we don't want to update for pre-delete sync

pkg/runtime/reconciler.go
```diff
@@ -1089,6 +1089,52 @@ func (r *resourceReconciler) patchResourceStatus(
 //
 // Returns a copy of the resource with the latest state either right before
 // deletion OR after a failed attempted deletion.
```
knottnt, line 1091: Should this be above the `deleteResource` function?

pkg/types/aws_resource_descriptor_pre_delete.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// AWSResourceDescriptorWithPreDeleteDelta is an optional extension of
+// AWSResourceDescriptor that provides a delta function including fields
+// normally excluded by compare.is_ignored. This is used during pre-delete
+// sync to ensure fields like DeletionProtectionEnabled are compared before
+// deletion.
+type AWSResourceDescriptorWithPreDeleteDelta interface {
+	// DeltaForPreDelete returns an `ackcompare.Delta` that includes fields
+	// normally excluded by compare.is_ignored configuration. Used during
```
knottnt: Q: Should this be limited to only check the fields needed to unblock a delete operation?

pkg/runtime/reconciler.go
knottnt: This seems like a reasonable strategy for keeping the update minimal.

pkg/types/aws_resource_descriptor_pre_delete.go
sapphirew: yes, had a discussion with @michaelhtm. Going to limit down only to fields needed to unblock a delete operation

pkg/runtime/reconciler.go
sapphirew: yes, good catch

knottnt (approved):

pkg/runtime/reconciler.go
```diff
@@ -1084,6 +1084,57 @@ func (r *resourceReconciler) patchResourceStatus(
...
+	// in the delta overwritten from desired. This ensures rm.Update only
+	// changes the fields that DeltaForPreDelete detected as different
+	// while all other fields retain their current AWS values.
+	updated, err := rm.Update(ctx, merged, observed, delta)
+	if err != nil {
+		rlog.Info(
+			"pre-delete sync: update failed, proceeding with delete",
+			"error", err,
+		)
+		return observed
```
michaelhtm: why are we not returning the error?

sapphirew: It's intentional since user already initiated a deletion. It should not be stuck in a deletion loop if this update keeps failing

sapphirew: discussed offline: only surface/wrap the error from rm.Update if rm.Delete() also errored out. So even if update failed, it would still proceed to deleting. Users would be able to see both errors in the condition if both operations failed

pkg/runtime/reconciler.go
```diff
@@ -1114,6 +1165,11 @@ func (r *resourceReconciler) deleteResource(
 		}
 		return current, err
 	}
+	// Pre-delete sync: reconcile spec differences before deletion.
+	// This ensures fields like DeletionProtectionEnabled are synced
+	// to AWS before the Delete call.
+	observed, preDeleteErr := r.preDeleteSync(ctx, rm, current, observed)
```
a-hilaly: What if the preDeleteSync is an async operation that will take a long time. shouldn't we requeue in here?

pkg/runtime/reconciler.go
```diff
@@ -1084,6 +1084,57 @@ func (r *resourceReconciler) patchResourceStatus(
 	return err
 }
 
+// preDeleteSync reconciles spec differences between the desired CR state
+// and the observed AWS resource state before deletion. This ensures fields
+// like DeletionProtectionEnabled are updated on the AWS resource before the
+// Delete API call. If the update fails, the original observed resource is
+// returned so that deletion can still proceed.
+func (r *resourceReconciler) preDeleteSync(
```
a-hilaly, line 1092: why don't we just pass `delta` object to `skdDelete` and let the controller decide what to do with it?

pkg/runtime/reconciler.go
sapphirew: It should be handled by sdkUpdate automatically if the update is an async operation, based on the implementation of individual controllers

pkg/runtime/reconciler.go
sapphirew: How is that different from the current implementation? What can we gain by moving the same business logic around?

michaelhtm (approved):

pkg/runtime/reconciler.go
sapphirew: offline discussion - will keep the logic for preDeleteSync in runtime since there would be multiple controller using this feature, like rds, sns and dsql etc

pkg/runtime/reconciler.go
sapphirew: offline discussion - will fail fast if preDelete sync fails
- preDeleteSync error → return error, no delete attempted, reconcile loop requeues
- preDeleteSync success → proceed to `rm.Delete`
- Requeue errors from sdkUpdate (like "resource is in UPDATING state") are naturally honored — they bubble up through `HandleReconcileError` which respects `RequeueNeededAfter` and requeues with the appropriate delay.

knottnt (approved):
/lgtm

PR #236, opened 2026-05-05
fix: report adoption errors as Terminal conditions

michaelhtm:
Description of changes:
Currently, if the user makes a typo when defining the ACK resource
adoption policy or adoption fields, they needed to rely on the logs to
figure out the issue. With these changes, we ensure these errors are
exposed via the resource status with a clear error message

### DIFF
--- a/pkg/condition/condition.go
+++ b/pkg/condition/condition.go
@@ -324,6 +324,21 @@ func WithReferencesResolvedCondition(
 	return ko
 }
 
+// WithTerminalCondition returns a new AWSResource with the
+// ConditionTypeTerminal set based on the err parameter
+func WithTerminalCondition(
+	resource acktypes.AWSResource,
+	err error,
+) acktypes.AWSResource {
+	ko := resource.DeepCopy()
+
+	if err != nil {
+		errString := err.Error()
+		SetTerminal(ko, corev1.ConditionTrue, &errString, nil)
+	}
+	return ko
+}
+
 // LateInitializationInProgress return true if ConditionTypeLateInitialized has "False" status
 // False status means that resource has LateInitializationConfig but has not been completely
 // late initialized yet.
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -451,7 +451,7 @@ func (r *resourceReconciler) handlePopulation(
 	// TODO (michaelhtm) change PopulateResourceFromAnnotation to understand
 	// adopt-or-create, and validate Spec fields are not nil...
 	if err != nil && adoptionPolicy != AdoptionPolicy_AdoptOrCreate {
-		return nil, err
+		return nil, fmt.Errorf("Error populating adoption fields: %v", err)
 	}
 
 	return populated, nil
@@ -528,7 +528,7 @@ func (r *resourceReconciler) Sync(
 	needAdoption := NeedAdoption(desired) && !r.rd.IsManaged(desired) && r.cfg.FeatureGates.IsEnabled(featuregate.ResourceAdoption)
 	adoptionPolicy, err := GetAdoptionPolicy(desired)
 	if err != nil {
-		return nil, err
+		return ackcondition.WithTerminalCondition(desired, err), ackerr.Terminal
 	}
 	if !needAdoption {
 		adoptionPolicy = ""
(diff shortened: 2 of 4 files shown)


pkg/condition/condition.go
```diff
@@ -324,6 +324,21 @@ func WithReferencesResolvedCondition(
...
+// ConditionTypeTerminal set based on the err parameter
+func WithTerminalCondition(
+	resource acktypes.AWSResource,
+	err error,
+) acktypes.AWSResource {
+	ko := resource.DeepCopy()
+
+	if err != nil {
+		errString := err.Error()
+		SetReferencesResolved(ko, corev1.ConditionTrue, nil, &errString)
```
knottnt: Why are we using SetReferencesResolved here? Should this not be setting a terminal condition instead?

pkg/runtime/reconciler.go
```diff
@@ -451,7 +451,7 @@ func (r *resourceReconciler) handlePopulation(
 	// TODO (michaelhtm) change PopulateResourceFromAnnotation to understand
 	// adopt-or-create, and validate Spec fields are not nil...
 	if err != nil && adoptionPolicy != AdoptionPolicy_AdoptOrCreate {
-		return nil, err
+		return nil, fmt.Errorf("populating adoption fields: %v", err)
```
knottnt: nit: Wording should highlight the fact that there was an error.
```suggestion
		return nil, fmt.Errorf("Error populating adoption fields: %v", err)
```

pkg/runtime/util.go
```diff
@@ -194,7 +194,7 @@ func GetAdoptionPolicy(res acktypes.AWSResource) (AdoptionPolicy, error) {
 	}
 
 	if policy != string(AdoptionPolicy_Adopt) && policy != string(AdoptionPolicy_AdoptOrCreate) {
-		return "", fmt.Errorf("unrecognized adoption policy")
+		return "", fmt.Errorf("unrecognized adoption policy %s. Supported policies are \"%s\" and \"%s\" are supported", policy, AdoptionPolicy_Adopt, AdoptionPolicy_AdoptOrCreate)
```
knottnt: nit: error message has some redundant wording.
```suggestion
		return "", fmt.Errorf("unrecognized adoption policy %s. Supported policies are \"%s\" and \"%s\"", policy, AdoptionPolicy_Adopt, AdoptionPolicy_AdoptOrCreate)
```

pkg/runtime/reconciler.go
```diff
@@ -582,10 +582,10 @@ func (r *resourceReconciler) Sync(
 			return latest, err
 		}
 		if adoptionPolicy == AdoptionPolicy_Adopt || isAdopted {
-			return nil, ackerr.AdoptedResourceNotFound
+			return latest, ackerr.AdoptedResourceNotFound
 		}
 		if isReadOnly {
-			return nil, ackerr.ReadOnlyResourceNotFound
+			return latest, ackerr.ReadOnlyResourceNotFound
```
knottnt: Q: What's the reason for changing these cases to return `latest`?

pkg/runtime/reconciler.go
```diff
@@ -554,7 +554,7 @@ func (r *resourceReconciler) Sync(
 	if needAdoption {
 		populated, err := r.handlePopulation(ctx, desired)
 		if err != nil {
-			return nil, err
+			return ackcondition.WithTerminalCondition(desired, err), ackerr.Terminal
```
knottnt, line 557: Q: Why do need to both change the return to the contain  a resource modified to contain Terminal condition and a generic `ackerr.Terminal`? Is there a reason simply wrapping the error in a `TerminalError` with `NewTerminalError()` wouldn't work?

pkg/condition/condition.go
michaelhtm: you're right. Should be terminal

pkg/runtime/reconciler.go
michaelhtm: it's not necessary..removing

pkg/runtime/reconciler.go
michaelhtm: We stop requeuing only on ackerr.Terminal
https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/reconciler.go#L1338-L1339

This way, we still ensure the actual error is exposed via the status condition

knottnt: I see this providing similar behavior to `rm.OnError()`. Makes sense.

knottnt (approved):
/lgtm

knottnt:
Related to [2682](https://github.com/aws-controllers-k8s/community/issues/2682)

PR #239, opened 2026-05-07
feat(runtime): Add --enable-cross-namespace flag

sapphirew:
Issue #, if available:

Description of changes:

Introduce a single CLI flag that controls all three categories of cross-namespace behavior: resource references, secret references, and field exports. It defaults to true with deprecation warnings to start with. It will default to false later in order to follow best security practice.

Changes:
- Add `--enable-cross-namespace` CLI flag with default value `true`
- Update ValidateCrossNamespaceReference to return (string, bool, error)
- Add ValidateCrossNamespaceReferenceString convenience wrapper
- Add ACK.CrossNamespaceDeprecation condition type
- Integrate cross-namespace check into SecretValueFromReference
- Integrate cross-namespace check into field export reconciler
- Update error messages to reference new flag name

### DIFF
--- a/apis/core/v1alpha1/conditions.go
+++ b/apis/core/v1alpha1/conditions.go
@@ -70,6 +70,11 @@ const (
 	// to manage the AWSResource. If none are selected, this condition will be removed
 	// and we'll use the custom role to manage the AWSResource
 	ConditionTypeIAMRoleSelected ConditionType = "ACK.IAMRoleSelected"
+	// ConditionTypeCrossNamespaceOptInRequired indicates that the resource uses
+	// cross-namespace behavior (resource references, secret references, or
+	// field exports) that will require explicit opt-in via
+	// --enable-cross-namespace=true in a future release.
+	ConditionTypeCrossNamespaceOptInRequired ConditionType = "ACK.CrossNamespaceOptInRequired"
 )
 
 // Condition is the common struct used by all CRDs managed by ACK service
--- a/pkg/errors/resource_reference.go
+++ b/pkg/errors/resource_reference.go
@@ -47,6 +47,14 @@ var (
 	ResourceReferenceMissingTargetField = fmt.Errorf(
 		"the referenced resource is missing the target field",
 	)
+	// ResourceReferenceCrossNamespaceNotAllowed indicates that a resource
+	// reference specifies a Namespace that differs from the namespace of the
+	// resource containing the reference, and the
+	// --enable-cross-namespace flag is not enabled.
+	ResourceReferenceCrossNamespaceNotAllowed = fmt.Errorf(
+		"cross-namespace resource reference is not allowed. " +
+			"Set '--enable-cross-namespace' flag to allow it",
+	)
 )
 
 // ResourceReferenceOrIDRequiredFor returns a ResourceReferenceOrIDRequired error
@@ -89,4 +97,19 @@ func ResourceReferenceMissingTargetFieldFor(resource string, namespace string,
 	return fmt.Errorf("%w. resource:%s, namespace:%s, name:%s"+
 		", targetField:%s", ResourceReferenceMissingTargetField,
 		resource, namespace, name, targetField)
+}
+
+// ResourceReferenceCrossNamespaceNotAllowedFor returns a
+// ResourceReferenceCrossNamespaceNotAllowed error annotated with the
+// source namespace, target namespace, and referenced resource name.
+func ResourceReferenceCrossNamespaceNotAllowedFor(
+	sourceNamespace string,
+	targetNamespace string,
+	name string,
+) error {
+	return fmt.Errorf(
+		"%w. sourceNamespace:%s, targetNamespace:%s, name:%s",
+		ResourceReferenceCrossNamespaceNotAllowed,
+		sourceNamespace, targetNamespace, name,
+	)
 }
--- a/pkg/runtime/field_export_reconciler.go
+++ b/pkg/runtime/field_export_reconciler.go
@@ -153,6 +153,23 @@ func (r *fieldExportReconciler) Sync(
 		r.patchResourceStatus(ctx, &desired, latest)
 	}()
 
+	// Validate cross-namespace access for the field export target
+	resolvedNamespace, isCrossNamespace, nsErr := r.validateFieldExportNamespace(&desired)
+	if nsErr != nil {
+		return desired, r.onError(ctx, &desired, ackerr.NewTerminalError(nsErr))
+	}
+	if isCrossNamespace {
+		r.log.V(0).Info(
+			"cross-namespace field export detected; this behavior will be "+
+				"disabled by default in a future release. Set --enable-cross-namespace "+
+				"to preserve this behavior.",
+			"fieldExportNamespace", desired.Namespace,
+			"targetNamespace", r.getTargetNamespace(&desired),
+			"targetName", *desired.Spec.To.Name,
+		)
+		r.setCrossNsOptInRequiredCondition(&desired)
+	}
+
 	// Get the field from the resource
 	value, err := r.getSourcePathFromResource(from, *desired.Spec.From.Path)
 	if err != nil {
@@ -163,11 +180,11 @@ func (r *fieldExportReconciler) Sync(
 
 	switch desired.Spec.To.Kind {
 	case ackv1alpha1.FieldExportOutputTypeConfigMap:
-		if err = r.writeToConfigMap(ctx, *value, &desired); err != nil {
+		if err = r.writeToConfigMap(ctx, *value, &desired, resolvedNamespace); err != nil {
 			return desired, r.onError(ctx, &desired, err)
 		}
 	case ackv1alpha1.FieldExportOutputTypeSecret:
-		if err = r.writeToSecret(ctx, *value, &desired); err != nil {
+		if err = r.writeToSecret(ctx, *value, &desired, resolvedNamespace); err != nil {
 			return desired, r.onError(ctx, &desired, err)
 		}
 	}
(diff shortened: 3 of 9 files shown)


cheeseandcereal (commented):
First-pass review against the [ACK rubric](https://github.com/aws-controllers-k8s/ack-dev-skills/blob/main/skills/ack-review/references/pr-review-rubric.md). Single squashed commit (rule I passes); no `generator.yaml`/helm/E2E surface in this PR (rules A–H, J don't apply — this is runtime). Most concerns are around scope alignment between the PR description and what actually ships on this branch.

pkg/runtime/reconciler.go
```diff
@@ -131,21 +131,28 @@ func (r *reconciler) SecretValueFromReference(
...
-			namespace = ctxNamespace
+	if ctxResourceNamespace != nil {
+		if ctxNamespace, ok := ctxResourceNamespace.(string); ok {
+			ownerNamespace = ctxNamespace
 		}
 	}
 
+	// Determine the namespace to use for fetching the secret.
+	// Cross-namespace validation, logging, and condition-setting are handled
+	// by the generated code before calling this method.
```
cheeseandcereal, line 150: The PR description says this PR "Integrate[s] cross-namespace check into `SecretValueFromReference`", but this hunk does the opposite — it removes namespace logic and the comment defers validation to "the generated code before calling this method." That generated-side change isn't in this PR (and would live in code-generator, not runtime).

Net effect on this branch: `--enable-cross-namespace=false` does **not** block cross-namespace secret references — only field exports get gated, and resource references are gated indirectly by `WithReferencesResolvedCondition`. Two ways to close this:

1. Wire `ValidateCrossNamespaceReferenceString(r.cfg.EnableCrossNamespace, ownerNamespace, ref.Namespace, ref.Name)` into `SecretValueFromReference` here, parallel to `field_export_reconciler.go`. Keeps the runtime self-contained for tagged releases.
2. If you'd rather centralize in codegen, please trim the PR description to match scope and link the follow-up code-generator PR.

pkg/runtime/field_export_reconciler.go
```diff
@@ -153,6 +153,23 @@ func (r *fieldExportReconciler) Sync(
...
+	if isCrossNamespace {
+		r.log.V(0).Info(
+			"cross-namespace field export detected; this behavior will be "+
+				"disabled by default in a future release. Set --enable-cross-namespace "+
+				"to preserve this behavior.",
+			"fieldExportNamespace", desired.Namespace,
+			"targetNamespace", r.getTargetNamespace(&desired),
+			"targetName", *desired.Spec.To.Name,
+		)
+		r.setCrossNsOptInRequiredCondition(&desired)
```
cheeseandcereal, line 170: `*desired.Spec.To.Name` is dereferenced unconditionally on the log line above — but `validateFieldExportNamespace` (line 204) explicitly guards `desired.Spec.To.Name != nil` before reading the same field. Pre-existing `writeToConfigMap`/`writeToSecret` paths already deref unguarded too, so this PR isn't *introducing* the panic, but the inconsistency within new code is worth resolving: either drop the nil-guard in `validateFieldExportNamespace` (callers downstream already assume non-nil), or thread the safe `targetName` through and use it here. Pick one and be consistent.

pkg/runtime/field_export_reconciler.go
```diff
@@ -177,6 +194,55 @@ func (r *fieldExportReconciler) Sync(
...
+	message := "Cross-namespace field export detected: FieldExport in namespace \"" +
+		desired.Namespace + "\" targets namespace \"" + r.getTargetNamespace(desired) +
+		"\". Cross-namespace behavior will require explicit opt-in in a future release. " +
+		"Set --enable-cross-namespace=true to preserve this behavior."
+	condition := &ackv1alpha1.Condition{
+		Type:    ackv1alpha1.ConditionTypeCrossNamespaceOptInRequired,
+		Status:  corev1.ConditionTrue,
+		Message: &message,
+	}
+	desired.Status.Conditions = append(desired.Status.Conditions, condition)
```
cheeseandcereal, line 243: `setCrossNsOptInRequiredCondition` appends unconditionally. Safe today because `resetConditions` runs at the top of `Sync` (line 151), but it diverges from the lookup-or-create pattern used by `patchRecoverableCondition` (line 517) and `patchTerminalCondition` (line 551). If `resetConditions` is ever moved or skipped, this will accumulate duplicates silently. Prefer the existing pattern (search by type, update if present, append if not) so this code reads the same as its neighbors.

pkg/errors/resource_reference.go
```diff
@@ -47,6 +47,14 @@ var (
 	ResourceReferenceMissingTargetField = fmt.Errorf(
 		"the referenced resource is missing the target field",
 	)
+	// ResourceReferenceCrossNamespaceNotAllowed indicates that a resource
+	// reference specifies a Namespace that differs from the namespace of the
+	// resource containing the reference, and the
+	// --enable-cross-namespace flag is not enabled.
+	ResourceReferenceCrossNamespaceNotAllowed = fmt.Errorf(
```
cheeseandcereal, line 54: `--enable-cross-namespace` is a bool flag; `Set --enable-cross-namespace=true to allow it.` is what operators actually need to type. Worth flipping the message text. Also: while the flag still defaults to `true`, this terminal error will only ever fire when an operator explicitly opted out — so the wording is fine. When the default flips, consider also referencing the `ACK.CrossNamespaceOptInRequired` condition so users can audit their fleet before the flip lands.

pkg/runtime/resource_reference.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	// Same namespace: always allowed, not cross-namespace.
+	if *refNamespace == ownerNamespace {
+		return ownerNamespace, false, nil
+	}
+	// Differing namespace: gated by CLI flag.
+	if enableCrossNamespace {
+		return *refNamespace, true, nil
+	}
+	return "", false, ackerr.ResourceReferenceCrossNamespaceNotAllowedFor(
```
cheeseandcereal, line 61: Minor: returning `""` for `resolvedNamespace` alongside the error is technically fine — the contract says callers must check `err` first — but it's a foot-gun if a future caller ignores the error and falls through to an empty-namespace `Get`. Either return `ownerNamespace` defensively (so a misuse fails *closed* on the owner ns rather than cluster-scoped), or return `*refNamespace` so the rejected value is at least diagnostic.

pkg/runtime/reconciler_test.go
```diff
@@ -2405,3 +2405,74 @@ func TestPreDeleteSync_UpdateAndDeleteBothFail(t *testing.T) {
...
+	_, err := r.Sync(ctx, rm, desired)
+	require.NotNil(err)
+	assert.True(errors.Is(err, ackerr.ResourceReferenceCrossNamespaceNotAllowed),
+		"error should wrap ResourceReferenceCrossNamespaceNotAllowed sentinel")
+
+	// ReadOne must NOT be invoked when references fail to resolve
+	rm.AssertNotCalled(t, "ReadOne", ctx, desired)
+	// No downstream operations should occur
+	rm.AssertNotCalled(t, "Create", ctx, desired)
+	rm.AssertNotCalled(t, "Update")
```
cheeseandcereal, line 2475: Good coverage on the reference-resolution rejection path. Three gaps still uncovered for the new code paths most likely to regress later:

1. `field_export_reconciler.Sync` rejection branch (line 158–159) when `EnableCrossNamespace=false` and the FieldExport target is a different namespace.
2. `setCrossNsOptInRequiredCondition` is set on the FieldExport when flag=true and namespaces differ (the deprecation-warning happy path).
3. `WithReferencesResolvedCondition` (`pkg/condition/condition.go:317-318`): a unit test asserting the new `ResourceReferenceCrossNamespaceNotAllowed` branch resolves to `ConditionFalse` (not `ConditionUnknown`).

Non-blocking, but easy adds while the helper signature is fresh.

apis/core/v1alpha1/conditions.go
```diff
@@ -70,6 +70,11 @@ const (
 	// to manage the AWSResource. If none are selected, this condition will be removed
 	// and we'll use the custom role to manage the AWSResource
 	ConditionTypeIAMRoleSelected ConditionType = "ACK.IAMRoleSelected"
+	// ConditionTypeCrossNamespaceOptInRequired indicates that the resource uses
+	// cross-namespace behavior (resource references, secret references, or
```
cheeseandcereal, line 74: Naming inconsistency: PR description calls this `ACK.CrossNamespaceDeprecation`, the type added here is `ACK.CrossNamespaceOptInRequired`. The latter is the better of the two (describes the action operators must take), but please update the PR body so downstream release notes don't spread the wrong name — and be aware that the condition `Type` string is part of the user-visible API.

kprahulraj:
The ec2-controller-test failure is unrelated to this PR. It's a known issue where system tags (services.k8s.aws/*) leak into spec.tags. A possible fix should be https://github.com/aws-controllers-k8s/runtime/pull/246

kprahulraj:
Doc changes is added - https://github.com/aws-controllers-k8s/docs/pull/37

pkg/runtime/reconciler.go
kprahulraj: Addressed in https://github.com/aws-controllers-k8s/runtime/pull/248 - updated PR description to clarify that cross-namespace validation for secrets is handled by generated code in https://github.com/aws-controllers-k8s/code-generator/pull/699.

SecretValueFromReference here only extracts ownerNamespace; validation is deferred to the caller

kprahulraj:
Addressed the comments in https://github.com/aws-controllers-k8s/runtime/pull/248

pkg/runtime/field_export_reconciler.go
kprahulraj: Fixed in https://github.com/aws-controllers-k8s/runtime/pull/248 - using nil-safe `targetName` variable with guard check before deref, consistent with `validateFieldExportNamespace`.

pkg/runtime/field_export_reconciler.go
kprahulraj: Fixed in https://github.com/aws-controllers-k8s/runtime/pull/248 — now uses lookup-or-create pattern (search by type, update if present, append if not) matching `patchRecoverableCondition`/`patchTerminalCondition`.

pkg/runtime/resource_reference.go
kprahulraj: Fixed in https://github.com/aws-controllers-k8s/runtime/pull/248 - returns `ownerNamespace` instead of `""` so a misuse fails closed on the owner namespace.

pkg/runtime/reconciler_test.go
kprahulraj: Will add as a follow up

apis/core/v1alpha1/conditions.go
kprahulraj: Fixed in https://github.com/aws-controllers-k8s/runtime/pull/248 - PR description updated to use `ACK.CrossNamespaceOptInRequired`.

pkg/errors/resource_reference.go
kprahulraj: Fixed in https://github.com/aws-controllers-k8s/runtime/pull/248 — error message now says `Set --enable-cross-namespace=true to allow it`

sapphirew:
addressed in https://github.com/aws-controllers-k8s/runtime/pull/248

/close

PR #240, opened 2026-05-08
feat: add --lazy-bind-reconcilers flag for deferred CRD binding

michaelhtm:
Issue #, if available:

Description of changes:
When a reconciler is bound for a CRD that doesn't exist in the cluster,
controller-runtime's cache sync times out and crash-loops the pod. This
adds an opt-in flag that defers reconciler binding for missing CRDs
until they become available, allowing the controller to start serving
installed CRDs immediately.

When enabled, a background goroutine polls the discovery API every 10s
for each missing CRD, binding the reconciler once detected. The context
is respected for graceful shutdown, and bind failures trigger os.Exit(1)
since they indicate a permanent configuration error.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -68,6 +68,8 @@ const (
 	flagReconcileResourceMaxConcurrency = "reconcile-resource-max-concurrent-syncs"
 	flagFeatureGates                    = "feature-gates"
 	flagReconcileResources              = "reconcile-resources"
+	flagLazyBindReconcilers             = "lazy-bind-reconcilers"
+	flagLazyBindRetryInterval           = "lazy-bind-retry-interval"
 	envVarAWSRegion                     = "AWS_REGION"
 )
 
@@ -113,6 +115,8 @@ type Config struct {
 	ReconcileDefaultMaxConcurrency  int
 	ReconcileResourceMaxConcurrency []string
 	ReconcileResources              string
+	LazyBindReconcilers             bool
+	LazyBindRetryInterval           time.Duration
 	// TODO(a-hilaly): migrate to k8s.io/component-base and implement a proper parser for feature gates.
 	FeatureGates    featuregate.FeatureGates
 	featureGatesRaw string
@@ -279,6 +283,17 @@ func (cfg *Config) BindFlags() {
 		"",
 		"A comma-separated list of resource kinds to reconcile. If unspecified, all resources will be reconciled.",
 	)
+	flag.BoolVar(
+		&cfg.LazyBindReconcilers, flagLazyBindReconcilers,
+		false,
+		"When enabled, reconcilers for CRDs that are not yet installed in the cluster will be started "+
+			"in the background once the CRDs become available, instead of failing at startup.",
+	)
+	flag.DurationVar(
+		&cfg.LazyBindRetryInterval, flagLazyBindRetryInterval,
+		10*time.Second,
+		"The interval (as a duration string, e.g. '10s', '1m') at which the controller checks for CRD availability when lazy-bind-reconcilers is enabled.",
+	)
 }
 
 // SetupLogger initializes the logger used in the service controller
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -279,6 +283,17 @@ func (cfg *Config) BindFlags() {
...
+	flag.BoolVar(
+		&cfg.LazyBindReconcilers, flagLazyBindReconcilers,
+		false,
+		"When enabled, reconcilers for CRDs that are not yet installed in the cluster will be started "+
+			"in the background once the CRDs become available, instead of failing at startup.",
+	)
+	flag.DurationVar(
+		&cfg.LazyBindRetryInterval, flagLazyBindRetryInterval,
+		10*time.Second,
+		"The interval at which the controller checks for CRD availability when lazy-bind-reconcilers is enabled.",
```
knottnt: nit: Mentioning the unit of time in the description and/or the flag itself would be helpful.

pkg/runtime/service_controller.go
```diff
@@ -81,6 +82,14 @@ type serviceController struct {
...
 	// changing the public ServiceController interface.
 	cfg ackcfg.Config
+	// resourceInstalledCheckFunc is an optional override for checking whether a
+	// CRD is installed. When nil, the default getResourceInstalled is used.
+	// This exists to support testing.
+	resourceInstalledCheckFunc func(ctrlrt.Manager, schema.GroupVersionResource) (bool, error)
+	// bindReconcilerFunc is an optional override for binding a reconciler.
+	// When nil, the default bindReconciler is used. This exists to support
+	// testing without hitting controller-runtime's global controller registry.
+	bindReconcilerFunc func(ctrlrt.Manager, acktypes.AWSResourceManagerFactory, ackcfg.Config, ackrtcache.Caches, *iamroleselector.Cache, bool, logr.Logger) error
```
knottnt: Q: Is this being used?

pkg/runtime/service_controller.go
```diff
@@ -298,25 +316,104 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
...
+
+		log.Info("CRD is now installed, binding reconciler")
+
+		c.metaLock.Lock()
+		err := c.bindReconciler(mgr, rmf, cfg, carmCache, irsCache, exporterInstalled, exporterLogger)
+		c.metaLock.Unlock()
+
+		if err != nil {
+			log.Error(err, "failed to bind reconciler after CRD became available, exiting")
+			os.Exit(1)
```
knottnt: It looks like this would immediately exit the process. Would it be possible to initiate a graceful shutdown instead? Terminating immediately could prevent the controller from patching status updates which we've seen result in orphaned AWS resources.

https://github.com/aws-controllers-k8s/community/issues/2497

pkg/runtime/service_controller_test.go
```diff
@@ -146,6 +156,12 @@ func TestServiceController(t *testing.T) {
 			Kind:  "fakeBook",
 		},
 	)
+	rd.On("GroupVersionResource").Return(
+		schema.GroupVersionResource{
```
knottnt, line 185: nit: should this also include a version?

pkg/runtime/service_controller.go
```diff
@@ -100,18 +109,28 @@ func (c *serviceController) GetResourceManagerFactories() map[string]acktypes.AW
 	return c.rmFactories
 }
 
+// checkResourceInstalled returns whether the given GVR is installed in the
+// cluster. It delegates to resourceInstalledCheckFunc if set (for testing),
+// otherwise falls through to getResourceInstalled.
+func (c *serviceController) checkResourceInstalled(mgr ctrlrt.Manager, gvr schema.GroupVersionResource) (bool, error) {
```
knottnt: Instead of creating a function to wrap the test func it might better to update the getResourceInstalled() function to accept what we need to mock as a dependency. Looking at `getResourceInstalled` by itself it isn't clear that it should be wrapped by the testing function and could lead to different code paths using  getResourceInstalled and checkResourceInstalled in the future. 

I do see that ctrlrt.Manager already has a [GetRESTMapper](https://github.com/kubernetes-sigs/controller-runtime/blob/589525a057391c2be40fd0de7c31ad3bac69a5d7/pkg/manager/internal.go#L271) that we may be able to mock.

PR #241, opened 2026-05-08
Create compare zero functions

gustavodiaz7722:
## Add zero-value comparison utilities

Adds `IsZeroValue`, `IsNilEqualsZero`, and `IsZeroEqualsNil` helper functions to the `compare` package. Related issue

https://github.com/aws-controllers-k8s/community/issues/2873

### Problem

AWS APIs often return empty strings for unset optional fields, while the desired state in a CR has `nil` for fields not specified by the user. This causes permanent drift detection on every reconciliation loop (e.g., SNS Topic `DisplayName`).

### Solution

New comparison utilities that determine whether a nil value and a zero value (e.g., `""`, `0`, `false`) should be considered equivalent:

- `IsZeroValue(i)` — returns `true` if the value (or dereferenced pointer) equals the zero value for its type. Returns `false` for `nil` and nil pointers.
- `IsNilEqualsZero(a, b)` — returns `true` if `a` is nil and `b` is a zero value.
- `IsZeroEqualsNil(a, b)` — returns `true` if `a` is nil and `b` is a zero value.

### Testing

Table-driven unit tests covering nil, nil pointers, pointers to zero/non-zero values, and direct value types.

Fixes: https://github.com/aws-controllers-k8s/community/issues/2873Issue #, if available:

Description of changes:

### DIFF
--- a/pkg/compare/zero.go
+++ b/pkg/compare/zero.go
@@ -1,0 +1,29 @@
+package compare
+
+import (
+	"reflect"
+)
+
+func IsZeroValue(i interface{}) bool {
+	if i == nil {
+		return false
+	}
+	v := reflect.ValueOf(i)
+	// Dereference pointer to check if the pointed-to value is zero
+	if v.Kind() == reflect.Ptr {
+		if v.IsNil() {
+			return false
+		}
+		return v.Elem().IsZero()
+	}
+	return v.IsZero()
+}
+
+// IsNilEqualsZero returns true if 'a' satisfies IsNil and 'b' satisfies IsZeroValue.
+func IsNilEqualsZero(a interface{}, b interface{}) bool {
+	return IsNil(a) && IsZeroValue(b)
+}
+
+func IsZeroEqualsNil(a interface{}, b interface{}) bool {
+	return IsNil(a) && IsZeroValue(b)
+}
--- a/pkg/compare/zero.go
+++ b/pkg/compare/zero.go
@@ -25,5 +25,5 @@ func IsNilEqualsZero(a interface{}, b interface{}) bool {
 }
 
 func IsZeroEqualsNil(a interface{}, b interface{}) bool {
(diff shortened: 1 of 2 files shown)


pkg/compare/zero.go
```diff
@@ -0,0 +1,29 @@
+package compare
```
sapphirew, line 1: did you miss to commit the file where these new functions are invoked?

gustavodiaz7722: No, I was planning to reference these functions in a code generator change. They will be used in similar fashion to nil.go is used in code generator.

sapphirew: I see, but how are we going to use these functions? Will `IsZeroEqualsNil` and `IsZeroEqualsNil` be used at the same time? If yes, should we expose one single function that `return (IsNil(a) && IsZeroValue(b)) || (IsZeroValue(a) && IsNil(b))`

gustavodiaz7722: No they will not be used at the same time. For now I plan to complete implementation for the following configuration option in code generator
```
	// NilEqualsZeroValue indicates a nil pointer and zero-value pointed-to
	// value should be considered equal for the purposes of comparison
	NilEqualsZeroValue bool `json:"nil_equals_zero_value"`
```

https://github.com/aws-controllers-k8s/code-generator/blob/ece451c52fa72a9f7cfca3b086b8a4729b592ae4/pkg/config/field.go#L272

This field was created in code generator, but for some unknown reason it was never implemented. I plan to use IsNilEqualsZero to complete implementation.

If we wanted to use IsZeroEqualsNil I would suggest creating a seperate config option to enable this, such as the following

```
	ZeroValueEqualsNil bool `json:"zero_value_equals_nil"`
```

sapphirew (approved):
/lgtm

PR #244, opened 2026-05-11
feat: add DocumentEqual for semantic JSON/YAML string comparison

michaelhtm:

### DIFF
--- a/pkg/compare/document.go
+++ b/pkg/compare/document.go
@@ -1,0 +1,42 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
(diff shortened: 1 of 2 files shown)


pkg/compare/document.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// If either string cannot be parsed as JSON or YAML, falls back to trimmed
+// string comparison.
+func DocumentEqual(a, b string) bool {
+	if a == b {
+		return true
+	}
+	var aObj, bObj interface{}
+	if err := k8syaml.Unmarshal([]byte(a), &aObj); err != nil {
+		return strings.TrimSpace(a) == strings.TrimSpace(b)
```
gustavodiaz7722: Question: if we encounter an error should we return false? What scenario would a white space trim comparison solve an error?

pkg/compare/document_test.go
```diff
@@ -0,0 +1,160 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package compare
+
+import (
+	"testing"
+)
+
+func TestDocumentEqual(t *testing.T) {
```
gustavodiaz7722, line 20: Can you grab some of the unit test from this change for validating specific issues are fixed for json documents?

https://github.com/aws-controllers-k8s/runtime/pull/245

pkg/compare/document.go
michaelhtm: You're right. Handling this the same way we handle iam_policy compare

pkg/compare/document_test.go
michaelhtm: Done

PR #245, opened 2026-05-11
Create json compare utility functions

gustavodiaz7722:
### Summary

Adds `JSONStringEqual` to the `compare` package for semantic comparison of JSON strings. This resolves infinite reconciliation loops caused by formatting differences between JSON values in Kubernetes spec fields and AWS API responses.

### Problem

Controllers like EKS (Addon `configurationValues`) and SNS (Subscription `filterPolicy`) enter perpetual update loops because:
- YAML block scalars (`|`) add trailing newlines that AWS responses don't include
- AWS may return compact JSON while specs contain pretty-printed JSON
- Key ordering may differ between spec and API response

These are semantically identical but fail string comparison.

### Solution

`JSONStringEqual(a, b string) (bool, error)` unmarshals both strings and uses `reflect.DeepEqual` on the parsed structures, making comparison independent of formatting, whitespace, and key order.

This pattern is common and can be found used across multiple controllers where string comparison is insufficient. 

### Testing

22 test cases covering both issue scenarios directly, plus edge cases for invalid JSON, nested structures, key ordering, and whitespace variations.

### Related Issues

- [#2869 (EKS Addon `configurationValues` trailing newline)](https://github.com/aws-controllers-k8s/community/issues/2872)
- [#2877 (SNS Subscription `filterPolicy` formatting)](https://github.com/aws-controllers-k8s/community/issues/2872)

### DIFF
--- a/pkg/compare/json.go
+++ b/pkg/compare/json.go
@@ -1,0 +1,40 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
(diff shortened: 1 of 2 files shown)


gustavodiaz7722:
Looks like we can scrap changes in this PR, since there is already an open PR to implement something similar 

https://github.com/aws-controllers-k8s/runtime/pull/244

We should use the above PR instead since it handles both YAML and JSON documents.

PR #250, opened 2026-06-02
fix: Update K8s Spec for read-only resources when AWS state diverges

### LINKED ISSUE #2856: [eks controller] adopted resources with read only does not update the cluster version
**Describe the bug**
Using EKS controller [v1.11.4](https://github.com/aws-controllers-k8s/eks-controller/releases/tag/v1.11.4).

When an EKS cluster resource is adopted with read only mode using the following annotations

```
apiVersion: eks.services.k8s.aws/v1alpha1
kind: Cluster
metadata:
  annotations:
    services.k8s.aws/adoption-fields: '{"name": "my-test-cluster"}'
    services.k8s.aws/adoption-policy: adopt
    services.k8s.aws/deletion-policy: retain
    services.k8s.aws/read-only: 'true'
```

And an EKS cluster upgrade is performed, the EKS resource's spec.version is not updated and keep showing the old information.

For example, the resource was initially adopted at version `1.33` - the cluster was upgraded to 1.34, but the spec.version shows `1.33` and does not get updated.

The controller logs 

```
{
  "level": "info",
  "ts": "2026-04-13T22:20:17.502Z",
  "logger": "ackrt",
  "msg": "desired resource state has changed, but resource is read-only - skipping update",
  "kind": "Cluster",
  "namespace": "controllers",
  "name": "my-test-cluster",
  "account": "xxxxx",
  "role": "",
  "region": "us-west-2",
  "is_adopted": true,
  "is_read_only": true,
  "generation": 2,
  "skipped": true,
  "diff": [
    {
      "Path": {
        "Parts": [
          "Spec",
          "Version"
        ]
      },
      "A": "1.33",
      "B": "1.34"
    }
  ]
}
```

**Steps to reproduce**

Create an EKS cluster resource - in read only and adoption policy as adopt.  Trigger an upgrade and notice the spec.version does not get updated.

**Expected outcome**
The spec.version should be updated when the cluster resource is adopted in read only mode.

**Environment**

* Kubernetes version: 1.33 and 1.34
* Using EKS (yes/no), if so version?  yes
* AWS service targeted (S3, RDS, etc.) EKS

Mallikarjunadevops:
Fixes aws-controllers-k8s/community#2856

## What is this PR?
This PR fixes a bug where adopted resources with read-only mode enabled do not update their K8s Spec when the underlying AWS resource changes (e.g., an EKS cluster upgrade from 1.33 to 1.34).

## Why?
When a resource is read-only, the `aws-controllers-k8s/runtime` reads the latest state from AWS, compares it to the desired state in K8s, and detects differences. However, previously the reconciler would log `"desired resource state has changed, but resource is read-only - skipping update"` and return the resource WITHOUT patching the K8s API. This caused the K8s resource to become perpetually stale.

With this PR, when the runtime detects a difference in `Spec` for a read-only resource, it will now correctly patch the K8s API with the latest state fetched from AWS.

## Testing Evidence
Ran existing unit test suite in `pkg/runtime/...` via `go test` which successfully passed. Verified the changes apply exactly as intended for read-only sync logic.

### DIFF
--- a/apis/core/v1alpha1/annotations.go
+++ b/apis/core/v1alpha1/annotations.go
@@ -75,11 +75,12 @@ const (
 	// resource manager will leave the AWS resource intact when the K8s resource
 	// is deleted.
 	AnnotationDeletionPolicy = AnnotationPrefix + "deletion-policy"
-	// AnnotationReadOnly is an annotation whose value is a boolean indicating
-	// whether the resource is read-only. If this annotation is set to true on a
-	// CR, that means the user is indicating to the ACK service controller that
-	// the resource is read-only and should not be created/patched/deleted by the
-	// ACK service controller.
+	// AnnotationReadOnly is an annotation indicating whether the resource is
+	// read-only. If this annotation is set to "true" on a CR, the resource
+	// is read-only and will not be created/patched/deleted in AWS, and its
+	// K8s Spec will not be synced. If the value is "always", it acts as
+	// read-only but the controller WILL patch the K8s Spec and Metadata to
+	// reflect the latest observed state from AWS.
 	AnnotationReadOnly = AnnotationPrefix + "read-only"
 	// AnnotationAdoptionPolicy is an annotation whose value is the identifier for whether
 	// we will attempt adoption only (value = adopt-only) or attempt a create if resource 
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -609,14 +609,22 @@ func (r *resourceReconciler) Sync(
 	} else if isReadOnly {
 		delta := r.rd.Delta(desired, latest)
 		if delta.DifferentAt("Spec") {
-			rlog.Info(
-				"desired resource state has changed, but resource is read-only - skipping update",
-				"skipped", true,
-				"diff", delta.Differences,
-			)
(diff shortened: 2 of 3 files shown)


michaelhtm:
Hey @Mallikarjunadevops 
Skipping patching metadata.Spec when the resource is read-only was done by design. We can maybe try using a new annotation, or an annotation value `read-only: always`, that can patch the metadata.Spec
This may require a design PR exploring different options

Mallikarjunadevops:
@michaelhtm Thanks for the review and the insight! That makes total sense.

I have updated the PR to follow your suggestion: instead of always syncing the spec for read-only resources, I updated the logic so it strictly preserves the default `read-only: true` behavior (skips patching). It now specifically checks if the annotation value is `read-only: always` before proceeding to patch the K8s Spec and Metadata.

### Testing Evidence:
- I have run the existing unit test suite via `go test ./pkg/runtime/...` locally and all tests continue to pass cleanly with this modified logic.

Please let me know if this implementation looks good or if you'd still prefer a separate design PR to explore other options!

michaelhtm:
Hey @Mallikarjunadevops 
Just seeing your updated changes. They look good for the most part. I just wanted to get more eyes on it, specifically on exploring different options

Mallikarjunadevops:
@michaelhtm - Thank you for the update!

PR #254, opened 2026-06-05
feat(runtime): Surface cross-namespace deprecation notice via ACK.Advisory condition

sapphirew:
Issue #, if available:

Description of changes:

Surface the Phase 1 cross-namespace deprecation **notice** consistently across all three cross-namespace categories (resource references, field exports, and secret references), and move cross-namespace **validation** for secret references into the runtime so every caller is covered.

### Notice surface: `ACK.Advisory` condition

The deprecation notice is set on the **existing `ACK.Advisory`** condition, carrying the stable reason `CrossNamespaceOptInRequired` (constant `CrossNamespaceOptInRequiredReason`). It is **not** a dedicated condition type.

Rationale: advisory conditions carry no programmatic dependents, which sidesteps the Kubernetes API convention that *removing or repurposing a condition type is a breaking change*. This keeps the temporary notice cheap to add now and cheap to stop setting once the rollout completes, while staying greppable for support. An earlier revision of this PR introduced a dedicated `ACK.CrossNamespaceOptInRequired` type; that type has been removed in favor of `ACK.Advisory`.

- New `condition.SetAdvisory` / `condition.AdvisoryWithReason` helpers — reason-keyed lookup-or-create so distinct advisories on the same resource coexist without clobbering each other.

### Secret references: validation in `SecretValueFromReference`

Cross-namespace validation for secret references lives in `SecretValueFromReference` so every caller is covered, including hand-written custom update functions (e.g. RDS `db_cluster/custom_update.go`) and hooks that call it directly rather than going through generated `sdk.go`.

- When `--enable-cross-namespace` is false and the secret targets a different namespace, return a terminal error (wrapping the `ResourceReferenceCrossNamespaceNotAllowed` sentinel). The caller's existing condition machinery sets `ACK.Terminal` from the message.
- When the flag is true, log a deprecation warning, set the `ACK.Advisory` notice, and resolve the secret from the target namespace (Phase 1 behavior).
- Same-namespace and empty-namespace refs are unaffected.

### Mechanism: context-stash (no signature change)

`SecretValueFromReference` keeps its existing signature `(ctx, ref)`. To reach the owning resource (so the advisory can be set on it) without a breaking interface change, the reconciler stashes the resource in the context:

- `Reconcile()` stashes the resource via `WithConditionManager(ctx, desired)`.
- `SecretValueFromReference` retrieves it via `ConditionManagerFromContext(ctx)`.

> An earlier revision instead added a `ConditionManager` argument to `SecretValueFromReference` (a cleaner signature), but that is a **breaking change** for the hand-written controller callers above and would force every such controller to be updated in lockstep with the runtime release. The context-stash keeps the change additive and non-breaking; mocks are regenerated for the unchanged 2-arg signature.

This lets the code-generator drop the per-call secret validation block it previously emitted in `set_sdk.go` (which only covered generated call sites). Companion change: aws-controllers-k8s/code-generator#699.


pkg/runtime/reconciler.go
```diff
@@ -145,12 +145,32 @@ func (r *reconciler) SecretValueFromReference(
...
+		r.cfg.EnableCrossNamespace,
+		ownerNamespace,
+		ref.Namespace,
+		ref.Name,
+	)
+	if err != nil {
+		return "", ackerr.NewTerminalError(err)
+	}
+	if isCrossNamespace {
+		r.log.V(0).Info(
```
knottnt, line 166: Do we need to also apply a condition to the resource similar to FieldExports and ResourceReferences? If yes, we may need to change the function signature. We'd still need to manually update the `SecretValueFromReference` calls in custom code, but at least they'd fail loudly.

sapphirew: discussed offline - will use condition manager to surface an ACK.Advisory condition for secret references

sapphirew:
The secret path uses the **context-stash** instead — `Reconcile()` stashes the resource via `WithConditionManager`, and `SecretValueFromReference` retrieves it with `ConditionManagerFromContext` to set the notice. Covers all callers (generated sdk.go, hooks, custom update funcs) with no interface change and no controller regeneration churn.

The **ACK.Advisory** notice surface (reason `CrossNamespaceOptInRequired`) is unchanged. Mocks regenerated for the 2-arg signature; full suite green.

sapphirew:
Fixing flaky tests caused by cred expiration issue in https://github.com/aws-controllers-k8s/test-infra/pull/971

pkg/runtime/reconciler.go
```diff
@@ -237,6 +272,12 @@ func (r *resourceReconciler) Reconcile(ctx context.Context, req ctrlrt.Request)
 	// will be reflected in the context.
 	ctx = context.WithValue(ctx, ackrtlog.ContextKey, rlog)
 	ctx = context.WithValue(ctx, "resourceNamespace", req.Namespace)
+	// Stash the resource being reconciled so that code paths which only
+	// receive a context (e.g. SecretValueFromReference) can set conditions on
+	// it without a signature change. The conditions set here are deep-copied
+	// into the returned resource during sdkCreate/sdkUpdate, so they persist
+	// to the status patch.
+	ctx = WithConditionManager(ctx, desired)
```
knottnt, line 280: Q: `desired` can be deep copied a several points during the reconcile loop so the pointer being stashed here may not be the same pointer passed to `sdkCreate/sdkUpdate`. An example, of where this may happen is when a resource is using `adopt-or-create`. In these cases do we still see the Advisory condition applied?

https://github.com/aws-controllers-k8s/runtime/blob/14be98880374b9722ce4c3a077bcee2437664cc1/pkg/runtime/reconciler.go#L568

sapphirew: Discussed offline - added a fix to keep the advisory from being lost when the reconciled resource is deep-copied during Sync.

Problem: The advisory is set on the resource stashed in the reconcile context (the original `desired`). When Sync deep-copies that resource before it flows into `sdkCreate`/`sdkUpdate`, the object that gets patched back (`latest`) is detached from the stashed one. The clearest case is adopt-or-create, which runs resolved = resolved.DeepCopy() before create — so the advisory written via the context lands on desired but never reaches latest, and is not surfaced on the resource.

Fix: Reconcile the advisory from desired onto latest in `HandleReconcileError` — the single choke point that always patches `latest` on every path. Because the context never changes, the advisory is reliably present on desired regardless of how many copies Sync makes, so copying it onto latest right before `patchResourceStatus` closes the gap without re-stashing or any signature change. The copy is idempotent (keyed on type+reason) and preserves all other conditions.

knottnt (approved):

PR #256, opened 2026-06-19
feat(runtime): add ignore-field-drift annotation for selective field reconciliation

sapphirew:
## Description

Adds an opt-in, runtime-level mechanism that lets a user tell ACK to stop reconciling **drift** on specific fields of a custom resource, via a new annotation. This is useful when a field is legitimately managed outside of ACK — for example, organization tooling that applies dynamic tags to a resource that ACK would otherwise try to remove on every reconcile.

The behavior is gated behind a new `IgnoreFieldDrift` feature gate (Alpha, **disabled by default**), so there is no change to existing controllers unless an operator explicitly enables it.

## Behavior

A field listed in the `services.k8s.aws/ignore-field-drift` annotation is **still created from `Spec` and late-initialized normally** — ACK simply stops reconciling drift on it:

- **Create:** the field is sent from `Spec` as usual (this establishes the baseline).
- **Update / drift:** the field is excluded from the reconcile delta, so an out-of-band change on the AWS side does not trigger an Update and is not reset to `Spec`.
- **Spec retained:** the user's declared value stays in the CR `Spec`; ACK never overwrites it with the observed AWS value. Removing the annotation therefore cleanly resumes full reconciliation against the declared value — it is non-destructive.
- **Late initialization:** runs normally for an ignored field. A field that is both ignored and late-initialized but left unset by the user is still populated once from the AWS-resolved value (the baseline), after which drift on it is suppressed; the user's declared value is never overwritten.
- **Observability:** drift on an ignored field is surfaced via a controller log line; `ACK.ResourceSynced` stays `True` (the user opted out of managing those fields).

```yaml
metadata:
  annotations:
    services.k8s.aws/ignore-field-drift: "spec.tags, spec.description"
```

The annotation value is a comma-separated list of dotted, JSON-style field paths.

## Implementation

- New annotation constant `AnnotationIgnoreFieldDrift` (`services.k8s.aws/ignore-field-drift`).
- New `IgnoreFieldDrift` feature gate (Alpha, disabled by default).
- **The delta filter is a runtime-side wrapper, not generated code.** A new `resourceReconciler.filteredDelta(a, b)` wraps `r.rd.Delta` and strips ignored-path differences (`filterIgnoredDeltaDifferences`) before the reconciler consumes the result. This is sufficient because the generated per-resource `newResourceDelta` has exactly **one production caller** fleet-wide — the generated descriptor's `Delta` method, which only the runtime invokes (no controller hooks or custom `sdkUpdate` code call it directly, and no generated `IsSynced` consults a delta — it checks `synced.when` status conditions only). Wrapping the reconciler's delta call sites (the update decision, the read-only skip, and the requeue/patch gate) therefore covers every delta-driven decision **without any generated-code change and without process-global feature gates** — the filter reads the controller's own `cfg.FeatureGates`.
- `applyIgnoredFields` merges the observed value into a deep copy of `desired` before the delta is computed (suppresses drift; avoids clobbering the external value when an unrelated field triggers an Update). `restoreIgnoredFields` puts the user's declared value back before the spec write-back (retain). `filterIgnoredDeltaDifferences` removes ignored-path differences from a computed delta (applied via the `filteredDelta` wrapper above).
- **Anti-clobber holds across custom update paths.** Three mechanisms compose so an ignored field is protected regardless of custom update code: (1) `applyIgnoredFields` puts the AWS-observed value into the `desired` the update code is handed, so re-sending it is a no-op; (2) the filtered delta makes an ignored-only drift an empty delta → `rm.Update` is never called, and any `delta.DifferentAt`-gated send is skipped for ignored fields; (3) `restoreIgnoredFields` resets ignored paths to the declared value before the spec write-back. A real clobber would require a custom update that **both** sends an ignored field *ungated* **and** sources the value from something other than the merged `desired` (a fresh `ReadOne`, `latest`, or the AWS response). A trace of the 64 controllers with custom update logic (incl. `opensearchservice/Domain`, `memorydb/Cluster`+`ACL`, `rds`, `lambda`, `documentdb`, `eks`, `sagemaker`, …) found **no controller that does both** — every generated `newUpdateRequestPayload` builds from the merged `desired`, and the `resp → ko.Spec` copies some controllers do are output-mapping that `restoreIgnoredFields` cleans up, not AWS writes.
- Drift detection for the observability log reuses the **generated per-resource `Delta`** (the raw `r.rd.Delta`, which is unfiltered by construction since the filter lives in the runtime wrapper). This means fields the code-generator compares *semantically* — `is_iam_policy` (`IAMPolicyDocumentEqual`) and `is_document` (`DocumentEqual`) — are judged with their real comparator. AWS canonicalizes such values on read (re-orders keys, collapses single-element arrays, URL-encodes, reindents), so a naive byte compare would report perpetual, spurious drift on every reconcile for an ignored policy/document field; delegating to the generated comparator makes the log fire only on genuine drift.
- Malformed annotation paths are surfaced via a warning log. `warnMalformedIgnorePaths` runs a **syntactic** well-formedness check (a regex over each dotted segment) and logs any path with illegal characters, empty segments, an array index, or a segment not starting with a letter (e.g. `spec/tags`, `spec..tags`, `spec.tags[0]`). Such a path is otherwise a silent no-op, so this warns the user that the field they meant to ignore will keep being reconciled. This does not verify a well-formed path resolves to a real field on the resource (that needs the CRD schema — see follow-ups).
- The stored CR is never mutated for ignored fields except to retain the declared value.

## Testing

- Unit tests for the merge, the declared-value restore, the delta filter, the semantic-vs-byte drift-log behavior for `is_iam_policy` fields, malformed-path detection, and a reconciler-level behavior suite covering create / drift-suppression / anti-clobber / retain / late-initialization / annotation-removal across the annotated and unannotated cases.
- Verified end-to-end on a live account with the iam-controller (feature gate enabled):
  - `ignore-field-drift: "spec.tags"` — an externally-added tag survives reconciliation, the resource stays `Synced`, and the declared value is retained.
  - `ignore-field-drift: "spec.assumeRolePolicyDocument"` (the `is_iam_policy` field) — an out-of-band trust-policy change is not reverted while ignored; no spurious drift log across steady-state reconciles; removing the annotation reconciles the policy back to the declared value.

No code-generator change is required: the delta filtering is applied entirely by the runtime wrapper around the generated `Delta`.

## Known risks / limitations in this MVP

- **Malformed annotation paths are warned, not rejected.** A path with illegal syntax is logged (see Implementation) but not blocked; a well-formed-but-wrong path is still a silent no-op (see follow-ups).

## Scope / follow-ups (not in this PR)

This is the MVP for the first iteration. Deliberately deferred:

- **Schema-level annotation-path validation.** This PR validates path *syntax* and warns on malformed paths (see Implementation). It does not catch a well-formed but *wrong* path (e.g. `spec.tagz` for a resource whose field is `spec.tags`) — that resolves to no field and is still a silent no-op. Validating each path against the CRD schema (and warning on an unknown-but-well-formed path) is deferred; the runtime does not carry the OpenAPI schema at reconcile time, so it is best done in an admission webhook or via a `generator.yaml` allowlist.
- **Rejecting ignore on identity / primary-key fields.**
- **CR-visible drift signal** (a condition) — v1 is log-only, pending multi-advisory condition handling.
- **Member-level / partial-collection management** (`additive-members`) and **sub-field ignore inside list entries**.
- **Guardrail against future custom-update clobbers.** No controller clobbers an ignored field today (see the anti-clobber note in [Implementation](#implementation)), because every custom `sdkUpdate` builds its request from the merged `desired`. That is not enforced, though — a future custom update that sends an ignored field *ungated* from a value that is neither the passed-in `desired` nor the observed state (e.g. a synthesized/stale value from a fresh `ReadOne` or the API response) would clobber. A code-generator lint could flag that pattern so the anti-clobber contract can't silently regress; an optional stronger form re-applies the observed value onto the built request payload before the SDK call.
- Promotion of `IgnoreFieldDrift` from Alpha to Beta after soak.

Refs aws-controllers-k8s/community#2367


sapphirew:
Companion code-generator PR: https://github.com/aws-controllers-k8s/code-generator/pull/714 — it adds the call to `FilterIgnoredDeltaDifferences` in the generated delta template and depends on **this** PR being merged + released first.

sapphirew:
Updated: the annotation is now `services.k8s.aws/ignore-field-drift` and the design is **retain** (the declared value stays in the CR `Spec`; late-init runs normally; only drift is suppressed). Companion code-generator PR: aws-controllers-k8s/code-generator#714 — it depends on this PR being merged and released first.

apis/core/v1alpha1/annotations.go
```diff
@@ -91,4 +91,16 @@ const (
...
+	// "spec.description, spec.someConfig"). For each listed field, the ACK
+	// service controller does not let drift on the named field drive an Update
+	// or overwrite the user's declared value: the field is excluded from the
+	// reconcile delta so observed drift never triggers an Update, and the
+	// declared value is retained in the CR spec rather than being replaced by
+	// the AWS-observed value. The field is still created from the Spec value as
+	// the baseline and is late-initialized normally. This is the declarative
+	// analog of Terraform's lifecycle.ignore_changes and is only honored when
+	// the SelectiveReconciliation feature gate is enabled.
+	AnnotationIgnoreFieldDrift = AnnotationPrefix + "ignore-field-drift"
```
knottnt, line 104: I think we can drop this reference to another IAC tool.

pkg/featuregate/features.go
```diff
@@ -34,16 +37,23 @@ const (
 
 	// IAMRoleSelector is a feature gate for enabling the IAMRoleSelector feature and reconciler.
 	IAMRoleSelector = "IAMRoleSelector"
+
+	// SelectiveReconciliation is a feature gate for enabling selective field
+	// reconciliation via the services.k8s.aws/ignore-field-drift annotation. It
+	// alters reconciliation behavior, so it is disabled by default and must be
+	// explicitly enabled by the controller operator.
+	SelectiveReconciliation = "SelectiveReconciliation"
```
knottnt: nit: From the name alone it isn't immediately clear that this is related to the the ignored field annotations.

pkg/featuregate/features.go
```diff
@@ -101,6 +111,37 @@ func (fg FeatureGates) GetFeatureNames() []string {
 	return names
 }
 
+// globalFeatureGates holds the process-wide feature gate configuration. It is
+// set once during controller startup (via SetGlobalFeatureGates, called from
+// the runtime's BindControllerManager) and read by generated, package-level
+// code that has no access to the controller's config object (e.g. the
+// per-resource newResourceDelta function). It defaults to the ACK default
+// feature gates so that code reading it before initialization sees sane,
+// disabled-by-default values.
```
knottnt: Q: When the feature flag is set to a non-default value is there a race condition here where some controller code would still see the default value?

pkg/runtime/reconciler.go
```diff
@@ -725,6 +731,14 @@ func (r *resourceReconciler) ensureConditions(
 		synced := false
 		condMessage := ackcondition.NotSyncedMessage
 		var condReason string
+		// Selective reconciliation: rm.IsSynced (generated per-resource code)
+		// derives the in-sync determination from the per-resource
+		// newResourceDelta. That generated delta now calls
```
knottnt: Is this the case? I believe the latest resource is what we use for the IsSynced call. I would expect Delta filtering to impact this.

pkg/runtime/reconciler.go
```diff
@@ -939,20 +953,65 @@ func (r *resourceReconciler) updateResource(
...
+				"failed to apply ignore-field-drift annotation; "+
+					"reconciling authoritatively",
+				"error", mergeErr.Error(),
+			)
+		} else {
+			reconcileDesired = merged
+			r.logIgnoredFieldDrift(ctx, desired, latest)
+		}
+		// Warn about any ignore-field-drift paths that are not well-formed field
+		// paths (typos / illegal characters); they silently have no effect.
```
knottnt: Should we apply a terminal error here? If there's an invalid field path and we continue the reconcile loop we could end up mutating a field that the user intended for us to ignore. In that case failing fast and throwing an error seems safer.

pkg/runtime/ignore_field_drift.go
```diff
@@ -0,0 +1,587 @@
...
+// (which the runtime cannot see; see the design doc's validation follow-up).
+var fieldPathSegmentRegex = regexp.MustCompile(`^[A-Za-z][A-Za-z0-9_]*$`)
+
+// isValidFieldPath reports whether p is a well-formed dotted JSON-style field
+// path: one or more non-empty segments joined by ".", each matching
+// fieldPathSegmentRegex. It rejects malformed paths such as ones with illegal
+// characters (e.g. "spec/tags", "spec.tags!"), empty segments (e.g. "spec..tags",
+// ".spec", "spec."), array indices (e.g. "spec.tags[0]" -- sub-element ignore is
+// out of v1 scope), and the empty string.
+func isValidFieldPath(p string) bool {
```
knottnt, line 311: Would we be able to validate this against the actual fields available in the Resource Kind's type?

pkg/featuregate/features.go
knottnt: See [comment](https://github.com/aws-controllers-k8s/code-generator/pull/714#discussion_r3562284515) on the code-generator PR. I'm wondering if we can drop this entirely by bringing the FilterIgnoredDeltaDifferences into the runtime framework.

knottnt (commented):
@sapphirew left a few initial comments.

apis/core/v1alpha1/annotations.go
sapphirew: Dropped in [be084aa](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/be084aa757619985232ebece5e327442d194fbab).

pkg/featuregate/features.go
sapphirew: Renamed to `IgnoreFieldDrift` in [d3456f7](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/d3456f74b4a869c42594b7775e3fcd57dc643f34) to match the annotation. The gate never shipped, so no compatibility impact.

pkg/featuregate/features.go
sapphirew: Gone with your wrapper suggestion: the filter now reads `cfg.FeatureGates` and the global accessors are deleted in [8e9e706](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/8e9e70653007367c3e9f96470d480122d36b7c4b).

pkg/runtime/reconciler.go
sapphirew: You were right, the comment was wrong. Fixed in [8e9e706](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/8e9e70653007367c3e9f96470d480122d36b7c4b). Generated `IsSynced` never uses the delta, it returns true or checks `synced.when` conditions. Ignored drift stays out of `ResourceSynced` upstream: the filtered delta means `rm.Update` is never called.

pkg/runtime/reconciler.go
sapphirew: We chose warn only for the MVP. A terminal error stops managing the whole resource over one typo, and recovery needs a spec edit since annotation changes alone do not trigger a reconcile. A syntax check also is not the real safety line: a well formed typo like `spec.tagz` passes it. The real fix is schema validation at admission time, per your other comment.

pkg/runtime/ignore_field_drift.go
sapphirew: Yes, deferred as a follow up. The runtime lacks the CRD schema at reconcile time, so this belongs in an admission webhook or a codegen emitted allowlist. Tracked in the PR description.

pkg/featuregate/features.go
sapphirew: Done in [8e9e706](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/8e9e70653007367c3e9f96470d480122d36b7c4b). Verified nothing bypasses it: the generated delta's only caller is the descriptor's `Delta`, invoked only by the runtime; no hooks or `IsSynced` use it. New `filteredDelta` wraps `r.rd.Delta` at the reconciler's decision points. Closed aws-controllers-k8s/code-generator#714 as no longer needed.

pkg/runtime/ignore_field_drift.go
```diff
@@ -0,0 +1,585 @@
...
+// MergeFrom diff carry the value, so the patch persists it once.
+//
+// A path is rebased ONLY when BOTH:
+//
+//   - the resource carries an ACK.LateInitialized condition — i.e. late-init is
+//     actually configured for the resource and ran. This is the signal that the
+//     value on the ignored field is a late-initialized (service-defaulted)
+//     value rather than plain post-creation drift. It is precisely what
+//     distinguishes "late-init wins" (persist, matrix Row 6) from drift
+//     suppression on a non-late-init ignored field (do NOT persist, Row 5); AND
```
knottnt: nit: Row 5/6 doesn't refer to anything in the codebase and can be removed.

pkg/runtime/ignore_field_drift_test.go
knottnt, line 1: Would we be able to add test cases for nest fields like `spec.a.b.c` ? It looks like most of the tests cover only top level fields.

pkg/runtime/reconciler.go
```diff
@@ -725,6 +731,12 @@ func (r *resourceReconciler) ensureConditions(
 		synced := false
 		condMessage := ackcondition.NotSyncedMessage
 		var condReason string
+		// Note: rm.IsSynced (generated per-resource code) checks
+		// `synced.when` status conditions only and never consults the
+		// resource delta, so ignore-field-drift needs no suppression here.
+		// Drift on ignored fields is kept out of the synced determination
+		// upstream, where the reconciler's delta-driven decisions
+		// (update/requeue) go through filteredDelta.
```
knottnt: nit: we can likely drop this comment since it isn't covering new behavior. Generally, synced behavior is based on status and not user set spec as well.

pkg/runtime/reconciler.go
knottnt: IMO it would be good to fail fast here instead of preceding. Agree that the current check doesn't cover every malformed input and validating that the path matches a valid field in the resource schema would be a good add. However, we do know that if this check fails something is wrong with the user input and we may be mutating something the user doesn't want us to. 

As for the reconcile not being triggered by an annotation change. I believe we do have the option of adding a custom predicate [here](https://github.com/aws-controllers-k8s/runtime/blob/4b18c8cbcf77d0eda191c3eb1ece194cc52f733a/pkg/runtime/reconciler.go#L114) that could allow us to trigger on more than just spec/generation changes.

pkg/runtime/ignore_field_drift.go
sapphirew: Done — removed the Row 5/6 references in [f4841cb](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/f4841cb).

pkg/runtime/ignore_field_drift_test.go
sapphirew: Done — added four tests exercising deeply nested paths (`spec.a.b.c` style) in [f4841cb](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/f4841cb):

- `TestApplyIgnoredFields_NestedPath` — merges `spec.network.vpc.cidr` from latest, preserves sibling fields at same depth
- `TestApplyIgnoredFields_NestedPathAbsentInLatest` — handles nested path absent in latest
- `TestRestoreIgnoredFields_NestedPath` — restores declared value for a nested path after update
- `TestFilterIgnoredDeltaDifferences_NestedPaths` — strips a nested ignored delta entry while preserving siblings at the same nesting depth

pkg/runtime/reconciler.go
sapphirew: Done — removed the comment in [f4841cb](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/f4841cb).

pkg/runtime/reconciler.go
sapphirew: Agreed — changed to fail fast with a `TerminalError` in [f4841cb](https://github.com/aws-controllers-k8s/runtime/pull/256/commits/f4841cb). The reconciler now checks `malformedIgnorePaths` before any merge/delta logic and returns immediately if the annotation is syntactically invalid. This puts the resource into an error condition, making it visible to the user that their annotation needs fixing.

Also added `AnnotationChangedPredicate` to the controller builder (via `predicate.Or(GenerationChangedPredicate{}, AnnotationChangedPredicate{})`) as you suggested, so annotation changes now trigger an immediate reconcile rather than waiting for the next resync or spec change. This means the terminal error surfaces right away when a user sets a bad annotation.

Added `TestIgnoreFieldDrift_MalformedPath_FailsFast` to verify the behavior end-to-end (terminal error returned, Update never called).

pkg/featuregate/features.go
```diff
@@ -34,6 +36,12 @@ const (
 
 	// IAMRoleSelector is a feature gate for enabling the IAMRoleSelector feature and reconciler.
 	IAMRoleSelector = "IAMRoleSelector"
+
+	// IgnoreFieldDrift is a feature gate for enabling selective field
+	// reconciliation via the services.k8s.aws/ignore-field-drift annotation. It
+	// alters reconciliation behavior, so it is disabled by default and must be
+	// explicitly enabled by the controller operator.
```
knottnt, line 43: nit: comment is a bit verbose.

```suggestion
	// IgnoreFieldDrift is a feature gate for enabling selective field
	// reconciliation via the services.k8s.aws/ignore-field-drift annotation.
```

pkg/runtime/ignore_field_drift.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+	l, err := k8srt.DefaultUnstructuredConverter.ToUnstructured(latest.RuntimeObject())
+	if err != nil {
+		return desired, err
+	}
+
+	for _, p := range ignorePaths {
+		parts := pathParts(p)
+		if v, found, err := unstructured.NestedFieldCopy(l, parts...); err == nil && found {
+			_ = unstructured.SetNestedField(d, v, parts...)
```
knottnt: Q: What's the behavior here when the parent of an ignored field is not present in desired, but is in latest? For example if I've ignored the field `spec.network.vpc.cidr`, but my desired spec only has `spec.network` what happens if the`spec.network.vpc.cidr` is found in latest? Does this function populate empty values for the rest of `spec.network.vpc`?

pkg/runtime/ignore_field_drift.go
```diff
@@ -0,0 +1,559 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// toDeltaPath converts a JSON-style annotation path (e.g. "spec.tags") into the
+// capitalized form used by the generated Delta paths (e.g. "Spec.Tags"). It
+// title-cases the first letter of each dotted segment, which is sufficient for
+// the top-level field paths these annotations target.
```
knottnt: A couple things here
1. Are we only targeting top level fields as this comment suggests?
2. I think this logic will break down for fields that contain acronyms. In those cases more than just the first character will be capitalized. An example of such a field would be [KMSKeyID](https://github.com/aws-controllers-k8s/rds-controller/blob/2feea62780b5ce80ed2ba48050cc4241a15af0cb/apis/v1alpha1/db_cluster.go#L381) in RDS DBCluster

pkg/runtime/reconciler.go
```diff
@@ -112,7 +112,10 @@ func (r *resourceReconciler) BindControllerManager(mgr ctrlrt.Manager) error {
 	).For(
 		rd.EmptyRuntimeObject(),
 	).WithEventFilter(
-		predicate.GenerationChangedPredicate{},
+		predicate.Or(
+			predicate.GenerationChangedPredicate{},
+			predicate.AnnotationChangedPredicate{},
```
knottnt: Should the inclusion of this predicate be behind the feature gate?

pkg/runtime/ignore_field_drift.go
knottnt: Q: What's the reason for swallowing this error?

pkg/runtime/ignore_field_drift.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+	d, err := k8srt.DefaultUnstructuredConverter.ToUnstructured(desired.RuntimeObject())
+	if err != nil {
+		return err
+	}
+
+	for _, p := range paths {
+		parts := pathParts(p)
+		if v, found, err := unstructured.NestedFieldCopy(d, parts...); err == nil && found {
+			_ = unstructured.SetNestedField(u, v, parts...)
```
knottnt: Q: Should we be swallowing this error?

pkg/runtime/reconciler.go
```diff
@@ -939,20 +948,68 @@ func (r *resourceReconciler) updateResource(
...
 
+	// If the IgnoreFieldDrift feature is enabled and the resource carries
+	// the ignore-field-drift annotation, merge the observed (latest) values for
+	// the ignored fields into a copy of desired. The merged copy is used for
+	// delta computation and the Update call so that drift on ignored fields does
+	// not drive an Update. The stored CR is never mutated.
+	reconcileDesired := desired
+	if r.cfg.FeatureGates.IsEnabled(featuregate.IgnoreFieldDrift) &&
+		HasIgnoreFieldDrift(desired) {
+		if bad := malformedIgnorePaths(desired); len(bad) > 0 {
```
knottnt: Q: Should we perform this check earlier in the reconcile loop? For example, should we proceed with a create operation if this annotation is malformed?

pkg/featuregate/features.go
sapphirew: Applied in 467b03b.

pkg/runtime/ignore_field_drift.go
sapphirew: No — `SetNestedField` creates only the intermediate maps needed to reach the leaf and sets just that leaf from `latest`; undeclared siblings (e.g. `spec.network.vpc.region`) aren't populated. Added `TestApplyIgnoredFields_ParentAbsentInDesired` in 467b03b.

pkg/runtime/ignore_field_drift.go
sapphirew: Both fixed in 467b03b. (1) Nested paths are supported — that comment was stale and is gone. (2) Real bug: `toDeltaPath` title-cased only the first letter, so `spec.kmsKeyID` → `Spec.KmsKeyID` never matched `Spec.KMSKeyID` and drift went unsuppressed. Deleted it and now match segments case-insensitively via new `compare.Path.ContainsFold` (unambiguous since Go forbids exported fields differing only by case). Tests added.

pkg/runtime/reconciler.go
sapphirew: Gated in 467b03b — `AnnotationChangedPredicate` is added only when `IgnoreFieldDrift` is enabled, so non-users keep the prior trigger behavior.

pkg/runtime/ignore_field_drift.go
sapphirew: No good reason — fixed in 467b03b. Swallowing would leave drift unsuppressed (merge path) or overwrite the declared value with the AWS value (restore path). Both call sites (plus `rebaseIgnoredFieldsForPersist`) now propagate the error.

pkg/runtime/ignore_field_drift.go
sapphirew: Fixed in 467b03b along with the sibling call at line 105 — see that thread. `SetNestedField` only errors on a path/type collision; both now propagate rather than swallow.

pkg/runtime/reconciler.go
sapphirew: Agreed — moved to early in `Sync` (before any AWS call) in 467b03b, so create/update/read-only are all covered; removed the redundant check from `updateResource`.

pkg/compare/path_test.go
```diff
@@ -0,0 +1,49 @@
...
+	// Exact match.
+	assert.True(t, p.ContainsFold("Spec.KMSKeyID"))
+	// Case-insensitive per segment: the JSON/spec name folds onto the
+	// acronym-cased Go field name.
+	assert.True(t, p.ContainsFold("spec.kmsKeyID"))
+	assert.True(t, p.ContainsFold("SPEC.kmskeyid"))
+	// Prefix (at-or-under) matching, folded.
+	assert.True(t, p.ContainsFold("spec"))
+	// Longer than the path -> no match.
+	assert.False(t, p.ContainsFold("spec.kmsKeyID.extra"))
```
knottnt, line 46: Q: Do we want to return false when a sub-field of the ignored path is in the delta? If I've ignored `Spec.Foo.Bar` and `Bar` is an object it seems like any field under `Bar` should be implicitly ignored and also removed from the delta.

knottnt: Nvm, I see that the `filterIgnoredDeltaDifferences` is using the Delta path as the caller so this logic has the effect of implicitly ignoring children of an ignored path.

pkg/runtime/ignore_field_drift_test.go
```diff
@@ -192,6 +192,182 @@ func TestApplyIgnoredFields_IgnoreFieldAbsentInLatest(t *testing.T) {
...
+		logging, ok := config.(map[string]interface{})["logging"]
+		if ok {
+			_, found := logging.(map[string]interface{})["level"]
+			assert.False(t, found, "nested ignored field should be removed when absent in latest")
+		}
+	}
+}
+
+// TestApplyIgnoredFields_ParentAbsentInDesired covers the case raised in review:
+// the parent of an ignored path is absent in desired but present in latest
```
knottnt: nit: no need to mention the review comment in the test comment unless we want to include a link. Outside of this PR the current comment lacks context on what was being discussed.

pkg/runtime/ignore_field_drift_test.go
```diff
@@ -192,6 +192,182 @@ func TestApplyIgnoredFields_IgnoreFieldAbsentInLatest(t *testing.T) {
...
+	// Sibling that desired declared is preserved.
+	assert.Equal(t, "keep", network["name"])
+	// The intermediate "vpc" map is created to reach the leaf.
+	vpc := network["vpc"].(map[string]interface{})
+	// The ignored leaf takes the AWS value.
+	assert.Equal(t, "10.1.0.0/16", vpc["cidr"])
+	// Only the leaf is set: the sibling "region" that desired never declared is
+	// NOT copied over from latest (the merge is scoped to the ignored path).
+	_, hasRegion := vpc["region"]
+	assert.False(t, hasRegion, "only the ignored leaf should be set, not its siblings")
```
knottnt, line 296: IMO this is an a potentially interesting edge case for users that we may want to at least call out in documentation. I don't necessarily think the behavior here is wrong since it matches what the user declared, but it is a case where user could unexpectedly zero out a value unintentionally.

knottnt: In cases where the sibling values are generally defaulted by AWS I'm assuming that we'd still propagate those defaults to the desired manifest either via the Create response or late-init fields.

pkg/runtime/ignore_field_drift_test.go
sapphirew: Updated comments for this test

pkg/runtime/ignore_field_drift_test.go
sapphirew: Good call, will make sure to cover it in the doc. In this case if a field is late-initialized or comes back on the create response, it's already in desired, so the merge preserves it and there's nothing to zero. The gap is only fields that are neither declared nor late-initialized.

knottnt (approved):
/lgtm

PR #262, opened 2026-07-23
fix: register per-resource FieldExport reconcilers

### LINKED ISSUE #2964: runtime: variable shadowing disables the per-resource FieldExport reconcilers, exports never retry after creation
**Affected component:** ACK runtime v0.61.0 (observed via apigatewayv2-controller v1.3.3, applies to every controller embedding this runtime)
**File:** `pkg/runtime/service_controller.go` (`BindControllerManager`)

**Describe the bug**

```go
exporterInstalled := false
exporterLogger := c.log.WithName("exporter")

if cfg.EnableFieldExportReconciler {
    exporterInstalled, err := c.GetFieldExportInstalled(mgr) // := shadows the outer variable
    ...
}
...
if cfg.EnableFieldExportReconciler && exporterInstalled {    // outer variable, always false
    feRec := NewFieldExportReconcilerForAWSResource(...)
```

The short variable declaration inside the `if` block shadows the outer `exporterInstalled`, which therefore stays `false`, so the per-resource `field-export.<kind>` reconcilers (the ones that re-run exports whenever a source resource changes, bound with `ResourceVersionChangedPredicate`) are never registered, regardless of configuration. Observable in any controller on this runtime version: the `fieldexport` controller logs its startup, but no `field-export.*` named controllers ever appear.

Consequence: a `FieldExport` is only evaluated at its own creation (and own spec changes). In the common GitOps case where the FieldExport and its source resource are applied in the same sync, the exported status field does not exist yet at that moment, the initial attempt fails with `FieldExportPathDoesNotExist` wrapped in `requeue.None`, and the export is never attempted again. The target ConfigMap or Secret stays empty forever, with nothing surfaced (the success path `onSuccess` is also a no-op, so there is no condition to distinguish "exported" from "never ran").

This is very likely the root cause of aws-controllers-k8s/community#2135 (FieldExport is not populating the ConfigMap), where the reporter's source DBInstance was also not synced yet at FieldExport creation time, and of the reports there that a controller restart "fixes" it (a restart replays the FieldExport create event after the source has its field).

**Steps to reproduce**

1. Apply an ACK resource and a `FieldExport` referencing one of its status fields (for example `.status.apiEndpoint` of an apigatewayv2 `API`) in the same kubectl apply or ArgoCD sync.
2. Wait for the source resource to sync and the status field to appear.
3. The FieldExport target ConfigMap never receives the value.
4. Delete and recreate the FieldExport: the value is exported immediately, confirming that only the create-time path runs.

**Expected outcome**

The per-resource field export reconcilers register (as the code intends) and re-drive exports when source resources change, so a FieldExport created before its source field materializes converges once the field appears. Suggested fix: assign instead of redeclaring inside the block, for example declare `err` separately or use `=` for `exporterInstalled`.

**Environment**

* Kubernetes version: v1.35.6-eks-8f14419
* Using EKS (yes/no), if so version? Yes, v1.35
* AWS service targeted (S3, RDS, etc.): API Gateway v2 (apigatewayv2-controller v1.3.3, Helm chart install, IRSA), applies to all services

michaelhtm:
Description of changes:
Assign instead of shadowing exporterInstalled so the per-resource
field-export reconcilers register and exports retry after creation.

Fixes aws-controllers-k8s/community#2964

### DIFF
--- a/pkg/runtime/service_controller.go
+++ b/pkg/runtime/service_controller.go
@@ -244,7 +244,7 @@ func (c *serviceController) BindControllerManager(mgr ctrlrt.Manager, cfg ackcfg
 	exporterLogger := c.log.WithName("exporter")
 
 	if cfg.EnableFieldExportReconciler {
-		exporterInstalled, err := c.GetFieldExportInstalled(mgr)
+		exporterInstalled, err = c.GetFieldExportInstalled(mgr)
 		if err != nil {
 			exporterLogger.Error(err, "unable to determine if the FieldExport CRD is installed in the cluster")
 		} else if !exporterInstalled {
--- a/pkg/runtime/service_controller_test.go
+++ b/pkg/runtime/service_controller_test.go
@@ -15,7 +15,9 @@ package runtime
 
 import (
 	"context"
+	"encoding/json"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


knottnt (commented):
Change looks good, but having a unit test to guard against regression would be good.

knottnt (approved):
/lgtm

PR #264, opened 2026-08-05
feat: Add client-side SDK rate limiting via --sdk-max-tps flag

NicholasBlaskey:
Issue #, if available:

Description of changes:

Adds a configurable client-side rate limiter that wraps the AWS SDK HTTP client with a token bucket. This enforces a maximum request rate (TPS) across all AWS API calls, preventing controllers from exceeding service- level rate limits.

New flags:
  --sdk-max-tps=N    Maximum requests per second (0 = disabled, default)
  --sdk-max-burst=N  Burst size for rate limiter (default: 5)

Use case: services like Route53 enforce a hard 5 req/s account-level limit shared across all API actions. Setting --sdk-max-tps=3 ensures the controller never consumes more than 3 req/s, leaving headroom for other consumers (external-dns, Terraform, etc.) sharing the same account.

The rate limiter applies to all AWS API calls from the controller's SDK client (the primary service), but not to STS calls for credential refresh which use a separately constructed client.

Tested and verified Route53 runs with no more than 3 TPS with a 3 limit

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -57,6 +57,8 @@ const (
 	flagUnsafeAWSEndpointURLs           = "allow-unsafe-aws-endpoint-urls"
 	flagAWSEndpointUsePathStyle         = "aws-endpoint-use-path-style"
 	flagHTTPClientTimeout               = "http-client-timeout"
+	flagSDKMaxTPS                       = "sdk-max-tps"
+	flagSDKMaxBurst                     = "sdk-max-burst"
 	flagLogLevel                        = "log-level"
 	flagResourceTags                    = "resource-tags"
 	flagWatchNamespace                  = "watch-namespace"
@@ -104,6 +106,8 @@ type Config struct {
 	AllowUnsafeEndpointURL          bool
 	UsePathStyle                    bool
 	HTTPClientTimeout               time.Duration
+	SDKMaxTPS                       float64
+	SDKMaxBurst                     int
 	LogLevel                        string
 	ResourceTags                    []string
 	ResourceTagKeys                 []string
@@ -223,6 +227,17 @@ func (cfg *Config) BindFlags() {
 		&cfg.HTTPClientTimeout, flagHTTPClientTimeout,
 		60*time.Second,
 		"Timeout for HTTP requests made by the AWS SDK client. Set to 0 to disable.",
+	)
+	flag.Float64Var(
+		&cfg.SDKMaxTPS, flagSDKMaxTPS,
+		0,
+		"Maximum AWS SDK requests per second (client-side rate limit). Set to 0 to disable. "+
+			"Useful for services with low account-level API rate limits (e.g. Route53 at 5 req/s).",
+	)
+	flag.IntVar(
+		&cfg.SDKMaxBurst, flagSDKMaxBurst,
+		5,
+		"Maximum burst size for SDK rate limiting. Only used when --sdk-max-tps is set.",
 	)
 	flag.StringVar(
 		&cfg.LogLevel, flagLogLevel,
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -224,6 +228,19 @@ func (cfg *Config) BindFlags() {
 		60*time.Second,
 		"Timeout for HTTP requests made by the AWS SDK client. Set to 0 to disable.",
 	)
+	flag.Float64Var(
+		&cfg.SDKMaxTPS, flagSDKMaxTPS,
+		0,
+		"Maximum AWS SDK requests per second (client-side rate limit). Set to 0 to disable. "+
+			"Useful for services with low account-level API rate limits (e.g. Route53 at 5 req/s). "+
+			"Can also be set via ACK_SDK_MAX_TPS environment variable.",
```
michaelhtm: where do we consume the environment variable?

pkg/runtime/config.go
```diff
@@ -76,8 +77,26 @@ func (c *serviceController) NewAWSConfig(
...
 	}
+
+	// Wrap the HTTP client with a rate limiter if configured.
+	// This provides client-side throttling to avoid exceeding AWS service
+	// API rate limits (e.g. Route53's 5 req/s account-level limit).
+	var sdkHTTPClient aws.HTTPClient
+	if c.cfg.SDKMaxTPS > 0 {
+		burst := c.cfg.SDKMaxBurst
+		if burst <= 0 {
+			burst = int(c.cfg.SDKMaxTPS) + 1
```
michaelhtm: not sure if this is okay..if burst is a negative number it may still remain negative. Can we instead validate these flags here: https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/config/config.go#L344

pkg/runtime/config.go
```diff
@@ -109,3 +128,18 @@ func formatUserAgent(name, version string, extra ...string) string {
...
+
+// rateLimitedHTTPClient wraps an aws.HTTPClient with a token bucket rate
+// limiter. It enforces a maximum request rate (TPS) across all AWS SDK API
+// calls made through this client.
+type rateLimitedHTTPClient struct {
+	client  aws.HTTPClient
+	limiter *rate.Limiter
+}
+
+func (c *rateLimitedHTTPClient) Do(r *http.Request) (*http.Response, error) {
```
michaelhtm, line 133: is the rate limiting specific per API? (eg. create and read have separate token buckets)

pkg/config/config.go
NicholasBlaskey: removed these. My mistake I thought the flag library was like vulcan and automatically converted these.

pkg/runtime/config.go
NicholasBlaskey: added validation there

pkg/runtime/config.go
NicholasBlaskey: Right now its specific to client

To be fair, for Route53 the read and write APIs take out of the same quota but yes for other services it might be per write or per API

knottnt:
@NicholasBlaskey researching the aws-sdk-go-v2 it does look like they have support for an "adaptive" retry mode which enables an additional client-side rate limiter. It's a bit coarser, but wondering if user could get similar benefits since the retry mode can be enabled already with an [environment variable](https://github.com/aws/aws-sdk-go-v2/blob/bbecb94b8f4abeab32d24a18f8e469421b4ec603/config/env_config.go#L72)?

NicholasBlaskey:
> @NicholasBlaskey researching the aws-sdk-go-v2 it does look like they have support for an "adaptive" retry mode which enables an additional client-side rate limiter. It's a bit coarser, but wondering if user could get similar benefits since the retry mode can be enabled already with an [environment variable](https://github.com/aws/aws-sdk-go-v2/blob/bbecb94b8f4abeab32d24a18f8e469421b4ec603/config/env_config.go#L72)?

My understanding of adaptive retry mode is it essentially works off throttling signals. 

This doesn't really help noisy neighbor since it still fundamentally causes throttles. It will find the max TPS it can take and use that amount.

I suppose if both external-dns and route53 used it they could share the quota better?

PR #263, opened 2026-08-03
Update golang.org/x/net and golang.org/x/text  dependencies

knottnt:
Issue #, if available: [2989](https://github.com/aws-controllers-k8s/community/issues/2989)

Description of changes:
Update dependencies listed in CVE scan.

- upgrade golang.org/x/text  to v0.39.0
- upgrade golang.org/x/net to v0.56.0

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #265, opened 2026-08-24
fix: give Update a copy of desired carrying the observed status

gustavodiaz7722:
Issue #, if available: none filed yet.

Description of changes:

`updateResource` handed the stored `desired` resource straight to `rm.Update`. ACK permits a resource manager's `Update` to return the object it was given, and hand-written `customUpdate` implementations do — ecr-controller's `customUpdateRepository` and ec2-controller's `customUpdateManagedPrefixList` both end in `return desired, nil`.

When that happens `latest` and `desired` are the same object, and that object is also the base of the status merge patch:

```go
dobj := desired.DeepCopy().RuntimeObject()   // base
lobj := latest.DeepCopy().RuntimeObject()    // target
patch := client.MergeFrom(dobj)
```

Diffing an object against itself yields `{}`, so the status write reaches the API server as a no-op and any status the manager just set — conditions, observed state — is discarded.

### Effect

A resource stuck on a stale condition. A controller that writes `ResourceSynced=False` from a `sdk_create_post_set_output` hook and relies on the update path to clear it never gets it cleared by the reconcile that performs the update. Two things compound: the status patch is `{}`, so etcd keeps `False`; and the in-memory condition *is* `True`, so `handleRequeues` takes the synced branch and requeues after `resyncPeriod` instead of the 30s `TemporaryOutOfSync` delay.

Measured on a live cluster with eventbridge-controller at the commit from aws-controllers-k8s/eventbridge-controller#102 that precedes its own controller-side fix, so both arms ran identical controller code and the runtime was the only variable. 60s resync, `EventBus` created with `spec.policy`:

| runtime | policy live in AWS | CR `Synced=True` | stale window |
|---|---|---|---|
| released v0.62.0 | t+33s | t+90s | 63s (one resync period) |
| this change | t+33s | t+33s | 0s |

At the chart default `reconcile.defaultResyncPeriod: 36000` that window is up to 10 hours, during which anything gating on `Ready` sees an unhealthy resource while AWS is fully configured.

ecr-controller has both halves on `main` today: its create hook writes `Synced=False` when `spec.policy` or `spec.lifecyclePolicy` is set, and `customUpdateRepository` returns `desired` uncopied.

Observed state is dropped the same way. ec2-controller's `customUpdateManagedPrefixList` sets `Status.State` and `Status.Version` from the `ModifyManagedPrefixList` response and then returns `desired`, so neither write reaches etcd.

### Fix

Two changes, both in `updateResource`:

1. **Deep-copy `desired` before handing it to `Update`**, so a manager that returns what it was given cannot alias the patch base. This generalizes the precedent immediately below it — the ignore-field-drift path already replaces `reconcileDesired` with a merged copy, and so already avoids this aliasing whenever that gate is enabled.

2. **Apply the observed status to that copy, keeping the copy's own conditions.** `desired` carries the status last persisted to etcd, while `latest` carries what the service reports now; an `Update` should reason about observed state. This runs after the ignore-field-drift branch, which replaces the copy with one derived from `desired` and would otherwise discard it.

`AWSResource.SetStatus` cannot express change 2 on its own. `Status` holds both service-reported state and ACK's conditions, and `SetStatus` replaces the whole struct, so it cannot adopt observed state without also adopting `latest`'s conditions. A `setStatusWithoutConditions` helper withholds them, because neither side's conditions belong on the copy:

- Adopting `latest`'s would suppress recomputation. `resetConditions` clears conditions each reconcile so `ensureConditions` can rebuild them, and `ensureConditions` only runs while `Synced` is nil — a `ReadOne` hook that marks a resource not-yet-synced would stick. ec2-controller's `VPCEndpoint` does exactly this while an endpoint is pending.
- Dropping the copy's own would lose conditions set after `resetConditions`, notably `ACK.ReferencesResolved`.

The result is what the released runtime already gives `Update` condition-wise, plus the observed non-condition status.

Scoped to the `Update` call site rather than to the patch base, which keeps `desired` as the base, leaves steady-state write behaviour unchanged for every controller, and requires no new exported API. Generated `sdkUpdate` implementations are unaffected either way — they already return `&resource{ko}` built from `desired.ko.DeepCopy()`.

### Prerequisite

aws-controllers-k8s/ec2-controller#368 — merged.

Before that landed, `ec2-controller-test` failed on `test_managed_prefix_list.py::test_update_entries` and `::test_update_tags`. Both were true positives this change exposed rather than caused.

`ManagedPrefixList` compared `Spec.Entries` with an order-sensitive `DeepEqual` while `GetManagedPrefixListEntries` returns them in an order AWS chooses, so the delta never cleared. `customUpdateManagedPrefixList` re-issued `ModifyManagedPrefixList` carrying `CurrentVersion` with no entry modification on every reconcile, which AWS rejects with `InvalidParameterCombination`. That loop was present with or without this change, since the delta and the request are identical either way. What this change altered is only whether the CR reported it: `customUpdateManagedPrefixList` ended in `return desired, nil`, so on the released runtime the returned object was also the status merge-patch base, the patch was `{}`, and the CR kept the stale `Synced=True` from creation — which the test's `wait_on_condition(..., "True")` matched.

Confirmed on a live cluster: with this change and an unfixed ec2-controller, the CR sat at `Synced=Unknown` with reason `InvalidParameterCombination` for 40 polls while AWS reported `modify-complete` with all four entries. With #368 applied it reached `Synced=True` in ~35s and held it across a resync period. Same job and same tests for comparison: build `2084372351612358656` on #263 (released runtime) passed both, build `2092394037469253632` here failed both.

### Known impact

Change 2 alters what an `Update` observes in `desired.Status`, from the last-persisted status to the observed one. One resource in the fleet depends on the difference between the two: quicksight-controller's `Dashboard` compares `a.ko.Status.VersionNumber` against `b.ko.Status.VersionNumber` in its `delta_post_compare` hook, and `sdkUpdate` uses that delta to decide whether to publish a draft dashboard version.

With the observed status applied, the two sides are equal, the delta no longer fires, and `UpdateDashboardPublishedVersion` is never called — a dashboard would get a new draft version that is never published. Confirmed against `newResourceDelta` directly: modelling draft version 2 with published version 1, `DifferentAt("Status.VersionNumber")` is `true` before and `false` after.

No other controller reads the first delta argument's status or branches on a `Status.*` difference. A follow-up will move Dashboard's publish decision onto observed state (`ListDashboardVersions` against the published version) so it no longer relies on persisted status.

### Testing

`TestReconcilerUpdate_PassesCopyOfDesiredToUpdate` models a `customUpdate` that hands back the resource it was given. It asserts `Update` received a copy rather than the stored `desired`, that the copy was given the observed status, and that the copy kept its own conditions rather than adopting `latest`'s. It fails when either change is removed.

`resourceMocks` accepts any `SetStatus` argument, since that argument is no longer always the receiver, and tests reaching the update path stub `Conditions`.

`go build ./...` clean, `go test ./...` passes (10 packages), `gofmt` clean. `go vet ./pkg/runtime/` reports only warnings pre-existing on `main`.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -948,6 +948,24 @@ func (r *resourceReconciler) delayedReadOneAfterCreate(
 	return observed, nil
 }
 
+// setStatusWithoutConditions copies src's status onto dst, keeping dst's own
+// conditions.
+//
+// Status holds both service-reported state and ACK's conditions, and SetStatus
+// replaces the whole struct, so it cannot adopt observed state without also
+// adopting src's conditions. Neither side's conditions belong in the result:
+// adopting src's suppresses recomputation, since resetConditions clears
+// conditions each reconcile so ensureConditions can rebuild them and
+// ensureConditions only runs while Synced is nil; dropping dst's loses
+// conditions set after resetConditions, notably ACK.ReferencesResolved.
+//
+// src is deep-copied so the two do not share status pointers.
+func setStatusWithoutConditions(dst, src acktypes.AWSResource) {
+	conditions := dst.Conditions()
+	dst.SetStatus(src.DeepCopy())
+	dst.ReplaceConditions(conditions)
+}
+
 // updateResource calls one or more AWS APIs to modify the backend AWS resource
 // and patches the CR's Metadata and Spec back to the Kubernetes API.
 //
@@ -987,18 +1005,16 @@ func (r *resourceReconciler) updateResource(
 	// paths are known syntactically valid; no need to re-check.
 	// Hand Update a copy, never the stored `desired` itself. ACK permits a
 	// resource manager's Update to return the object it was given, and several
-	// hand-written customUpdate implementations do exactly that. If that object
-	// were `desired`, it would also be the base of the status merge patch in
-	// patchResourceStatus -- base and target would be the same object, the
-	// computed patch would be empty, and any status the manager just set
-	// (conditions, observed state) would be silently dropped.
+	// hand-written customUpdate implementations do. Were that object `desired`,
+	// it would also be the base of the status merge patch in
+	// patchResourceStatus: base and target identical, patch empty, and any
+	// status the manager just set silently dropped.
 	reconcileDesired := desired.DeepCopy()
 	if r.cfg.FeatureGates.IsEnabled(featuregate.IgnoreFieldDrift) &&
 		HasIgnoreFieldDrift(desired) {
-		// Transform the copy, not `desired`. applyIgnoredFields returns its
-		// first argument unchanged on several paths, so feeding it the copy
-		// keeps the "never hand Update the stored desired" invariant true here
-		// regardless of which path it takes, and avoids a second deep copy.
+		// Transform the copy, not `desired`: applyIgnoredFields returns its first
+		// argument unchanged on several paths, so passing the copy keeps the
+		// invariant above regardless of path, and avoids a second deep copy.
 		merged, mergeErr := applyIgnoredFields(reconcileDesired, latest, r.cfg.FeatureGates)
 		if mergeErr != nil {
 			rlog.Info(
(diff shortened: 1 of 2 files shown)


pkg/runtime/reconciler.go
```diff
@@ -985,7 +985,14 @@ func (r *resourceReconciler) updateResource(
...
+	// resource manager's Update to return the object it was given, and several
+	// hand-written customUpdate implementations do exactly that. If that object
+	// were `desired`, it would also be the base of the status merge patch in
+	// patchResourceStatus -- base and target would be the same object, the
+	// computed patch would be empty, and any status the manager just set
+	// (conditions, observed state) would be silently dropped.
+	reconcileDesired := desired.DeepCopy()
 	if r.cfg.FeatureGates.IsEnabled(featuregate.IgnoreFieldDrift) &&
 		HasIgnoreFieldDrift(desired) {
 		merged, mergeErr := applyIgnoredFields(desired, latest, r.cfg.FeatureGates)
```
michaelhtm: can you pass `reconcileDesired` here as well? I see we still return desired here: https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/ignore_field_drift.go#L88

gustavodiaz7722: Good catch, that would essentially overwrite this fix. I've updated ```applyIgnoredFields(desired, latest, r.cfg.FeatureGates)``` so we correctly pass the copy reconcileDesired.

michaelhtm (approved):

sapphirew (approved):

michaelhtm (approved):

sapphirew (approved):

michaelhtm (approved):
/lgtm

PR #267, opened 2026-08-27
feat: restore nested cross-resource references after Create and Update

### LINKED ISSUE #2431: The ACK Lambda Controller modifies the object spec
**Describe the bug**
I encountered an issue while creating a Function using `vpcConfig.securityGroupRefs`:

```yaml
status:
  ackResourceMetadata:
    arn: arn:aws:lambda:us-east-1:1234:function:pr13344879357-ci-va6-cw2splunk
    ownerAccountID: "1234"
    region: us-east-1
  conditions:
  - lastTransitionTime: "2025-04-17T11:22:25Z"
    message: Reference resolution failed
    reason: 'both resource reference wrapper and ID cannot be used together: VPCConfig.SecurityGroupIDs,VPCConfig.SecurityGroupRefs'
    status: Unknown
    type: ACK.ReferencesResolved
  lastModified: 2025-04-17T09:41:44.372+0000
  lastUpdateStatus: Successful
  revisionID: 93e97a3b-c3b0-451c-af76-b2928cb415bd
  state: Active
```

My spec contains these lines:
```yaml
    securityGroupRefs:
      - from:
          name: {{ include "cluster-build.clusterFullName" $ }}-cw2splunklambda
```

**Steps to reproduce**

I've also created a simple manifest with this in its spec:

```yaml
  vpcConfig:
    securityGroupRefs:
      - from:
          name: pr13344879357-ci-va6-cw2splunklambda
```

And the controller took this snippet of code and transformed it into:

```yaml
  vpcConfig:
    securityGroupIDs:
    - sg-042a062cc8a002db3
```

```sh
kubectl get securitygroup -n ci-clusters  pr13344879357-ci-va6-cw2splunklambda 
NAME                                   ID
pr13344879357-ci-va6-cw2splunklambda   sg-042a062cc8a002db3
```

**Expected outcome**

The ACK Lambda Controller accepts both securityGroupRefs and securityGroupIDs / just one of them, but it does not transform securityGroupRefs into securityGroupIDs.

I assume that AWS Lambda template doesn't support reference by name 
**Environment**

* Kubernetes version 1.30
* Using EKS (yes/no), if so version?  1.30
* AWS service targeted (S3, RDS, etc.) lambda-controller:1.6.1

gustavodiaz7722:
## Summary

A cross-resource reference (`*Ref`) is generated as a **sibling** of the concrete field it resolves into — `spec.vpcConfig.subnetRefs` next to `spec.vpcConfig.subnetIDs`. A resource manager builds its return value from an AWS API response, which has no concept of a reference, so rebuilding the containing struct drops every `*Ref` inside it.

That disables `ClearResolvedReferences`, which suppresses a resolved value only while it can still see the sibling:

```go
if ko.Spec.VPCConfig != nil {
	if len(ko.Spec.VPCConfig.SubnetRefs) > 0 {   // false once the *Ref is gone
		ko.Spec.VPCConfig.SubnetIDs = nil
	}
}
```

So the spec patch **deletes the declared `*Ref` and stores the resolved value in its place** — what aws-controllers-k8s/community#2431 reports: a declared `securityGroupRefs` replaced by `securityGroupIDs`.

Reconciliation continues until the manifest is applied again, from Helm, Argo, Flux or `kubectl apply`. That apply restores the `*Ref` beside the now-stored value, and `validateReferenceFields` rejects the pair:

```
message: Reference resolution failed
reason: 'both resource reference wrapper and ID cannot be used together:
         VPCConfig.SubnetIDs,VPCConfig.SubnetRefs'
```

This PR adds an optional `ReferenceEnsurer` interface and invokes it on the object a resource manager hands back from `Create` and from `Update`, sourcing the references from the declared resource.

Fixes aws-controllers-k8s/community#2431, and addresses the struct-nested half of aws-controllers-k8s/community#2361. Pairs with aws-controllers-k8s/code-generator#738, which generates the method.

## Backwards compatible

`ReferenceEnsurer` is deliberately separate from `ReferenceManager` and reached through a type assertion, so every controller generated before the method existed still satisfies `AWSResourceManager` and compiles unchanged. Those controllers take the existing path untouched and opt in by regenerating. `TestReconcilerUpdate_WithoutEnsurerIsUnaffected` pins that.

## Where the restoration runs

On what a resource manager returns from `Create` and from `Update` — the two paths where the object about to be patched back was rebuilt from an API response while the patch base is still the resource the user declared. It runs before the error is inspected, since a manager may hand back a resource alongside a requeue error while an asynchronous operation is in flight and that object reaches the caller either way.

**Not on `AdoptionPolicy_Adopt`.** Under that policy the spec is populated from the observed AWS resource, so a declared spec is expected to be replaced rather than preserved, and a declared `*Ref` is replaced along with every other declared field. Restoring it would make the reference the one exception. `TestReconcilerAdopt_DoesNotEnsureReferences` asserts that.

**`AdoptionPolicy_AdoptOrCreate` does keep a declared reference**, for a structural reason rather than because the restoration runs: that branch marks the resource managed and adopted and requeues, and that patch's base is a `DeepCopy` of its own target, so nothing ever patches the spec with the declared resource as the base. `TestReconcilerAdoptOrCreate_PreservesReferences` asserts that too — both halves, because moving the restoration into `patchResourceMetadataAndSpec` would pick up the adopt branch automatically and silently change it.

Also not on the late-initialization patch, whose base is the AWS-observed object and carries no references — `TestReconcilerUpdate_LateInitializeIsNotAffectedByEnsureReferences` keeps that boundary asserted.

## Why `desired` and not `reconcileDesired`

`reconcileDesired` is what gets handed to `rm.Update`, and a resource manager may mutate the object it is given: `apigateway`'s `ApiKey` `sdkUpdate` assigns `desired.ko.Spec.StageKeys` straight from the `UpdateApiKey` response (`sdk.go:331`), and `applyIgnoredFields` merges observed values into it for a resource carrying the ignore-field-drift annotation. Only `desired` is a clean record of what the user declared, and it is also the base of the spec patch that follows, so the two agree by construction.

## Create now gets a copy of `desired`

`createResource` passed `desired` itself to `rm.Create`. Generated `sdkCreate` only
deep-copies the resource it is given partway through -- a `custom_implementation`
returns before that point, and a `sdk_create_pre_build_request` hook runs before it
-- so either could mutate the object the user declared. `desired` is both the patch
base and the reference source, so it has to stay a clean record, and `Create` gets
`desired.DeepCopy()`. This mirrors `updateResource`, which already hands `Update` a
copy for the same reason.

The copy is taken after `setResourceManaged` and `EnsureTags`, so it carries the
finalizer and the controller tags.

No hook template in the fleet writes to `desired.ko`, and neither custom Create
implementation (`elasticache`'s `CustomCreateSnapshot`, `apigatewayv2`'s
`customCreateApi`) mutates its input -- the former deep-copies first, the latter
only reads -- so this is hardening rather than a fix.
`TestReconcilerCreate_PassesCopyToCreate` pins it.

## What the generated method does

Detailed in aws-controllers-k8s/code-generator#738. Summarised here because it bounds this PR's blast radius:

- **top-level `*Ref`** — nothing emitted; it cannot be lost, since every write path starts from a `DeepCopy` of the object it was handed.
- **reached through structs** — only the reference field is assigned, so every concrete value the service reported stands. This is the whole of what is emitted, and it codifies a pattern `eks/cluster`, `lambda/function` and `opensearchservice/domain` already hand-maintain in `sdk_*_post_set_output` hooks.
- **reached through a list** — nothing emitted; no fixed address to assign to, and no sound way to pair an observed element with a declared one. These behave exactly as they do today.

So **117 struct-nested references across 37 resources in 25 controllers** change behaviour; the 315 top-level and 38 list-nested ones are untouched.

## Testing

Eleven tests in `pkg/runtime/reconciler_test.go`:

| Test | Asserts |
|---|---|
| `TestReconcilerCreate_EnsuresReferencesAfterCreate` | runs after `Create`, sourced from `desired` |
| `TestReconcilerUpdate_EnsuresReferencesAfterUpdate` | the same after `Update`, and the returned object is what gets patched |
| `TestReconcilerCreate_EnsuresReferencesOnCreateError` | still runs when `Create` returns an error |
| `TestReconcilerUpdate_EnsuresReferencesOnUpdateError` | the same on the update error path |
| `TestReconcilerAdopt_DoesNotEnsureReferences` | `AdoptionPolicy_Adopt` deliberately does not restore |
| `TestReconcilerAdoptOrCreate_PreservesReferences` | `AdoptionPolicy_AdoptOrCreate` keeps a declared reference |
| `TestReconcilerUpdate_LateInitializeIsNotAffectedByEnsureReferences` | never runs on the late-init patch |
| `TestReconcilerUpdate_EnsureReferencesToleratesNilLatest` | a manager returning `(nil, err)` passes through instead of panicking |
| `TestReconcilerUpdate_WithoutEnsurerIsUnaffected` | a manager not implementing the interface is untouched |
| `TestReconcilerCreate_PassesCopyToCreate` | `Create` gets a copy, so `desired` stays a clean record |
| `TestReconcilerUpdate_PassesCopyOfDesiredToUpdate` | the pre-existing counterpart on the update path |

Full runtime suite passes.

Verified on a cluster, A/B against the unmodified controller. An `ecs` `Service` declaring `networkConfiguration.awsVPCConfiguration.subnetRefs` and `.securityGroupRefs` keeps both through create, update, re-apply and repeated resyncs, with no resolved IDs written to the spec. The same manifest under the unmodified controller loses both and stores the resolved `subnets`/`securityGroups` in their place; re-applying it then halts the resource on `both resource reference wrapper and ID cannot be used together`, which is the failure this prevents. Deletion completes cleanly. `adopt-or-create` against a pre-existing AWS resource keeps the references; strict `adopt` replaces them, as intended. Separately, an `ecr` `Repository` whose `EncryptionConfiguration.KMSKey` is both a reference and `late_initialize` reaches `ACK.LateInitialized=True` with the reference kept and no churn.

## Not addressed

**References reached through a list.** No fixed address to assign to, and no sound way to pair an observed element with a declared one: an AWS response need not preserve request order, and 75 of the roughly 470 configured references resolve to a `Spec.*` path with no uniqueness guarantee. These behave exactly as they do today. This is the remaining half of #2361.

**`AdoptionPolicy_Adopt`** replaces a declared reference along with the rest of the declared spec. That is intentional, not a gap — see above.

**The existing read-path hooks.** `lambda/function`, `eks/cluster` and `opensearchservice/domain` restore these references by hand in `sdk_read_one_post_set_output`. Their create-path halves are subsumed. **The read-path halves must stay** — the `AdoptionPolicy_Adopt` and delete-path spec writes are not covered here.


pkg/types/reference_manager.go
```diff
@@ -36,3 +36,47 @@ type ReferenceManager interface {
...
+// reconciliation. See aws-controllers-k8s/community#2361 and #2431.
+//
+// Kept separate from ReferenceManager and reached through a type assertion, so
+// controllers generated before the method existed still satisfy
+// AWSResourceManager. They opt in by regenerating.
+//
+// Only a reference reached through STRUCTS is restored. One reached through a LIST
+// is not: it has no fixed address, and replacing the whole outermost list instead
+// would discard whatever the service populated inside it. Those references behave
+// as they did before this interface existed.
```
knottnt: nit: This is a property of the generated code. In the future we could improve the generated code to handle this better or at least in some cases where a consistent entry identifier can be used making this statement outdated.

pkg/types/reference_manager.go
```diff
@@ -36,3 +36,47 @@ type ReferenceManager interface {
...
+	// EnsureReferences returns a copy of `latest` with the reference fields it is
+	// missing restored from `desired`. Only reference fields are written, so every
+	// concrete value on `latest` stands.
+	//
+	// `desired` must be the DECLARED resource with its references resolved, and
+	// must not be an object that has been through a resource manager: managers may
+	// mutate what they are handed, and some write API response values into it.
+	//
+	// Unlike ResolveReferences this takes no context and returns no error: it makes
+	// no API calls and reads nothing outside the two resources it is given.
```
knottnt: nit: this is a bit redundant since this info is largely present in the function signature.

gustavodiaz7722: Removed

pkg/types/reference_manager.go
gustavodiaz7722: Agreed, will update comment

knottnt (approved):

knottnt (commented):
@gustavodiaz7722 Changes to the primary create/update paths look good. However, there may be a remaining edge case in the pre-delete-sync(prd) path where the prd update succeeds and the reference is overwritten, but for whatever reason the update fails or is delayed. I believe the most common path that edge case could be hit is a resource with an async delete that use a requeue error to inform the controller that it needs to wait for the resource to disappear or hit a terminal DELETED state.

knottnt (approved):
/lgtm

